PR #127, opened 2022-05-29
fix(start_argus): Use master process for uWSGI and replace bash shell

k0machi:
This addresses the problem where systemd would kill the entire service
when the "main" worker exits. Use of exec removes extra "bash" process
from the resulting CGroup


fruch (approved):
LGTM

bentsi (approved):

PR #130, opened 2022-06-05
improvement(argus): Add linter and formatter for JavaScript

k0machi:
This commit adds ESLint for linting and enforcing a code style and
Prettier code formatter for formatting files. It also adds matching
configurations to them.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,47 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "node": true,
+        "es2021": true
+    },
+    "extends": [
+        "eslint:recommended",
+        "plugin:@typescript-eslint/recommended"
+    ],
+    "parser": "@typescript-eslint/parser",
+    "parserOptions": {
+        "ecmaVersion": "latest",
+        "sourceType": "module"
+    },
+    "plugins": [
+        "@typescript-eslint",
+        "svelte3"
+    ],
+    "overrides": [
+        {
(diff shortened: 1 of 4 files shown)


fruch:
@k0machi how about adding a github action for running it ? (or even adding pre-commit, and running it on an action) ?

.eslintrc.js
```diff
@@ -0,0 +1,47 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "node": true,
+        "es2021": true
+    },
+    "extends": [
+        "eslint:recommended",
```
fruch, line 8: isn't it better to pin those down ? to not get surprises ?

fruch:
@k0machi all the code is passing this reformatter and lint ? or those changes would be in a new PR/commit ?

.eslintrc.js
k0machi: It's a lot of rules that have to be defined explicitly and they're pretty universal, nothing opinionated.

k0machi:
> @k0machi all the code is passing this reformatter and lint ? or those changes would be in a new PR/commit ?

That would be next PR/commit, I still need to run it initially and fix all the problems in all files that crop up.

.eslintrc.js
fruch: I'm just saying it might be changed from time to time, it doesn't have versions ?

fruch (approved):
LGTM

PR #137, opened 2022-06-16
fix(NotificationManager): use main.runs route for run links

k0machi:
[Trello](https://trello.com/c/Az1zGsU6/5120-notifications-use-old-testrun-route)


fruch (approved):
LGTM

aleksbykov (approved):
LGTM

PR #149, opened 2022-07-12
feature(TestRunComments.svelte): Quoting when submitting new comment

### LINKED ISSUE #124: "Discussion" tab: please, add "reply" option to the message
"Discussion" tab: please, add "reply" option to the message.
It means if the reply is chosen, add user mention (@name) to the message body so the user will receive notificatrion

k0machi:
This commit adds a "Quote Reply" button to the Comment component,
allowing user to quote a specific comment when creating a new comment

Closes #124

### DIFF
--- a/frontend/Discussion/Comment.svelte
+++ b/frontend/Discussion/Comment.svelte
@@ -1,7 +1,7 @@
 <script>
     import * as marked from "marked";
     import Fa from "svelte-fa";
-    import { faEdit, faTrashAlt } from "@fortawesome/free-regular-svg-icons";
+    import { faEdit, faArrowAltCircleLeft, faTrashAlt } from "@fortawesome/free-regular-svg-icons";
     import humanizeDuration from "humanize-duration";
     import CommentEditor from "./CommentEditor.svelte";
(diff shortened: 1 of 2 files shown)


k0machi:
> Does it create a notification for the user I replies to?

No, this is to purely to copy comment text, if there's a user mention in the quote however, it will create a notification. 

For something like notifications for replies and such we need a subscription system for threads implemented first for it to not be too hacky.

roydahan:
The whole purpose of the issue from @juliayakovlev, was to send a notification to someone, once there was a reply.

k0machi:
I can append a mention on top of the reply to produce a notification this way, it's hacky but it's a start.

juliayakovlev:
> I can append a mention on top of the reply to produce a notification this way, it's hacky but it's a start.

it would be great to add a mention because it's expected that the autor of replyed message will get notification

k0machi:
Added comment author mention to emit a notification:
image

juliayakovlev (approved):

roydahan (approved):

roydahan:
I think it’s good enough for now.

PR #151, opened 2022-07-24
REST API Rework

k0machi:
This PR provides groundwork for the future REST API client and supporting multiple test types aside from SCT - it does so by implementing a plugin subsystem that dynamically loads available plugins into the web application on startup. For now, plugin subsystem is only used for client API router, refactors to use it within frontend and primary testrun service are planned. The PR also includes refactors that move all database handling to python scylla driver's Object Mapper, deprecating current ArgusDatabase interface.

The plugin subsystem documentation will follow in a different PR.

[Trello](https://trello.com/c/QXkz9V2P/4938-argus-rest-api-support-multiple-test-types)


roydahan (approved):
FWIW it looks good to me after the walkthrough you gave me.

PR #152, opened 2022-07-24
improvement(TestRunInfo.svelte): Show upgraded scylla version

k0machi:
Adds an "Upgraded scylla version" field to the test run card, visible
when packages field contains a scylla-server-upgraded package. Renames
normal "Scylla Version" field to "Base scylla version" field if that's
the case.

### DIFF
--- a/frontend/Common/RunUtils.js
+++ b/frontend/Common/RunUtils.js
@@ -5,3 +5,7 @@ export const getScyllaPackage = function (packages) {
 export const getKernelPackage = function (packages) {
     return packages.find((pkg) => pkg.name == "kernel");
 };
+
+export const getUpgradedScyllaPackage = function(packages) {
+    return packages.find((pkg) => pkg.name == "scylla-server-upgraded");
+};
--- a/frontend/TestRun/TestRunInfo.svelte
+++ b/frontend/TestRun/TestRunInfo.svelte
@@ -8,7 +8,7 @@
     import Fa from "svelte-fa";
     import { InProgressStatuses } from "../Common/TestStatus";
     import { timestampToISODate } from "../Common/DateUtils";
-    import { getScyllaPackage, getKernelPackage } from "../Common/RunUtils";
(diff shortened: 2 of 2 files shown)


frontend/Common/RunUtils.js
```diff
@@ -5,3 +5,7 @@ export const getScyllaPackage = function (packages) {
 export const getKernelPackage = function (packages) {
     return packages.find((pkg) => pkg.name == "kernel");
 };
+
+export const getUpgradedScyllaPackage = function(packages) {
+    return packages.find((pkg) => pkg.name == "scylla-server-upgraded");
```
soyacz, line 10: I think upgraded version is stored in `scylla-server` package. Upgraded version keeps version after upgrade.

k0machi: Doesn't seem to be the case, I'm looking at the latest rolling upgrade longevity and this is what I see:
image

soyacz: That's what I mean, 5.0.1 is version after upgrade - so shouldn't 'Upgraded scylla version' should take value from 'scylla-server` which points to start version (5.0 in this case)?

k0machi: The version in `scylla-server` is currently being output as `Base scylla version`, as for phrasing of this one, what would you suggest to clear it up?

fgelcer: > That's what I mean, 5.1.dev is version after upgrade - so shouldn't 'Upgraded scylla version' should take value from 'scylla-server` which points to start version (5.0 in this case)?

we should be upgrading (if the run was in the latest master one) from `5.0.0` (latest released OSS version) to latest master one (`5.1.dev`)

soyacz: ok, just wanted to confirm that :sweat_smile:

soyacz (approved):

roydahan (approved):

PR #160, opened 2022-08-10
feature(argus): REST HTTP Client

k0machi:
This PR adds a REST API client for Argus, as both a generic base
client and an SCT client.

Closes scylladb/qa-tasks#431

[Trello](https://trello.com/c/djMstpPv/5338-argus-rest-api-api-client)

### DIFF
--- a/argus/client/base.py
+++ b/argus/client/base.py
@@ -1,0 +1,171 @@
+import re
+import logging
+from dataclasses import asdict
+from typing import Any, Type
+from uuid import UUID
+
+import requests
+
+from argus.backend.util.enums import TestStatus
+from argus.client.sct.types import LogLink
+
+JSON = dict[str, Any] | list[Any] | int | str | float | bool | Type[None]
+LOGGER = logging.getLogger(__name__)
+
+
+class ArgusClientError(Exception):
+    pass
+
+
+class ArgusClient:
+    schema_version: str | None = None
+
+    class Routes():
+        # pylint: disable=too-few-public-methods
+        SUBMIT = "/testrun/$type/submit"
+        HEARTBEAT = "/testrun/$type/$id/heartbeat"
+        SET_STATUS = "/testrun/$type/$id/set_status"
+        SET_PRODUCT_VERSION = "/testrun/$type/$id/update_product_version"
+        SUBMIT_LOGS = "/testrun/$type/$id/logs/submit"
+        FINALIZE = "/testrun/$type/$id/finalize"
+
+    def __init__(self, auth_token: str, base_url: str, api_version="v1") -> None:
+        self._auth_token = auth_token
+        self._base_url = base_url
+        self._api_ver = api_version
+
+    @property
+    def auth_token(self) -> str:
+        return self._auth_token
+
+    def verify_location_params(self, endpoint: str, location_params: dict[str, str]) -> bool:
+        required_params: list[str] = re.findall(r"\$[\w_]+", endpoint)
+        for param in required_params:
+            if param.lstrip("$") not in location_params.keys():
+                raise ArgusClientError(f"Missing required location argument for endpoint {endpoint}: {param}")
+
+        return True
+
+    @staticmethod
+    def check_response(response: requests.Response, expected_code: int = 200):
+        if response.status_code != expected_code:
+            raise ArgusClientError(
+                f"Unexpected HTTP Response encountered - expected: {expected_code}, got: {response.status_code}",
+                expected_code,
+                response.status_code,
+                response.request,
+            )
+
+        response_data: JSON = response.json()
+        LOGGER.debug("API Response: %s", response_data)
+        if response_data.get("status") != "ok":
+            raise ArgusClientError(
+                f"API Error encountered using endpoint: {response.request.method} {response.request.path_url}",
+                response_data["message"],
+            )
+
+    def get_url_for_endpoint(self, endpoint: str, location_params: dict[str, str] | None) -> str:
+        if self.verify_location_params(endpoint, location_params):
+            for param, value in location_params.items():
+                endpoint = endpoint.replace(f"${param}", str(value))
+        return f"{self._base_url}/api/{self._api_ver}/client{endpoint}"
+
+    @property
+    def generic_body(self) -> dict:
+        return {
+            "schema_version": self.schema_version
+        }
+
+    @property
+    def request_headers(self):
+        return {
+            "Authorization": f"token {self.auth_token}",
+            "Accept": "application/json",
+            "Content-Type": "application/json",
+        }
+
+    def get(self, endpoint: str, location_params: dict[str, str] = None, params: dict = None) -> requests.Response:
+        response = requests.get(
+            url=self.get_url_for_endpoint(
+                endpoint=endpoint,
+                location_params=location_params
+            ),
+            params=params,
+            headers=self.request_headers
+        )
+
+        return response
+
+    def post(
+        self,
+        endpoint: str,
+        location_params: dict = None,
+        params: dict = None,
+        body: dict = None,
+    ) -> requests.Response:
+        response = requests.post(
+            url=self.get_url_for_endpoint(
+                endpoint=endpoint,
+                location_params=location_params
+            ),
+            params=params,
+            json=body,
+            headers=self.request_headers,
+        )
+
+        return response
+
+    def submit_run(self, run_type: str, run_body: dict) -> requests.Response:
+        return self.post(endpoint=self.Routes.SUBMIT, location_params={"type": run_type}, body={
+            **self.generic_body,
+            **run_body
+        })
+
+    def set_status(self, run_type: str, run_id: UUID, new_status: TestStatus) -> requests.Response:
+        return self.post(
+            endpoint=self.Routes.SET_STATUS,
+            location_params={"type": run_type, "id": str(run_id)},
+            body={
+                **self.generic_body,
+                "new_status": new_status.value
+            }
+        )
+
+    def update_product_version(self, run_type: str, run_id: UUID, product_version: str) -> requests.Response:
+        return self.post(
+            endpoint=self.Routes.SET_PRODUCT_VERSION,
+            location_params={"type": run_type, "id": str(run_id)},
+            body={
+                **self.generic_body,
+                "product_version": product_version
+            }
+        )
+
+    def submit_logs(self, run_type: str, run_id: UUID, logs: list[LogLink]) -> requests.Response:
+        return self.post(
+            endpoint=self.Routes.SUBMIT_LOGS,
+            location_params={"type": run_type, "id": str(run_id)},
+            body={
+                **self.generic_body,
+                "logs": [asdict(l) for l in logs]
+            }
+        )
+
+    def finalize_run(self, run_type: str, run_id: UUID) -> requests.Response:
+        return self.post(
+            endpoint=self.Routes.FINALIZE,
+            location_params={"type": run_type, "id": str(run_id)},
+            body={
+                **self.generic_body,
+            }
+        )
+
+    def heartbeat(self, run_type: str, run_id: UUID) -> None:
+        response = self.post(
+            endpoint=self.Routes.HEARTBEAT,
+            location_params={"type": run_type, "id": str(run_id)},
+            body={
+                **self.generic_body,
+            }
+        )
+        self.check_response(response)
--- a/argus/client/sct/client.py
+++ b/argus/client/sct/client.py
@@ -1,0 +1,216 @@
+from uuid import UUID
+from dataclasses import asdict
+from argus.backend.util.enums import ResourceState, TestStatus
+from argus.client.base import ArgusClient
+from argus.client.sct.types import EventsInfo, LogLink, Package
+
+
+class ArgusSCTClient(ArgusClient):
+    test_type = "scylla-cluster-tests"
+    schema_version: None = "v8"
+
+    class Routes(ArgusClient.Routes):
+        SUBMIT_PACKAGES = "/sct/$id/packages/submit"
+        SUBMIT_SCREENSHOTS = "/sct/$id/screenshots/submit"
+        CREATE_RESOURCE = "/sct/$id/resource/create"
+        TERMINATE_RESOURCE = "/sct/$id/resource/$name/terminate"
+        UPDATE_SHARDS_FOR_RESOURCE = "/sct/$id/resource/$name/shards"
+        SUBMIT_NEMESIS = "/sct/$id/nemesis/submit"
+        FINALIZE_NEMESIS = "/sct/$id/nemesis/finalize"
+        SUBMIT_EVENTS = "/sct/$id/events/submit"
+
+    def __init__(self, run_id: UUID, auth_token: str, base_url: str, api_version="v1") -> None:
+        super().__init__(auth_token, base_url, api_version)
+        self.run_id = run_id
+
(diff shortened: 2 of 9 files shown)


argus/client/sct/client.py
```diff
@@ -0,0 +1,156 @@
...
+        return self.post(
+            endpoint=self.Routes.SUBMIT_PACKAGES,
+            location_params={"id": str(self.run_id)},
+            body={
+                **self.generic_body,
+                "packages": packages,
+            }
+        )
+
+    def submit_screenshots(self, screenshots: list[str]) -> requests.Response:
```
soyacz, line 60: to be clear, name it `screenshots_links`

argus/client/sct/client.py
```diff
@@ -0,0 +1,156 @@
...
+    def update_scylla_version(self, version: str) -> requests.Response:
+        return super().update_product_version(run_type=self.test_type, run_id=self.run_id, product_version=version)
+
+    def submit_sct_logs(self, logs: list[dict[str, str]]) -> requests.Response:
+        return super().submit_logs(run_type=self.test_type, run_id=self.run_id, logs=logs)
+
+    def finalize_sct_run(self) -> requests.Response:
+        return super().finalize_run(run_type=self.test_type, run_id=self.run_id)
+
+    def submit_packages(self, packages: list[dict]):
```
soyacz, line 50: `packages` - list of dict doesn't say much about structure. I propose creating dataclass Package (or more precise name) so when using api, user can easily get what is needed.

soyacz: missing return type hint

argus/client/sct/client.py
```diff
@@ -0,0 +1,156 @@
...
+            "sct_config": sct_config,
+        })
+
+    def set_sct_run_status(self, new_status: TestStatus) -> requests.Response:
+        return super().set_status(run_type=self.test_type, run_id=self.run_id, new_status=new_status)
+
+    def update_scylla_version(self, version: str) -> requests.Response:
+        return super().update_product_version(run_type=self.test_type, run_id=self.run_id, product_version=version)
+
+    def submit_sct_logs(self, logs: list[dict[str, str]]) -> requests.Response:
```
soyacz, line 44: instead of `dict[str, str]` i propose:
```
@dataclass
class LogLinkInfo:
    log_name: str
    log_link: str
```

argus/client/sct/client.py
```diff
@@ -0,0 +1,156 @@
...
+        SUBMIT_NEMESIS = "/sct/$id/nemesis/submit"
+        FINALIZE_NEMESIS = "/sct/$id/nemesis/finalize"
+        SUBMIT_EVENTS = "/sct/$id/events/submit"
+
+    def __init__(self, run_id: UUID, auth_token: str, base_url: str, api_version="v1") -> None:
+        super().__init__(auth_token, base_url, api_version)
+        self.run_id = run_id
+
+    def submit_sct_run(self, job_name: str, job_url: str, started_by: str, commit_id: str,
+                       runner_public_ip: str, runner_private_ip: str, sct_config: str) -> requests.Response:
```
soyacz, line 26: what is `sct_config` string? Do you mean `sct_config_path` to refer path of the config?

argus/client/sct/client.py
```diff
@@ -0,0 +1,156 @@
...
+                "nemesis": {
+                    "name": name,
+                    "start_time": start_time,
+                    "status": status,
+                    "message": message,
+                }
+            }
+        )
+
+    def submit_events(self, events: list[dict]) -> requests.Response:
```
soyacz, line 145: here also dict is not telling the structure, better use some DTO

soyacz (changes requested):
I think Argus client methods shouldn't return `requests.Response` objects. As client methods are rather commands, they should return nothing or raise exception when something goes wrong.

aleksbykov (commented):
LGTM. 
Required may document with example of usage.
am i right, that we should have appropriate backend logic (Plugin?) which will response to the client  requests?

k0machi:
> I think Argus client methods shouldn't return `requests.Response` objects. As client methods are rather commands, they should return nothing or raise exception when something goes wrong.

Yes, that's a good idea. Currently REST methods themselves pretty much return nothing - they either echo the data sent, or reply with simple strings that indicate action completion, so we could instead return nothing and just handle `requests.Response` objects internally.

argus/client/sct/client.py
k0machi: > `packages` - list of dict doesn't say much about structure. I propose creating dataclass Package (or more precise name) so when using api, user can easily get what is needed.

I am actually thinking right now what would be better - introducing dataclasses or type hinted dicts with preset keys. Type Hinting could solve not understanding which keys you need to provide while keeping the serialization into json simple. Dataclass will most likely need an introduction of an (admittely simple) custom json encoder.

argus/client/sct/client.py
k0machi: Good catch, this is supposed to be dict and the intention that you pass whole `SCTConfiguration` class instance as a dict.

k0machi:
> LGTM. Required may document with example of usage. am i right, that we should have appropriate backend logic (Plugin?) which will response to the client requests?

Yes, backend logic is already implemented, basic test submission stuff is `argus/backend/controller/client_api.py`, while sct-specific one is `argus/backend/plugin/sct/controller.py`

argus/client/base.py
```diff
@@ -0,0 +1,133 @@
...
+        self._api_ver = api_version
+
+    @property
+    def auth_token(self) -> str:
+        return self._auth_token
+
+    def get_url_for_endpoint(self, endpoint: 'ArgusClient.Routes', location_params: dict[str, str] | None) -> str:
+        if location_params:
+            for param, value in location_params.items():
+                endpoint = endpoint.replace(f"${param}", value)
```
soyacz, line 33: add verification that all required params has been replaced (point to missing elements)

soyacz (changes requested):
@k0machi you requested review after accepting my comments, but I don't see any commits since my review - did you forget to push?

argus/client/sct/client.py
k0machi: Done, decided to settle on intermediary dataclasses

argus/client/sct/client.py
k0machi: Done

argus/client/sct/client.py
k0machi: Done

argus/client/sct/client.py
k0machi: Done

argus/client/base.py
k0machi: Done

k0machi:
> @k0machi you requested review after accepting my comments, but I don't see any commits since my review - did you forget to push?

Apologies, I accidentally re-requested it when I was looking and replying to some of your comments

k0machi:
> I think Argus client methods shouldn't return `requests.Response` objects. As client methods are rather commands, they should return nothing or raise exception when something goes wrong.

I'm thinking about this meanwhile, and I'm currently not sure how I want to handle this - should the base client have a generic handler for API and HTTP errors and raise an exception during one of the Verb methods (get/post)? Or should each client just handle errors itself? The latter one would lead to a lot of code duplication checking status code and json response, the former might reduce flexibility wrt to how API can respond.

soyacz:
> > I think Argus client methods shouldn't return `requests.Response` objects. As client methods are rather commands, they should return nothing or raise exception when something goes wrong.
> 
> I'm thinking about this meanwhile, and I'm currently not sure how I want to handle this - should the base client have a generic handler for API and HTTP errors and raise an exception during one of the Verb methods (get/post)? Or should each client just handle errors itself? The latter one would lead to a lot of code duplication checking status code and json response, the former might reduce flexibility wrt to how API can respond.

To avoid code duplication you could enhance generic post/get methods with argument of expected response code which would raise exception on mismatch. Another option would be adding function that would take response and expected response code and do validation. Not always the same code blocks are considered as code duplication - sometimes it is just a coincidence.

k0machi:
> > > I think Argus client methods shouldn't return `requests.Response` objects. As client methods are rather commands, they should return nothing or raise exception when something goes wrong.
> > 
> > 
> > I'm thinking about this meanwhile, and I'm currently not sure how I want to handle this - should the base client have a generic handler for API and HTTP errors and raise an exception during one of the Verb methods (get/post)? Or should each client just handle errors itself? The latter one would lead to a lot of code duplication checking status code and json response, the former might reduce flexibility wrt to how API can respond.
> 
> To avoid code duplication you could enhance generic post/get methods with argument of expected response code which would raise exception on mismatch. Another option would be adding function that would take response and expected response code and do validation. Not always the same code blocks are considered as code duplication - sometimes it is just a coincidence.

Alright, done.

argus/client/base.py
```diff
@@ -0,0 +1,170 @@
...
+                raise ArgusClientError(f"Missing required location argument for endpoint {endpoint}: {param}")
+
+        return True
+
+    @staticmethod
+    def check_response(response: requests.Response, expected_code: int = 200) -> JSON:
+        if response.status_code != expected_code:
+            raise ArgusClientError(
+                f"Unexpected HTTP Response encountered - expected: {expected_code}, got: {response.status_code}",
+                expected_code,
```
soyacz, line 53: why provide expected_code and status_code args to exception? It is shown in message.

argus/client/base.py
```diff
@@ -0,0 +1,170 @@
...
+    def verify_location_params(endpoint: str, location_params: dict[str, str]) -> bool:
+        required_params: list[str] = re.findall(r"\$[\w_]+", endpoint)
+        for param in required_params:
+            if param.lstrip("$") not in location_params.keys():
+                raise ArgusClientError(f"Missing required location argument for endpoint {endpoint}: {param}")
+
+        return True
+
+    @staticmethod
+    def check_response(response: requests.Response, expected_code: int = 200) -> JSON:
```
soyacz, line 49: I'd argue if `check_response` should return anything - name rather indicates verification, not processing any data.

argus/client/base.py
```diff
@@ -0,0 +1,170 @@
...
+    def finalize_run(self, run_type: str, run_id: UUID) -> requests.Response:
+        return self.post(
+            endpoint=self.Routes.FINALIZE,
+            location_params={"type": run_type, "id": str(run_id)},
+            body={
+                **self.generic_body,
+            }
+        )
+
+    def heartbeat(self, run_type: str, run_id: UUID) -> requests.Response:
```
soyacz, line 163: still returning response while it is not necessary.

argus/client/sct/client.py
```diff
@@ -0,0 +1,172 @@
...
+        UPDATE_SHARDS_FOR_RESOURCE = "/sct/$id/resource/$name/shards"
+        SUBMIT_NEMESIS = "/sct/$id/nemesis/submit"
+        FINALIZE_NEMESIS = "/sct/$id/nemesis/finalize"
+        SUBMIT_EVENTS = "/sct/$id/events/submit"
+
+    def __init__(self, run_id: UUID, auth_token: str, base_url: str, api_version="v1") -> None:
+        super().__init__(auth_token, base_url, api_version)
+        self.run_id = run_id
+
+    def submit_sct_run(self, job_name: str, job_url: str, started_by: str, commit_id: str,
```
soyacz, line 26: Why to return response?
If some method returns something interesting for future processing it should be named&returned explicitly.

soyacz (changes requested):
Still methods return `str|None` while it is not clear why it is needed for.
Also would be nice to have some docs lines for methods - especially when one method requires execution of other one (e.g. does `subimt_nemesis` requires later `finalize_nemesis`?)

argus/client/base.py
k0machi: This is so that in case you want to propagate the exception somewhere else by raising another exception you have access to the affected fields without parsing the message.

argus/client/base.py
k0machi: Would `parse_response` be better name? That's one I used previously for this method, but then switched to this one.

argus/client/base.py
k0machi: Oops, missed this one

k0machi:
> Still methods return `str|None` while it is not clear why it is needed for. Also would be nice to have some docs lines for methods - especially when one method requires execution of other one (e.g. does `subimt_nemesis` requires later `finalize_nemesis`?)

So all methods in theory should return a result string - Maybe introducing a Result enum would be a good idea here; As for docstrings - good idea, will push them today.

soyacz:
> > Still methods return `str|None` while it is not clear why it is needed for. Also would be nice to have some docs lines for methods - especially when one method requires execution of other one (e.g. does `subimt_nemesis` requires later `finalize_nemesis`?)
> 
> So all methods in theory should return a result string - Maybe introducing a Result enum would be a good idea here; As for docstrings - good idea, will push them today.

What is this result string used for? When we send command like `heartbeat` or `update_product_version` do we need any response to process further?

argus/client/base.py
soyacz: no, `check_response` name is ok, but should not return/process anything (here by process I mean `response_data: JSON = response.json()`). Generally, if name indicates only verification let's not add another purpose to method (single responsibility).

k0machi:
> > > Still methods return `str|None` while it is not clear why it is needed for. Also would be nice to have some docs lines for methods - especially when one method requires execution of other one (e.g. does `subimt_nemesis` requires later `finalize_nemesis`?)
> > 
> > 
> > So all methods in theory should return a result string - Maybe introducing a Result enum would be a good idea here; As for docstrings - good idea, will push them today.
> 
> What is this result string used for? When we send command like `heartbeat` or `update_product_version` do we need any response to process further?

Just an informational result, we could either use a boolean or string or nothing to indicate success and an exception to indicate failure. Considering the exception indicates failure, I'm thinking about returning nothing as a success and logging the received message string, but we could alternatively return False on failure and True on success, or use enum values there if we want to have control flow based on API responses.

soyacz:
> > > > Still methods return `str|None` while it is not clear why it is needed for. Also would be nice to have some docs lines for methods - especially when one method requires execution of other one (e.g. does `subimt_nemesis` requires later `finalize_nemesis`?)
> > > 
> > > 
> > > So all methods in theory should return a result string - Maybe introducing a Result enum would be a good idea here; As for docstrings - good idea, will push them today.
> > 
> > 
> > What is this result string used for? When we send command like `heartbeat` or `update_product_version` do we need any response to process further?
> 
> Just an informational result, we could either use a boolean or string or nothing to indicate success and an exception to indicate failure. Considering the exception indicates failure, I'm thinking about returning nothing as a success and logging the received message string, but we could alternatively return False on failure and True on success, or use enum values there if we want to have control flow based on API responses.

If not needed, better to return nothing. Let's add later to these methods where needed. Returning nothing on success and exception on failure is quite common solution.

argus/client/base.py
k0machi: Nevermind, that's a base client method, returning a response here is fine as it's consumed by `sct_heartbeat` method.

soyacz: why not to move `self.check_response(response)` to base client method and don't return anything?

roydahan (approved):
FWIW, looks ok to me.

roydahan:
@soyacz / @aleksbykov let's get it reviewed so we can merge it.

soyacz (approved):

aleksbykov (approved):

PR #162, opened 2022-08-19
feature(email_notification):Send email notification

aleksbykov:
Trello: https://trello.com/c/IX07Qglw
If user will be mentioned in comments, the email notification
will be sent to mentioned user.

### DIFF
--- a/argus/backend/service/notification_manager.py
+++ b/argus/backend/service/notification_manager.py
@@ -1,7 +1,9 @@
 import logging
+from typing import Any
 from uuid import UUID
-from flask import render_template
+from flask import current_app, render_template
 from argus.backend.models.web import ArgusNotification, ArgusNotificationSourceTypes, ArgusNotificationTypes, ArgusNotificationState, User
+from argus.backend.util.send_email import Email
 
 LOGGER = logging.getLogger(__name__)
 
@@ -11,6 +13,45 @@ class NotificationManagerException(Exception):
 
 
 class NotificationManagerService:
+
+    def __init__(self, notification_senders: list["NotificationSenderBase"] | None = None) -> None:
+        self.notification_services = [
+            ArgusDBNotificationSaver(),
+            EmailNotificationServiceSender()
+        ]
+        if notification_senders:
+            self.notification_services.extend(notification_senders)
+
+    def send_notification(self, receiver: UUID, sender: UUID,
+                          notification_type: ArgusNotificationTypes,
+                          source_type: ArgusNotificationSourceTypes,
+                          source_id: UUID, title: str | None = None,
+                          source_message: str | None = None, content_params: dict | None = None) -> Any:
+
+        for service in self.notification_services:
+            service.send_notification(receiver, sender, notification_type, source_type,
+                                      source_id, title, source_message, content_params)
+
+    def get_notificaton(self, receiver: UUID, notification_id: UUID) -> ArgusNotification:
+        return ArgusNotification.get(receiver=receiver, id=notification_id)
+
+    def get_unread_count(self, receiver: UUID) -> int:
+        query = ArgusNotification.filter(
+            receiver=receiver, state__eq=ArgusNotificationState.UNREAD.value).allow_filtering().all()
+        return len(query)
+
+    def read_notification(self, receiver: UUID, notification_id: UUID) -> ArgusNotification:
+        notification = ArgusNotification.get(receiver=receiver, id=notification_id)
+        notification.state = ArgusNotificationState.READ
+        return bool(notification.save().state)
+
+    def get_notifications(self, receiver: UUID, limit: int = 20, after: UUID | None = None) -> list[ArgusNotification]:
+        if after:
+            return ArgusNotification.filter(receiver=receiver, id__lte=after).all().limit(limit)
+        return ArgusNotification.filter(receiver=receiver).all().limit(limit)
+
+
+class NotificationSenderBase:
     NOTIFICATION_TITLES = {
         ArgusNotificationTypes.Mention: "You were mentioned in a comment",
         ArgusNotificationTypes.StatusChange: "A run you are assigned to changed status",
@@ -18,12 +59,7 @@ class NotificationManagerService:
         ArgusNotificationTypes.ScheduleChange: "You were assigned to a schedule",
     }
 
-    CONTENT_TEMPLATES = {
-        ArgusNotificationTypes.Mention: lambda p: render_template("notifications/mention.html.j2", **p if p else {})
-    }
-
-    def __init__(self) -> None:
-        pass
+    CONTENT_TEMPLATES = {}
 
     def _check_user(self, user_id: UUID) -> bool:
         try:
@@ -45,8 +81,23 @@ class NotificationManagerService:
         raise NotificationManagerException(
             f"Content renderer for notification type {content_type} not found.", content_type)
 
+    def send_notification(self, receiver: UUID, sender: UUID, 
+                          notification_type: ArgusNotificationTypes,
+                          source_type: ArgusNotificationSourceTypes,
+                          source_id: UUID,
+                          title: str | None = None,
+                          content: str | None = None,
+                          content_params: dict | None = None) -> Any:
+        raise NotImplementedError()
+
+
+class ArgusDBNotificationSaver(NotificationSenderBase):
+    CONTENT_TEMPLATES = {
+        ArgusNotificationTypes.Mention: lambda p: render_template("notifications/mention.html.j2", **p if p else {})
+    }
+
     def send_notification(self, receiver: UUID, sender: UUID, notification_type: ArgusNotificationTypes, source_type: ArgusNotificationSourceTypes,
-                          source_id: UUID, title: str = None, content: str = None, content_params: dict = None) -> ArgusNotification:
+                          source_id: UUID, title: str | None = None, content: str | None = None, content_params: dict | None = None) -> ArgusNotification:
         new_notification = ArgusNotification()
         for user in [sender, receiver]:
             if not self._check_user(user_id=user):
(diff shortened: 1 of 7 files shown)


argus/backend/service/notification_manager.py
```diff
@@ -11,19 +13,53 @@ class NotificationManagerException(Exception):
 
 
 class NotificationManagerService:
+
+    def __init__(self) -> None:
+        self.notification_services = []
+        self.attach_notification_sender(ArgusDBNotificationSaver())
+        self.attach_notification_sender(EmailNotificationServiceSender())
+
+    def attach_notification_sender(self, service):
```
soyacz, line 22: this method is redundant, in `__init__` we could populate `self.notification_services` without it.
It is better to add minimum required public methods.

argus/backend/service/notification_manager.py
```diff
@@ -11,19 +13,53 @@ class NotificationManagerException(Exception):
 
 
 class NotificationManagerService:
+
+    def __init__(self) -> None:
+        self.notification_services = []
+        self.attach_notification_sender(ArgusDBNotificationSaver())
```
soyacz, line 19: Better solution would be passing list of notification senders in `NotificationManagerService` constructor, so we could decouple these classes from each other allowing simpler unit tests (by providing fake sender instead ArugsDB/Email senders)

argus/backend/service/notification_manager.py
```diff
@@ -11,19 +13,53 @@ class NotificationManagerException(Exception):
...
+
+    def get_notificaton(self, receiver: UUID, notification_id: UUID) -> ArgusNotification:
+        return ArgusNotification.get(receiver=receiver, id=notification_id)
+
+    def get_unread_count(self, receiver: UUID) -> int:
+        query = ArgusNotification.filter(
+            receiver=receiver, state__eq=ArgusNotificationState.UNREAD.value).allow_filtering().all()
+        return len(query)
+
+    def read_notification(self, receiver: UUID, notification_id: UUID) -> ArgusNotification:
```
soyacz, line 43: maybe rename it to `mark_notification_as_read`?

argus/backend/util/send_email.py
```diff
@@ -0,0 +1,76 @@
...
+        self.sender = ""
+        self._password = ""
+        self._user = ""
+        self._server_host = ""
+        self._server_port = ""
+        self._conn = None
+        self._retrieve_credentials()
+        self._connect()
+
+    def _retrieve_credentials(self):
```
soyacz, line 27: what is the purpose of this method?

soyacz (changes requested):
Besides some comments, the biggest problem are missing unit tests, making modifications in future more problematic/longer.
But due tight couple to db, looks it is hard to unit test. @k0machi is it possible to abstract argus repository to some class that could be replaced by dummy db (e.g. in memory)?

argus/backend/service/notification_manager.py
aleksbykov: it could be better, if NotificationManagerService created once, some kind of singelton. But now it created Each time when need send , read notification, get status

argus/backend/service/notification_manager.py
aleksbykov: @k0machi WDYT?

argus/backend/service/notification_manager.py
```diff
@@ -11,19 +13,53 @@ class NotificationManagerException(Exception):
...
+        notification = ArgusNotification.get(receiver=receiver, id=notification_id)
+        notification.state = ArgusNotificationState.READ
+        return bool(notification.save().state)
+
+    def get_notifications(self, receiver: UUID, limit: int = 20, after: UUID | None = None) -> list[ArgusNotification]:
```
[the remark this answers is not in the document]
aleksbykov: Both supported from 3.10, why do you think it is better?

argus/backend/util/send_email.py
aleksbykov: Now credentials are written in file it is not securely. In future Secure store could be added. So it will be easy to change the provider of credentials. 
May it could be removed right now

argus/backend/util/send_email.py
```diff
@@ -0,0 +1,76 @@
...
+        :param recipients: iterable, list of recipients
+        :param html: True/False
+        :param files: paths of the files that will be attached to the email
+        :return:
+        """
+        email = self.prepare_email(subject, content, recipients, html)
+        self.send_email(recipients, email)
+
+    def send_email(self, recipients, email):
+        self.conn.sendmail(self.sender, recipients, email)
```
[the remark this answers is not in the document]
aleksbykov: yep. will add

argus/backend/util/send_email.py
soyacz: yes, private methods can be added/modified without any issues in future. Better not to add upfront.

argus/backend/service/notification_manager.py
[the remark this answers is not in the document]
soyacz: if it is supported by 3.10 then it is just my preference. You can ignore this comment.

argus/backend/service/notification_manager.py
k0machi: Singletons would be dangerous to use in this context, as flask instantiates new context for each request, meaning using a singleton that lives outside this context could lead to bugs related to data leakage. Sending notifications senders in the constructor would indeed be better, you can make constructor instantiate default ones if None was passed instead of a list of senders

argus/backend/service/notification_manager.py
k0machi: I like `verb_noun` scheme more, but sure, that can be changed

argus/backend/service/notification_manager.py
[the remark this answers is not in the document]
k0machi: You should use native types where possible instead of aliases, they are more useful when doing reflection on function signatures, plus we're using 3.10+ anyway.

templates/notifications/email_mention.html.j2
```diff
@@ -0,0 +1,2 @@
+<div>You were mentioned by <span class='fw-bold'>@{{ username }}</span> in a discussion of <a href='{{url_for("main.runs", **{ "test_id": test_id, "additionalRuns[]": run_id })}}'>{{ build_id }}#{{ build_number }}</a>. </div>
```
k0machi, line 1: keep in mind that bootstrap css classes like `fw-bold` and such will not be rendered in the sent email, so this would need `style` attribute to embolden the text instead or full html template document

k0machi (changes requested):

k0machi:
> Besides some comments, the biggest problem are missing unit tests, making modifications in future more problematic/longer. But due tight couple to db, looks it is hard to unit test. @k0machi is it possible to abstract argus repository to some class that could be replaced by dummy db (e.g. in memory)?

In theory it should be possible if you patch `ScyllaCluster.get` method - this instantiates cql engine and cluster connection. The problem is the cqlengine itself, we need to patch its database interaction methods as they're entirely unexposed to the user, _it seems_.

argus/backend/service/notification_manager.py
soyacz: Maybe it is just me, but current name was not clear enough. E.g. read_notification could be understood as fetching it from db and showing to user, while this method just marks notification as read.

argus/backend/service/notification_manager.py
soyacz: but does `notification_manager` depends on flask context? Isn't it a separate service that works in background?

soyacz:
> > Besides some comments, the biggest problem are missing unit tests, making modifications in future more problematic/longer. But due tight couple to db, looks it is hard to unit test. @k0machi is it possible to abstract argus repository to some class that could be replaced by dummy db (e.g. in memory)?
> 
> In theory it should be possible if you patch `ScyllaCluster.get` class method - this instantiates cql engine and cluster connection. The problem is the cqlengine itself, we need to patch its database interaction methods as they're entirely unexposed to the user, _it seems_.

Just for the future reference because it would require some major refactor here: for this kind of problems there's Repository Pattern. It's often useful to use it for working with databases.

argus/backend/service/notification_manager.py
aleksbykov: @soyacz this method was not changed by this commit. Better to rename it in another one

argus/backend/service/notification_manager.py
k0machi: For now it only runs within flask context, but doesn't depend on it. Background services aren't implemented yet.

k0machi:
> > > Besides some comments, the biggest problem are missing unit tests, making modifications in future more problematic/longer. But due tight couple to db, looks it is hard to unit test. @k0machi is it possible to abstract argus repository to some class that could be replaced by dummy db (e.g. in memory)?
> > 
> > 
> > In theory it should be possible if you patch `ScyllaCluster.get` class method - this instantiates cql engine and cluster connection. The problem is the cqlengine itself, we need to patch its database interaction methods as they're entirely unexposed to the user, _it seems_.
> 
> Just for the future reference because it would require some major refactor here: for this kind of problems there's Repository Pattern. It's often useful to use it for working with databases.

In theory cqlengine is an implementation of a repository pattern, but it feels a bit simplistic compared to other ORMs I've encountered.

soyacz (approved):

roydahan:
@k0machi please re-review and if approved, you can merge.

templates/notifications/email_mention.html.j2
```diff
@@ -0,0 +1,2 @@
+<div>You were mentioned by <span style="font-weight:bold">@{{ username }}</span> in a discussion of <a href='{{url_for("main.runs", **{ "test_id": test_id, "additionalRuns[]": run_id })}}'>{{ build_id }}#{{ build_number }}</a>. </div>
```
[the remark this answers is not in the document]
aleksbykov: Yes, it generated not full url.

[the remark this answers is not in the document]
aleksbykov: @k0machi , Do we store somewhere the domain name?

k0machi (approved):

Issue #168, opened 2022-09-05
SMTP Connection opens on every notification related request


k0machi:
The `Email` sender class is too eager to establish an smtp connection, leading to rate limiting on the smtp server side - it should only connect if there's an email about to be sent, otherwise an SMTP connection is established and closed on every request related to notifications - such as getting unread counts.

Problem line: https://github.com/scylladb/argus/blob/master/argus/backend/util/send_email.py#L26

PR #169, opened 2022-09-05
fix(send_email): Init smtp connection for each send or for all

aleksbykov:
Add to constructor parameter which allow to create connection
one time during initializtion or each time befor send

fix issue #168


aleksbykov:
@k0machi , @soyacz Could you review, please ?

soyacz (approved):

k0machi (approved):

PR #173, opened 2022-09-15
fix(frontend): Increase polling times for stat and testrun refreshes

k0machi:
This should alleviate the load on the particularly heavy requests such as stats and remove needless refreshing on the workspace page.

### DIFF
--- a/frontend/ReleaseDashboard/TestDashboard.svelte
+++ b/frontend/ReleaseDashboard/TestDashboard.svelte
@@ -167,10 +167,10 @@
 
     onMount(() => {
         fetchStats();
-
+        let refreshInterval = 300 + 15 + Math.round((Math.random() * 10));
         statRefreshInterval = setInterval(() => {
             fetchStats();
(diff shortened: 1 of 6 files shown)


frontend/ReleaseDashboard/TestDashboard.svelte
```diff
@@ -167,10 +167,10 @@
 
     onMount(() => {
         fetchStats();
-
+        let refreshInterval = 300 + 15 + Math.round((Math.random() * 10));
```
roydahan, line 170: Why are you using random?
So the refreshes won't happen at the same time?

k0machi: On workspace however the refreshes would be staggered to yeah, prevent them from happening at the same time, as workspace does N requests for N releases and if a person has several tabs open with workspace this could potentially create very high load if they all refresh at the same time.

roydahan (approved):

PR #186, opened 2022-12-06
Add scylla-operator package/version information

fruch:
* add those to web run info
* add those to issue template

Ref: https://github.com/scylladb/scylla-cluster-tests/pull/5512

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Python 3.10
         uses: actions/setup-python@v4
         with:
-          python-version: 3.10.3
+          python-version: '3.10'
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


fruch:
@k0machi 

seems like the github actions is broken 
```
Version 3.10.3 was not found in the local cache
Error: Version 3.10.3 with arch x[6](https://github.com/scylladb/argus/actions/runs/3627446497/jobs/6117345629#step:3:7)4 not found
```

I think it's using an old version of get-python action

k0machi:
@fruch I think it just wants latest 3.10 hotfix, so just update the version to 3.10.8

fruch:
> @fruch I think it just wants latest 3.10 hotfix, so just update the version to 3.10.8

even better, we should be pining specific minor releases...

fruch:
> > @fruch I think it just wants latest 3.10 hotfix, so just update the version to 3.10.8
> 
> even better, we should be pining specific minor releases...

and now it's works much better.

vponomaryov (approved):

PR #189, opened 2022-12-07
improvement(ReleaseManager): Allow selecting plugin for a test

k0machi:
This adds an ability for the admin user to adjust the plugin used for the specific test inside Release Manager Admin Panel

### DIFF
--- a/argus/backend/service/release_manager.py
+++ b/argus/backend/service/release_manager.py
@@ -62,7 +62,7 @@ class ReleaseManagerService:
 
         return new_group
 
-    def create_test(self, test_name, pretty_name, build_id, build_url, group_id, release_id) -> ArgusTest:
+    def create_test(self, test_name, pretty_name, build_id, build_url, group_id, release_id, plugin_name) -> ArgusTest:
         release = ArgusRelease.get(id=UUID(release_id))
         group = ArgusGroup.get(id=UUID(group_id))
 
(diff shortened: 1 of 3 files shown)


frontend/AdminPanel/TestCreator.svelte
```diff
@@ -39,6 +40,16 @@
                     bind:value={newTest.test_name}
                 />
             </div>
+            <div class="form-group mb-2">
+                <label for="" class="form-label">Plugin</label>
+                <select id="" class="form-select" bind:value={newTest.plugin_name}>
+                    {#each Object.keys(AVAILABLE_PLUGINS) as plugin}
+                        <option value={plugin}
+                            >{plugin}</option
+                        >
```
aleksbykov, line 49: should be on one line?

aleksbykov (approved):
LGTM

frontend/AdminPanel/TestCreator.svelte
k0machi: For inline elements like that formatting it like this avoids inserting extra spaces into markup, the last `>` is there to line up with opening `<option>`

argus/backend/service/release_manager.py
```diff
@@ -62,7 +62,7 @@ def create_group(self, group_name: str, pretty_name: str, build_system_id: str,
 
         return new_group
 
-    def create_test(self, test_name, pretty_name, build_id, build_url, group_id, release_id) -> ArgusTest:
+    def create_test(self, test_name, pretty_name, build_id, build_url, group_id, release_id, plugin_name) -> ArgusTest:
```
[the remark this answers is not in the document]
k0machi: it's mandatory, but can be null, endpoints for creating/modifying tests send the whole entity so that's not a problem

roydahan (approved):

PR #190, opened 2022-12-07
improvement(TestRuns): Add plugin selector for unknown plugins

k0machi:
This allows user to adjust test with an unknown plugin to another plugin if user knows what plugin the test belongs to


fruch:
> This allows user to adjust test with an unknown plugin to another plugin if user knows what plugin the test belongs to

What do you mean by user ? from the UI ? setting it in the DB ?

k0machi:
> > This allows user to adjust test with an unknown plugin to another plugin if user knows what plugin the test belongs to
> 
> What do you mean by user ? from the UI ? setting it in the DB ?

It allows users to select framework used in the UI, like so:
image

fruch:
> > > This allows user to adjust test with an unknown plugin to another plugin if user knows what plugin the test belongs to
> > 
> > 
> > What do you mean by user ? from the UI ? setting it in the DB ?
> 
> It allows users to select framework used in the UI, like so: image
> 
> Second commit in this PR also adds a mechanism where a run submitted by a client will check the corresponding test entity for this field - and fill it if necessary, since client code knows which plugin is being used.

sound like something that should be happen on it's own, the user shouldn't be aware of the fact there a plugin system.
for example if I'm sending this to someone from releng/developer, they won't know what todo with.

we should have a  fallback, for example default to SCT plugin (since it's the first)

k0machi:
@fruch @roydahan Basically, seeing this window is the last resort - the idea with hinting from pipeline is possible and something we should implement soon, this PR brings another mechanism where a submitted run will automatically fixup the test entity if it's missing the `plugin_name` field (since it knows what it got submitted from). This selector would be there if both of those fail - user can manually adjust the entity this way if for example he wants to do it before a run is submitted. 

Alternatively, we could restrict this window to the Admin/Manager role, making it into a convenience function for the Admin user, so that he could do it right in the workspace instead of finding the entity in the Admin Panel.

k0machi:
@fruch @roydahan I've locked this functionality behind having admin role, so now only people with `ROLE_ADMIN` will see that message and be able to fix it up.

fruch (approved):
LGTM

aleksbykov (approved):

roydahan (approved):

PR #196, opened 2022-12-21
fix(team_manager_service.py): Fetch ALL teams for the team list

k0machi:


roydahan (approved):

fruch:
@k0machi care to explain a bit what's the issue ? 

And why we need to fetch the whole list of users here?

k0machi:
@fruch Idea is to see all the teams in the team list, so that people not on some teams can see them too, in case one team leader wants to see what's going on in another team

fruch (approved):

fgelcer (approved):
LGTM

PR #208, opened 2023-02-22
feature(scripts): Stuck runs remover

k0machi:
This change adds an endpoint to the TestRun API allowing a periodic cleanup
of "stuck", i.e. runs that have not had their heartbeat updated for more than
45 minutes. The endpoint call will set them as "ABORTED" in such case.

Task: scylladb/qa-tasks#940


k0machi:
> Could you please add a document instruction or script that will this "Stuck runs remover" by CRON periodically? (in this or separate PR)

Will add in a separate PR.

PR #220, opened 2023-03-20
fix(TestRuns.svelte): Fix missing plugin selector

k0machi:
If we can't retrieve runs due to missing the plugin name, we can't fix
the plugin name because the block for that is housed inside the runs
block, which needs runs to be drawn. This is now fixed


roydahan (approved):
FWIW I approve this...

PR #224, opened 2023-04-13
fix(GithubIssue.svelte): Correct route for test runs

k0machi:
This fix corrects an issue where a test run link on an aggregated issue
would use the old test_run endpoint, which has been deprecated and no
longer works as runs endpoint now requires a valid test_id in addition
to run_id.


fgelcer:
@k0machi , was it deployed already?

k0machi:
> @k0machi , was it deployed already?

Yes.

fgelcer:
thank you

PR #229, opened 2023-04-27
improvement(events tab): show nemesis that happened

temichus:
show which nemesis ran during each event
on events tab

Screenshot from 2023-04-27 11-21-18
Screenshot from 2023-04-27 11-21-09

### DIFF
--- a/frontend/TestRun/EventsTab.svelte
+++ b/frontend/TestRun/EventsTab.svelte
@@ -95,7 +95,7 @@
     /**
      * @param {{last_events: string[], event_amount: int, severity: string }[]} events
      */
-    const prepareEvents = function (events) {
+    const prepareEvents = function (events, nemesisData) {
         let flatEvents = events.reduce((acc, val) => {
             displayCategories[val.severity].totalEvents = val.event_amount;
             displayCategories[val.severity].eventsSubmitted = val.last_events.length;
@@ -117,11 +117,27 @@
             return parsed;
         });
 
+        let sortedNemesises = nemesisData.sort((a, b) => a.start_time - b.start_time);
+        let nemesisIndex = 0;
+        for (let event of parsedEvents.sort((a, b) => a.time - b.time)) {
+            event.nemesis = undefined;
+            while(sortedNemesises[nemesisIndex]) {
+                if (event.time.getTime() < sortedNemesises[nemesisIndex].start_time * 1000) {
+                    break;
+                } else if(event.time.getTime() <= sortedNemesises[nemesisIndex].end_time * 1000) {
+                    event.nemesis = sortedNemesises[nemesisIndex];
+                    break;
+                }
+                nemesisIndex++;
+            }
+
+        }
+
         return parsedEvents.sort((a, b) => b.time - a.time);
     };
 
     onMount(() => {
-        parsedEvents = prepareEvents(testRun?.events ?? []);
+        parsedEvents = prepareEvents(testRun?.events ?? [], testRun?.nemesis_data ?? []);
     });
 </script>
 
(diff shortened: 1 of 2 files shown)


frontend/TestRun/EventsTab.svelte
```diff
@@ -117,11 +117,29 @@
             return parsed;
         });
 
+        let sortedNemesises = nemesisData.sort((a, b) => a.start_time - b.start_time);
+        let nemesisIndex = 0;
+        for (let event of parsedEvents.sort((a, b) => a.time - b.time)) {
+            event.nemesis = undefined;
+            if ( sortedNemesises[nemesisIndex]) {
+                while(sortedNemesises[nemesisIndex]){
```
aleksbykov, line 125: Do you need outer if statement? if sortedNemesises[nemesisInded] is false whille loop also will not run

aleksbykov (approved):
LGTM, Small question

frontend/TestRun/EventsTab.svelte
```diff
@@ -117,11 +117,29 @@
             return parsed;
         });
 
+        let sortedNemesises = nemesisData.sort((a, b) => a.start_time - b.start_time);
+        let nemesisIndex = 0;
+        for (let event of parsedEvents.sort((a, b) => a.time - b.time)) {
+            event.nemesis = undefined;
+                while(sortedNemesises[nemesisIndex]){
```
k0machi, line 124: You should correct the indentation of the while block and also fix spaces before `{` and after `}` characters (there should be a space in every case, e.g. `while (true) {`)

frontend/TestRun/EventsTab.svelte
```diff
@@ -117,11 +117,29 @@
...
+                    }else{
+                        nemesisIndex += 1;
+                    }
+                }
+
+        }
+
         return parsedEvents.sort((a, b) => b.time - a.time);
     };
+    0
```
k0machi, line 139: Looks like a typo here

frontend/TestRun/EventsTab.svelte
```diff
@@ -117,11 +117,29 @@
...
+        let nemesisIndex = 0;
+        for (let event of parsedEvents.sort((a, b) => a.time - b.time)) {
+            event.nemesis = undefined;
+                while(sortedNemesises[nemesisIndex]){
+                    if (event.time.getTime() < sortedNemesises[nemesisIndex].start_time * 1000) {
+                        break;
+                    }else if(event.time.getTime() <= sortedNemesises[nemesisIndex].end_time * 1000){
+                        event.nemesis = sortedNemesises[nemesisIndex];
+                        break;
+                    }else{
```
k0machi, line 130: The `else` not needed as we break out in every other conditional, can just `nemesisIndex++` here.

frontend/TestRun/StructuredEvent.svelte
```diff
@@ -27,8 +27,13 @@
 >
     <div class="event-header d-flex align-items-start flex-wrap">
         <div class="ms-2 mb-2 bg-dark text-light rounded px-2">{event.eventType}</div>
-        <div class="ms-2 mb-2 me-auto rounded px-2 severity-{event.severity.toLowerCase()}">{event.severity}</div>
-        <div class="ms-2 mb-2 rounded px-2">
+        <div class="ms-2 mb-2 rounded px-2 severity-{event.severity.toLowerCase()}">{event.severity}</div>
+        {#if event.nemesis}
+        <div class="ms-2 mb-2 rounded px-2 status-{event.nemesis.status.toLowerCase()}">{event.nemesis.name}</div>
```
k0machi, line 32: The `div`s inside conditional blocks should be embedded.

k0machi (changes requested):
Indendation for the while block is now offset, and some small comments about style.

roydahan (approved):

PR #243, opened 2023-05-23
fix(driver_matrix_tests): Do not alter scylla_version

k0machi:
Make scylla-version field as in to sync up with other types of runs we
have (previously we had been doing replacements in sct runs)


fruch (approved):
LGTM

PR #247, opened 2023-05-24
improvement(*TestRun): Allow viewing all issues relevant to the test

k0machi:
This change adjust the Issues tab to display both issues attached to
this particular run and a collapse of all issues associated with the
run's test.

Task: scylladb/qa-tasks#288


k0machi:
Demo:
image

image

PR #284, opened 2023-08-24
fix(argus.client): Add a pattern for CPP driver results

k0machi:
This fixes an issue where a C++ driver matrix test is always marked as
failed due to the filename regex not being compatible and ending up with
driver_name being null, where as at least one datastax and one scylla
result is required.

Task: scylladb/qa-tasks#1435

### DIFF
--- a/argus/client/driver_matrix_tests/client.py
+++ b/argus/client/driver_matrix_tests/client.py
@@ -117,8 +117,11 @@ class ArgusDriverMatrixClient(ArgusClient):
 
         return raw_cases
 
-    def get_driver_info(self, xml_name: str) -> dict[str, str]:
-        filename_re = r"(?P<name>[\w]*)\.(?P<driver_name>[\w]*)\.(?P<proto>v\d)\.(?P<version>[\d\.]*)\.xml"
+    def get_driver_info(self, xml_name: str, test_type: TestTypeType) -> dict[str, str]:
+        if test_type == "cpp":
+            filename_re = r"TEST-(?P<driver_name>[\w]*)-(?P<version>[\d\.]*)\.xml"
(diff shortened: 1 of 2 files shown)


argus/client/driver_matrix_tests/client.py
```diff
@@ -117,8 +117,11 @@ def get_test_cases(self, cases: list[ElementTree.Element]) -> list[RawMatrixTest
...
 
-    def get_driver_info(self, xml_name: str) -> dict[str, str]:
-        filename_re = r"(?P<name>[\w]*)\.(?P<driver_name>[\w]*)\.(?P<proto>v\d)\.(?P<version>[\d\.]*)\.xml"
+    def get_driver_info(self, xml_name: str, test_type: TestTypeType) -> dict[str, str]:
+        if test_type == "cpp":
+            filename_re = r"TEST-(?P<driver_name>[\w]*)-(?P<version>[\d\.]*)\.xml"
+        else:
+            filename_re = r"(?P<name>[\w]*)\.(?P<driver_name>[\w]*)\.(?P<proto>v\d)\.(?P<version>[\d\.]*)\.xml"
 
         match = re.match(filename_re, xml_name)
```
temichus, line 126: Now function will have different return for cpp case. Is it expected?

aleksbykov: It will not affect, because used only <driver_name>

aleksbykov: It shouldn't, because used only <driver_name> which is defined in both returned results.

aleksbykov (approved):
LGTM

temichus (approved):

PR #287, opened 2023-10-16
fix(argus_service.py): Do not use IN query for querying unassigned runs

k0machi:
This fix corrects an issue where a large group of tests would cause a
query error due to exceeding the size of available cartesian product for
the query. Additionally this corrects an issue where deleting a schedule
without an assignee would be impossible.

### DIFF
--- a/argus/backend/plugins/core.py
+++ b/argus/backend/plugins/core.py
@@ -27,7 +27,7 @@ class PluginModelBase(Model):
     _plugin_name = "unknown"
     # Metadata
     build_id = columns.Text(required=True, partition_key=True)
-    start_time = columns.DateTime(required=True, primary_key=True, clustering_order="DESC", default=datetime.now)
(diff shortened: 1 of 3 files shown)


argus/backend/plugins/core.py
```diff
@@ -27,7 +27,7 @@ class PluginModelBase(Model):
     _plugin_name = "unknown"
     # Metadata
     build_id = columns.Text(required=True, partition_key=True)
-    start_time = columns.DateTime(required=True, primary_key=True, clustering_order="DESC", default=datetime.now)
+    start_time = columns.DateTime(required=True, primary_key=True, clustering_order="DESC", default=datetime.now, custom_index=True)
```
fruch, line 30: Does this add the index ? Or what this means ?

fruch (approved):

argus/backend/plugins/core.py
k0machi: This is probably an issue with CQL Engine - there's no index, I can execute this query through CQLsh just fine without Allow Filtering, but cqlengine complains while validating this query.

PR #294, opened 2023-11-02
fix(models/web): Make default release non-perpetual

k0machi:
This change changes default release type to "Static", which is used for
scheduling purposes to determine if release needs weekly schedules (like
scylla-master for example) or endless static schedules like point
releases.


fruch (approved):
LGTM

fgelcer:
got it, thank you @k0machi

PR #308, opened 2023-11-27
improvement(TestRuns.svelte): Add jenkins job url to system messages.

### LINKED ISSUE #307: Usability enhancement  - Add ability to open a "Not Run" job - to show a link to its Jenkins job
At the moment when you click "Investigate" on a "Not Run" job, you get the screen to "Select a layout", but it won't show anything.
The enhancement here is to display the link to the Jenkins job (we should already have this information), so one can click it and run that specific job from Jenkins.

In terms of implementation (as much as I understand it), maybe we can set it as the default layout for any job and when we have results for the relevant layout, it will override it.

In terms of UI, the simplest UI to just show the link to the jenkins job.

k0machi:
This change adds a jenkins url to the system messages view (any state
when there are no runs available to display), simplifying finding job in
the CI.

Fixes #307

### DIFF
--- a/frontend/WorkArea/TestRuns.svelte
+++ b/frontend/WorkArea/TestRuns.svelte
@@ -359,7 +359,11 @@
                 {/each}
             </div>
         {:else}
-            <TestRunsMessage state={stateMap[currentState]} />
+            <TestRunsMessage state={stateMap[currentState]}>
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

Issue #302, opened 2023-11-22
Please make scylla-cluster-tests git sha version available in the logs and the Details tab


bhalevy:
I wanted to understand a verification issue SCT reported (https://github.com/scylladb/scylla-cluster-tests/issues/6840) but it was hard to track the logic (from the backtraces) without knowing which version was used.

Please add it to argus.

fruch:
we probably need those
- fork
- branch
- sha

k0machi:
Commit SHA of SCT is already being submitted, just not displayed, I'll add a field for that.

fruch:
@k0machi is this done ?

fruch:
> > @k0machi is this done ?
> 
> Yes, unless you want me to implement the branch and origin tracking via this issue

let complete those 2 as well

k0machi:
https://github.com/scylladb/scylla-cluster-tests/pull/6863

PR #309, opened 2023-11-28
feature(plugins/sct): Add fields for origin and branch names

k0machi:
This change adds fields for tracking sct repository and branch used for
a test run.

Task: #302


k0machi:
> LGTM
> 
> Now we need to update the client in SCT ?

Yes

Issue #297, opened 2023-11-06
Argus login - degraded mode or OAuth with Github App auth


k0machi:
Since Github Apps potentially provide finer permission control than normal OAuth flow we should switch to it to reduce the power of tokens granted to Argus.

roydahan:
@fruch / @k0machi let's start by estimating what's the quickest / lowest effort approach.
E.g. having a degraded mode for people who didn't authorize Argus, so they won't be able to see the issue status.
Or the full solution through Github Apps or whatever.

k0machi:
@roydahan If the github app flow is largely the same it should be a quick fix - I'll provide the estimation soon.

k0machi:
Okay, so the flow is a bit different - it would require changing authentication/token rotation logic. However moving to a github app will allow finer grained permissions (org only with issue/pr read access only to specific repositories - something that isn't possible with OAuth flow). 

For a quick fix we can implement a checkbox on the login page that says "Grant access to private repositories" and a hint that says that this will grant access to ALL your repositories.

fruch:
> Okay, so the flow is a bit different - it would require changing authentication/token rotation logic. However moving to a github app will allow finer grained permissions (org only with issue/pr read access only to specific repositories - something that isn't possible with OAuth flow).
> 
> For a quick fix we can implement a checkbox on the login page that says "Grant access to private repositories" and a hint that says that this will grant access to ALL your repositories.

we need estimate for both things, to figure out which is first

k0machi:
> we need estimate for both things, to figure out which is first

Checkbox is quick, can do that in a day, The second one would take a week at least.

k0machi:
Looking at the "degraded" mode it seems that everything is functioning as is... I'm talking to Boris so we can move the argus app from my account to scylla and that might cover issues for private repos (which are apparently currently bork regardless)

k0machi:
https://github.com/scylladb/argus/pull/311

k0machi:
Issue panel when using the degraded experience:



![Image](https://github.com/scylladb/argus/assets/7761415/1f4a5275-952d-47b0-9272-441ac7149e73)

fruch:
Schedule with boris to transfer the argus application

k0machi:
Sent an email to @BorishuV regarding transfer of the application.

k0machi:
Transfer done, also created test github app to investigate whether or not we should switch to that and also removed old argus-testers group as we don't need it anymore.

[...]

PR #311, opened 2023-11-29
feature(Login.svelte): Allow disabling `repo` scope for Github login

k0machi:
This change allows user to decline giving full repository read/write
permissions (as this is needed to read issue states inside organization
private repositories) via a checkbox on the login page. The state of the
checkbox is saved into local storage to prevent user from having to
constantly check it. Additionally, this commit moves the login page out
of a Jinja template and into a Svelte component.

Task: #297


fruch:
@k0machi would be nice to see how the login page now shows with it.

PR #321, opened 2023-12-21
fix(client_service.py): Mark Passed runs as Investigated automatically

### LINKED ISSUE #317: PASSED tests are marked as 'Not investigated'
See https://argus.scylladb.com/workspace?state=WyJiMmQyMzc4ZC05OTE0LTRjYjUtOTJkMy0wZDM5NzNiNTk2YTAiXQ for example. It makes no sense - we don't trust the results?
Please make sure those are marked as investigated, or can be ignored when looking if all runs of a test were looked at or not.

k0machi:
This change adjusts the run status logic to automatically mark any run
that passed as investigated, as there's nothing to investigate.

Fixes #317


fruch (approved):
Lgtm

roydahan (changes requested):
I disagree with the suggested behavior.

mykaul:
> I disagree with the suggested behavior.

Disagreeing is fine - please suggest an alternative though.

roydahan:
I wrote in the issue

k0machi:
Closing as this is not needed.

PR #331, opened 2024-01-29
feature(ReleaseDashboard.svelte): Improved version filtering

### LINKED ISSUE #315: 5.4 dashboard shows additional builds - 5.5.0-dev, 5.2.10, 5.2.11
https://argus.scylladb.com/dashboard/scylla-5.4 shows:

image

k0machi:
This change allows administrators to define release's version regular
expression, which filters out versions displayed on the Release
Dashboard. For this purpose, a new menu was introduced in admin panel to
edit the release - setting the version regex makes the version bar on
the dashboard only display matching versions, for example: setting it to
"5.4(.*)" would only display versions starting with "5.4". This is
useful to filter out dev builds, erroneous submissions, missing versions
and rolling upgrades. Additionally, the behaviour of fetching "No
Version" runs was changed - a new checkbox was added to the dashboard
allowing user to set preference whether or not to fetch runs containing
no version alongside any versioned run. In case the checkbox is not
checked, there is a filter option to filter down to "No Version". The
regex behaviour also receives its own checkbox, allowing user to set
preference whether or not to filter out extra versions from his
dashboard. Lastly, a button to delete a release was added alongside
general release editing window.

Fixes #315

### DIFF
--- a/argus/backend/controller/api.py
+++ b/argus/backend/controller/api.py
@@ -389,9 +389,10 @@ def release_stats_v2():
     release = request.args.get("release")
     limited = bool(int(request.args.get("limited")))
     version = request.args.get("productVersion", None)
+    include_no_version = bool(int(request.args.get("includeNoVersion", True)))
     force = bool(int(request.args.get("force")))
     stats = ReleaseStatsCollector(
-        release_name=release, release_version=version).collect(limited=limited, force=force)
+        release_name=release, release_version=version).collect(limited=limited, force=force, include_no_version=include_no_version)
 
     res = jsonify({
         "status": "ok",
--- a/argus/backend/controller/admin_api.py
+++ b/argus/backend/controller/admin_api.py
@@ -6,7 +6,7 @@ from flask import (
     Request,
 )
 from argus.backend.error_handlers import handle_api_exception
-from argus.backend.service.release_manager import ReleaseManagerService
+from argus.backend.service.release_manager import ReleaseEditPayload, ReleaseManagerService
 from argus.backend.service.user import api_login_required, check_roles
 from argus.backend.models.web import UserRoles
 
@@ -88,6 +88,36 @@ def set_release_dormancy():
         "status": "ok",
         "response": {
             "updated": result
+        }
+    }
+
+
+@bp.route("/release/edit", methods=["POST"])
+@check_roles(UserRoles.Admin)
+@api_login_required
+def edit_release():
+    payload: ReleaseEditPayload = get_payload(request)
+    result = ReleaseManagerService().edit_release(payload)
+
+    return {
+        "status": "ok",
+        "response": {
+            "updated": result
+        }
+    }
+
+
+@bp.route("/release/delete", methods=["POST"])
+@check_roles(UserRoles.Admin)
+@api_login_required
+def delete_release():
+    payload = get_payload(request)
+    result = ReleaseManagerService().delete_release(release_id=payload["releaseId"])
+
+    return {
+        "status": "ok",
+        "response": {
+            "deleted": result
         }
     }
 
--- a/argus/backend/models/web.py
+++ b/argus/backend/models/web.py
@@ -111,6 +111,7 @@ class ArgusRelease(Model):
     pretty_name = columns.Text()
     description = columns.Text()
     github_repo_url = columns.Text()
+    valid_version_regex = columns.Text()
     assignee = columns.List(value_type=columns.UUID)
     picture_id = columns.UUID()
     enabled = columns.Boolean(default=lambda: True)
--- a/argus/backend/service/release_manager.py
+++ b/argus/backend/service/release_manager.py
@@ -1,4 +1,5 @@
 import logging
+from typing import TypedDict
 from uuid import UUID
 from argus.backend.db import ScyllaCluster
 from argus.backend.plugins.sct.testrun import SCTTestRun
@@ -9,6 +10,16 @@ LOGGER = logging.getLogger(__name__)
 
 class ReleaseManagerException(Exception):
     pass
+
+
+class ReleaseEditPayload(TypedDict):
+    id: str
+    pretty_name: str
+    description: str
+    valid_version_regex: str | None
+    enabled: bool
+    perpetual: bool
+    dormant: bool
 
 
 class ReleaseManagerService:
@@ -154,6 +165,33 @@ class ReleaseManagerService:
         release.perpetual = perpetual
         release.save()
 
+        return True
+
+    def edit_release(self, payload: ReleaseEditPayload) -> bool:
+
+        release: ArgusRelease = ArgusRelease.get(id=payload["id"])
+        release.pretty_name = payload["pretty_name"]
+        release.perpetual = payload["perpetual"]
+        release.enabled = payload["enabled"]
+        release.dormant = payload["dormant"]
+        release.description = payload["description"]
+        release.valid_version_regex = payload["valid_version_regex"]
+
+
+        release.save()
+        return True
+
+    def delete_release(self, release_id: str) -> bool:
+
+        release: ArgusRelease = ArgusRelease.get(id=release_id)
+
+        release_groups = ArgusGroup.filter(release_id=release.id)
+        release_tests = ArgusTest.filter(release_id=release.id)
+
+        for entity in [*release_groups.all(), *release_tests.all()]:
+            entity.delete()
+
+        release.delete()
         return True
 
     def batch_move_tests(self, new_group_id: str, tests: list[str]) -> bool:
--- a/argus/backend/service/stats.py
+++ b/argus/backend/service/stats.py
@@ -343,7 +343,7 @@ class ReleaseStatsCollector:
         self.release_name = release_name
         self.release_version = release_version
 
-    def collect(self, limited=False, force=False) -> dict:
+    def collect(self, limited=False, force=False, include_no_version=False) -> dict:
         self.release: ArgusRelease = ArgusRelease.get(name=self.release_name)
         self.release_rows = [row for plugin in all_plugin_models()
                              for row in plugin.get_stats_for_release(release=self.release)]
@@ -353,8 +353,18 @@ class ReleaseStatsCollector:
             }
 
         if self.release_version:
-            self.release_rows = list(
-                filter(lambda row: row["scylla_version"] == self.release_version or not row["scylla_version"], self.release_rows))
+            if include_no_version:
+                expr = lambda row: row["scylla_version"] == self.release_version or not row["scylla_version"] 
+            elif self.release_version == "!noVersion":
+                expr = lambda row: not row["scylla_version"]
+            else:
+                expr = lambda row: row["scylla_version"] == self.release_version
+        else:
+            if include_no_version:
+                expr = lambda row: row
+            else:
+                expr = lambda row: row["scylla_version"]
+        self.release_rows = list(filter(expr, self.release_rows))
 
         self.release_stats = ReleaseStats(release=self.release)
         self.release_stats.collect(rows=self.release_rows, limited=limited, force=force)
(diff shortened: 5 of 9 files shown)


k0machi:
Demo:

image

image

image

image

image

image

mykaul:
Can I select all 4.6 versions? All RCs + GA for example?

roydahan:
> Can I select all 4.6 versions? All RCs + GA for example?

What's the use case?
When do you want to do it or what do you want to leave out?

argus/backend/controller/api.py
```diff
@@ -389,9 +389,10 @@ def release_stats_v2():
     release = request.args.get("release")
     limited = bool(int(request.args.get("limited")))
     version = request.args.get("productVersion", None)
+    include_no_version = bool(int(request.args.get("includeNoVersion", False)))
```
roydahan, line 392: I think that the default should be "True".
The more conservative approach is the default, if one wants to filter it out, he can later.

roydahan:
@aleksbykov there is one use case I'm afraid of, which I couldn't understand from the code what will happen.

Let's say I'm looking at 2024.1 release, and I have a filter in admin panel to show only 2024.1*.
Then, as part of tolling upgardes, I have a failed run from 2023.1 that don't have the "upgraded_version": 2024.1 (it failed before).
So, it should fall on the "no_versions", but wouldn't the regex filter going to filter it out?

fruch:
> @aleksbykov there is one use case I'm afraid of, which I couldn't understand from the code what will happen.
> 
> Let's say I'm looking at 2024.1 release, and I have a filter in admin panel to show only 2024.1*.
> Then, as part of tolling upgardes, I have a failed run from 2023.1 that don't have the "upgraded_version": 2024.1 (it failed before).
> So, it should fall on the "no_versions", but wouldn't the regex filter going to filter it out?
> 

It has a version, 2023.1, and it would be filtered by the regex, but with no filter applied you would see it.

What I suggest, to have a button with + or ... at the end that you can expend the versions that doesn't match the regex

To really address the issue of rolling upgrades you mention, you need to report the target version at the beginning of the test in SCT.

roydahan:
IIRC we said that the dashboard shouldn't filter or show above "versions" that are "base version" but only "upgraded version" or "scylla version".

k0machi:
Updated: 
image
image
image

argus/backend/controller/api.py
k0machi: Done

roydahan:
> @aleksbykov there is one use case I'm afraid of, which I couldn't understand from the code what will happen.
> 
> Let's say I'm looking at 2024.1 release, and I have a filter in admin panel to show only 2024.1*. Then, as part of tolling upgardes, I have a failed run from 2023.1 that don't have the "upgraded_version": 2024.1 (it failed before). So, it should fall on the "no_versions", but wouldn't the regex filter going to filter it out?

@k0machi can you please refer to that?
Under which versions a failed to upgrade falls into?

k0machi:
> > @aleksbykov there is one use case I'm afraid of, which I couldn't understand from the code what will happen.
> > Let's say I'm looking at 2024.1 release, and I have a filter in admin panel to show only 2024.1*. Then, as part of tolling upgardes, I have a failed run from 2023.1 that don't have the "upgraded_version": 2024.1 (it failed before). So, it should fall on the "no_versions", but wouldn't the regex filter going to filter it out?
> 
> @k0machi can you please refer to that? Under which versions a failed to upgrade falls into?

Provided it succesfully installed a base version it will show up as that, e.g. 5.3.0, otherwise yes, it will have `No version`

k0machi:
> > > @aleksbykov there is one use case I'm afraid of, which I couldn't understand from the code what will happen.
> > > Let's say I'm looking at 2024.1 release, and I have a filter in admin panel to show only 2024.1*. Then, as part of tolling upgardes, I have a failed run from 2023.1 that don't have the "upgraded_version": 2024.1 (it failed before). So, it should fall on the "no_versions", but wouldn't the regex filter going to filter it out?
> > 
> > 
> > @k0machi can you please refer to that? Under which versions a failed to upgrade falls into?
> 
> Provided it succesfully installed a base version it will show up as that, e.g. 5.3.0, otherwise yes, it will have `No version`. In this case the regex will filter it out from possible buttons to press until you press the `+` button

roydahan:
> Provided it succesfully installed a base version it will show up as that, e.g. 5.3.0, otherwise yes, it will have `No version`. The regex will hide the 5.3.0 button if it's set to `5.4(.*)`, until user presses the `+` button, while `No version` button (and the checkbox) is shown at all times.

Doesn't have to be for this PR, but can we change this behavior?
I think this is the root cause of all the workarounds we're trying to apply here.
If we wouldn't include "base versions" but only "Scylla Version" or "Upgraded Versions", we won't have any problem.
Failures, will fall under "No version" (we can try also workaround it from parsing the new_repo parameter).

fruch:
> > Provided it succesfully installed a base version it will show up as that, e.g. 5.3.0, otherwise yes, it will have `No version`. The regex will hide the 5.3.0 button if it's set to `5.4(.*)`, until user presses the `+` button, while `No version` button (and the checkbox) is shown at all times.
> 
> Doesn't have to be for this PR, but can we change this behavior? I think this is the root cause of all the workarounds we're trying to apply here. If we wouldn't include "base versions" but only "Scylla Version" or "Upgraded Versions", we won't have any problem. Failures, will fall under "No version" (we can try also workaround it from parsing the new_repo parameter).

Yes, me and @k0machi discussed it, and we think we'll leave the ..., with the extra versions for now, and remove the checkbox for adding no_version into filter.

next phase would be to change SCT to report versions and target version (based on the logic we had for enabling KMS), as soon as it can, to minimize the option of not having a version.

end result, we should have very few with NO VERSION, only cases of pipeline failures before tests starts, and the rest would have a version or a target version from the start of the test.

one could filter only no version or ALL to see those failures, all other filter won't show them.

roydahan:
> with the extra versions for now, and remove the checkbox for adding no_version into filter.

But this was the intention of this entire PR.
The main problem right now is that you can't understand what are the immediate failures to look at when you filter a patch release.
The checkbox must stay even for long term when we have the root cause solved.

fruch:
> > with the extra versions for now, and remove the checkbox for adding no_version into filter.
> 
> But this was the intention of this entire PR.
> The main problem right now is that you can't understand what are the immediate failures to look at when you filter a patch release.
> The checkbox must stay even for long term when we have the root cause solved.

I don't understand the use case, if one works on a patch release, looking at all should show him exactly what one needs
After a patch release is done, why would any would want or need to see those failures ? It's a look of fuss for nothing we ever needed

roydahan:
> I don't understand the use case, if one works on a patch release, looking at all should show him exactly what one needs
> After a patch release is done, why would any would want or need to see those failures ? It's a look of fuss for nothing we ever needed

This is not what I meant.
When you look at a patch release, you want to see what exactly passed / failed / not run.
There are 3 scenarios:
1. Simplest - failure that already published the scylla-version and we see the status when filtering.
2. Failure that happened VERY early in the job stage (e.g failed to checkout) - when filtering, this job should be displayed as "Not Run" (and probably this is the current case if the job exist in Argus).
3. Failure that happened BEFORE we have "Scylla-version" or with rolling upgrades before we have the "Target Version".
In this case, the failure is related to the patch release we're trying to release, hence we should see this failure when filtering by that patch release.

There is one gap that I'm not sure we currently handling, If the job was later re-run and passed for that patch release, it should be covered with that later status, so even if the "No version" is selected, the newest status for that release should "Win".

@aleksbykov can you please comment on what's working now and what's missing.

roydahan:
> I still don't get when one want to look on a previous release, and care for the failures. 99% of the time, you are looking/caring about the currently ongoing release and "ALL" would give you the information you care for on current status.
> 
> what is the use case I need to go back to 2023.1.2, and still care about the failed jobs ?

Where did I mention looking on old release?
I'm talking about working on a release that needs to be released and ALL doesn't give you this information because it takes the last status that it has, not caring on which release it ran.
So, if one want to see that all jobs passed for release X in order to release it, he cannot do it with ALL.

roydahan:
@k0machi what's the status of this one?
Did you understand the required behavior?
I showed the relevant use cases to @fruch, If still unclear I can show you as well.

fruch:
> @k0machi what's the status of this one? Did you understand the required behavior? I showed the relevant use cases to @fruch, If still unclear I can show you as well.

we did discuss it on monday

mainly that the "No Version" button would be able to toggle on it's own, and would be "AND" with the filter selected.
the rest should stay as implemented in this PR.

and we'll continue the work from SCT to supply version/target_version as soon as the test starts

aleksbykov:
> > I don't understand the use case, if one works on a patch release, looking at all should show him exactly what one needs
> > After a patch release is done, why would any would want or need to see those failures ? It's a look of fuss for nothing we ever needed
> 
> This is not what I meant. When you look at a patch release, you want to see what exactly passed / failed / not run. There are 3 scenarios:
> 
> 1. Simplest - failure that already published the scylla-version and we see the status when filtering.

This works. when on dashboard patch release version has been chosen, all jobs with this version are displayed. Here i am talking not about admin panel, but dashboard

> 2. Failure that happened VERY early in the job stage (e.g failed to checkout) - when filtering, this job should be displayed as "Not Run" (and probably this is the current case if the job exist in Argus).

if i understood the case correctly, on dashboard when patch release version chosen(filtered), and jenkins job failed on early stage, the job on dashboard displays status from previous run(and often it could previous patch version). So you need to go to jenkins and check why job was not started.

> 3. Failure that happened BEFORE we have "Scylla-version" or with rolling upgrades before we have the "Target Version".
>    In this case, the failure is related to the patch release we're trying to release, hence we should see this failure when filtering by that patch release.

for rolling-upgrade it is true, rolling upgrade jobs for different base version have same run id and if at least one rolling upgrade job report targe version, then even others don't report target version, all jobs will be under patch release version . But if all 3 rolling-upgrade jobs failed to report target version, then dashboard shows status from previous runs or patch release
> 
> There is one gap that I'm not sure we currently handling, If the job was later re-run and passed for that patch release, it should be covered with that later status, so even if the "No version" is selected, the newest status for that release should "Win".
> 
> @aleksbykov can you please comment on what's working now and what's missing.

I agree, that as soon we report target version, then most cases will be covered

k0machi:
@roydahan @fruch Yes, the no version itself will be a toggle, instead of being a separate button.

@aleksbykov Then we should start submitting target version as the initial version for argus, this way we will have *a* version as soon as possible and `no version` will gradually be relegated to runs that didn't even start.

k0machi:
With no version changes implemented:
https://github.com/scylladb/argus/assets/7761415/39d0ed32-55ad-4a95-9584-b3144ef82800

k0machi:
> > With no version changes implemented:
> > chrome_Ws4jYDXqV8.mp4
> 
> it's hard to tell the behavior from this video, it's confusing that you click on `No Version` and there are green passing jobs appearing.
> 
> maybe it we could use more recent version like 5.4 or 2024.1 data, we we know what we should be seeing.

Well, it's just to indicate that filter works in that demo, i.e. you're getting more runs when you click `no version`, I believe 4.6 data I have has no_version for most of the green artifact tests, I can switch to this branch on actual instance temporarily to let you test it yourself

roydahan:
I can’t really tell from the video if it works as expected.
It’s also confusing that it’s another button of filter that looks the same but this one toggles.
I thought we’re adding a checkbox that is checked by default.

k0machi:
@roydahan 
How about this kind of selector then?
image
image

roydahan:
It doesn't really matter at this point. merge whatever works for now.
We will need to work on how to improve the dashboard usability completely. 
BTW, the fact that the selector is available only when there is a filter is wrong.
Sometimes I would like to filter all failed runs with "no version". 
It's even the common case for the no-version, once I filter the other way around.

k0machi:
> It doesn't really matter at this point. merge whatever works for now. We will need to work on how to improve the dashboard usability completely. BTW, the fact that the selector is available only when there is a filter is wrong. Sometimes I would like to filter all failed runs with "no version". It's even the common case for the no-version, once I filter the other way around.

Okay, I've adjusted the behaviour to also be able to select noversion for `all` versions:

image

PR #333, opened 2024-01-29
improvement(plugins/sct): Store DC Name in CloudInstanceDetails

### LINKED ISSUE #222: Please, add column "DC name" to resources list ("Resources" tab)
Please, add column "DC name" to resources list ("Resources" tab).
We need it for multi DC tests - to know in which DC the instance is located.

Example:
https://argus.scylladb.com/test/1158bb3b-c12c-4c8c-a38f-d4492c398a4d/runs?additionalRuns[]=c4c3ee91-c03b-4dc9-9df9-4e1f42885db5

k0machi:
This change allows Argus to track datacenter names inside
CloudInstanceDetails.

Fixes #222

### DIFF
--- a/argus/client/sct/client.py
+++ b/argus/client/sct/client.py
@@ -148,7 +148,7 @@ class ArgusSCTClient(ArgusClient):
         self.check_response(response)
 
     def create_resource(self, name: str, resource_type: str, public_ip: str, private_ip: str,
-                        region: str, provider: str, shards_amount: int, state=ResourceState.RUNNING) -> None:
+                        region: str, provider: str, dc_name: str, rack_name: str, shards_amount: int, state=ResourceState.RUNNING) -> None:
         """
             Creates a cloud resource record in argus.
         """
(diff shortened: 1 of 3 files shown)


argus/client/sct/client.py
```diff
@@ -148,7 +148,7 @@ def submit_performance_results(self, performance_results: PerformanceResultsRequ
         self.check_response(response)
 
     def create_resource(self, name: str, resource_type: str, public_ip: str, private_ip: str,
-                        region: str, provider: str, shards_amount: int, state=ResourceState.RUNNING) -> None:
+                        region: str, provider: str, dc_name: str, shards_amount: int, state=ResourceState.RUNNING) -> None:
```
fruch, line 151: if we have region, that's in most cases 1to1 with dc_name, az would also be useful

fruch: in most cases it's the same:
region <-> datacenter
availability zone <-> rack

but then we have fake racks, and soon also fake regions
so datacenter and rack, and we need precise about where we get the information from on SCT end.

k0machi:
image

argus/client/sct/client.py
k0machi: Done

PR #336, opened 2024-02-09
fix(stats.py): Collect worst case comments and investigation

### LINKED ISSUE #317: PASSED tests are marked as 'Not investigated'

k0machi:
This change adjusts the behaviour of stats collector to always show last
status to be the worst case for the grouped runs - runs such as rolling
upgrades will now display comments and investigation status for the
worst case out of all of them.

Fixes #317

### DIFF
--- a/argus/backend/service/stats.py
+++ b/argus/backend/service/stats.py
@@ -126,7 +126,7 @@ def generate_field_status_map(
         field_name = "status",
         container_class = TestStatus,
         cmp_class = ComparableTestStatus
-    ) -> dict[int, str]:
+    ) -> dict[int, tuple[str, TestRunStatRow]]:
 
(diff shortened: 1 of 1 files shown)


k0machi:
image

roydahan (approved):

fruch (approved):
LGTM

roydahan:
@k0machi ?

PR #340, opened 2024-02-20
improvement(TestRun.svelte): Rebuild link for Jenkins

k0machi:
This change adds a link to Jenkins rebuild page for each of the
supported TestRun cards. This is initial part of the logic to eventually
have automatic rebuilds/clones of Jenkins jobs inside argus.

Task: scylladb/qa-tasks#1612

### DIFF
--- a/frontend/TestRun/DriverMatrixRunInfo.svelte
+++ b/frontend/TestRun/DriverMatrixRunInfo.svelte
@@ -1,6 +1,6 @@
 <script>
     import {
-        faBusinessTime,
+        faBusinessTime, faPlay,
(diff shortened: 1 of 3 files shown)


roydahan:
I think it would be better to put it next to all other actions we have in the bottom, it should be much clearer.

Screenshot 2024-02-21 at 20 18 16

k0machi:
@roydahan Updated:

chrome_M3JHjqK6NP

fruch (approved):
LGTM

roydahan:
@k0machi ping

PR #344, opened 2024-03-13
feature(argus): Job rebuild via Jenkins API 

k0machi:
This PR replaces Jenkins rebuild URL with a custom wizard that
allows user to execute a job rebuild from inside Argus. The process is
split into two distinct phases - fetching and editing job parameters and
starting and verifying the job. On the first phase the user is presented
with a simple editor (currently only supporting string values) which
shows all parameters used for the job (with a toggle to show all
parameters for the entire workflow, i.e. empty ones), allowing user to
make necessary changes before starting rebuild. On the second phase the
job is started and verified until jenkins provides a build number and a
job url - at this point the wizard is finished and user is notified of
success.

Closes https://github.com/scylladb/qa-tasks/issues/1612

### DIFF
--- a/frontend/TestRun/Jenkins/SCTParameterWizard.svelte
+++ b/frontend/TestRun/Jenkins/SCTParameterWizard.svelte
@@ -1,0 +1,549 @@
+<script>
+    import Fa from "svelte-fa";
+    import { sanitizeSelector } from "../../Common/TextUtils";
+    import { Collapse } from "bootstrap";
+    import DummySelectParam from "./DummySelectParam.svelte";
+    import SelectParam from "./SelectParam.svelte";
+    import StringParam from "./StringParam.svelte";
+    import { faExclamationCircle, faMinus, faPlus, faQuestionCircle } from "@fortawesome/free-solid-svg-icons";
+    import DummyCheckParam from "./DummyCheckParam.svelte";
+
+    /**@type {{[string]: string}}*/
+    export let params = {};
+    export let rawParams = [];
+
+    const determineVersionSource = function(params) {
+        const keysToCheck = {
+            scylla_version: "scylla_version",
+            scylla_repo: "scylla_repo",
+            new_scylla_repo: "rolling_upgrade",
+            scylla_ami_id: "scylla_image",
+            gce_image_db: "scylla_image",
+            azure_image_db: "scylla_image",
+        };
+
+        for (const [key, type] of Object.entries(keysToCheck)) {
+            if (params[key]) {
+                return type;
+            }
+        }
+        return keysToCheck.scylla_version;
+    };
+
+    const conditionWrapper = function(definition, params, allDefs) {
+        if (!(definition.internalName in params) && !definition.formOnly) return false;
+        const result = definition.condition(params, allDefs);
+        if (result && definition.onShow) {
+            definition.onShow(params);
+        } else if (!result && definition.onHide) {
+            definition.onHide(params);
+        }
+        return result;
+    };
+
+    const onChangeWrapper = function(event, definition, params) {
+        if (definition.onChange) definition.onChange(event, params);
+        if (definition.requiresValidation) {
+            definition.validated = definition.validate(params);
+        }
+        paramDefinitions = paramDefinitions;
+    };
+
+    const ALL_BACKENDS = {
+        azure: "Microsoft Azure",
+        aws: "Amazon AWS", 
+        "aws-siren": "Amazon AWS (Siren-tests)", 
+        "k8s-local-kind-aws": "Kubernetes in Docker on AWS", 
+        "k8s-eks": "Kubernetes (Amazon EKS)",
+        gce: "Google Cloud",
+        "gce-siren": "Google Cloud (Siren-tests)",
+        "k8s-local-kind-gce": "Kubernetes in Docker on GCE", 
+        "k8s-gke": "Kubernetes (Google Kubernetes Engine)",
+    };
+
+    /**
+     * 
+     * @param {string} backend
+     * @param {string} group
+     */
+    const inBackendGroup = function (backend, group) {
+        const BACKEND_GROUPS = {
+            aws: ["aws", "aws-siren", "k8s-local-kind-aws", "k8s-eks"],
+            azure: ["azure"],
+            gce: ["gce", "gce-siren", "k8s-local-kind-gce", "k8s-gke"],
+            k8s: ["k8s-local-kind-aws", "k8s-eks", "k8s-gke"],
+        };
+        return (BACKEND_GROUPS[group] ?? []).includes(backend);
+    };
+
+    let paramDefinitions = {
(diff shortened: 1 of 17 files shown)


k0machi:
Demo:


https://github.com/scylladb/argus/assets/7761415/5dce6310-d1c3-465a-9166-38851832f9ec

k0machi:
Demo for the Wizard component:

https://github.com/scylladb/argus/assets/7761415/6eabea0c-1e43-4386-b941-2091337b465e

https://github.com/scylladb/argus/assets/7761415/0f9c9b96-9d21-4f01-9497-8e1548a992d4

fruch:
I find it a bit confusion, that we have display name for the field values i.e. Destroy, or those long titles for backends.
it might confuse people later looking it up in jenkins or in logs  

I would rather see something like
- aws ( AWS )
- azure (Microsoft Azure)

and so on

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
```diff
@@ -0,0 +1,414 @@
...
+                dbNodes: {
+                    name: "Action for DB Cluster",
+                    type: SelectParam,
+                    internalName: "post_behavior_db_nodes",
+                    condition: (params, defs) => true,
+                    onChange: function (e, params) {
+                        //empty
+                    },
+                    values: ["keep", "keep-on-failure", "destroy"],
+                    labels: ["Preserve", "Preserve on Failure", "Destroy"],
```
fruch, line 327: please don't rename the values like that, it just gonna cause confusion

maybe we want to add some help text for each, that would explain a bit better the behavior, but like this I will tell someone to run with keep, and he won't find it..

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
```diff
@@ -0,0 +1,414 @@
...
+                    internalName: "email_recipients",
+                    condition: (params, defs) => true,
+                    onChange: function (e, params) {
+                        //empty
+                    }
+                },
+                stressDuration: {
+                    name: "Custom Stress Command Duration",
+                    type: StringParam,
+                    internalName: "stress_duration",
```
fruch, line 306: we need to have the internalName visible on the form, it would be confusing with it

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
```diff
@@ -0,0 +1,414 @@
...
+                    type: StringParam,
+                    internalName: "azure_image_db",
+                    condition: (params, defs) => (defs.scyllaVersion.currentSource == "scylla_image" && inBackendGroup(params.backend, "azure")),
+                    onChange: function (e, params) {
+                        params.scylla_version = "";
+                        params.scylla_repo = "";
+                    }
+                },
+                updateDbPackages: {
+                    name: "Package Update URI",
```
fruch, line 244: this a bit opaque, and missing the reference to DB 
again this need the explanation, otherwise no one would understand what it is

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
```diff
@@ -0,0 +1,414 @@
...
+                    onChange: function (e, params) {
+                        params.scylla_repo = "";
+                        params.scylla_ami_id = "";
+                        params.gce_image_db = "";
+                        params.azure_image_db = "";
+                    }
+                },
+                scyllaRepo: {
+                    name: "Scylla Repository",
+                    description: "HTTP(s) link to the repo file",
```
fruch, line 198: repo or list

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
```diff
@@ -0,0 +1,414 @@
...
+                provisionType: {
+                    name: "Provision Type",
+                    type: SelectParam,
+                    internalName: "provision_type",
+                    condition: (params, defs) => true,
+                    onChange: function (e, params) {
+                        //empty
+                    },
+                    values: ["spot", "on_demand", "spot_fleet"],
+                    labels: ["Spot Provisioning", "On Demand", "Spot Fleet"]
```
fruch, line 159: I don't think we need to prettify the name in this case

fruch (changes requested):
Overall it looks good.

Just I think we shouldn't try to prettify the values - cause it would cause confusion, we can add help text to values, but we need to see the actual names.

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
k0machi: Done

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
k0machi: Done

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
k0machi: removed `Provisioning` from the first value

k0machi:
> Overall it looks good.
> 
> Just I think we shouldn't try to prettify the values - cause it would cause confusion, we can add help text to values, but we need to see the actual names.

We can discuss this on today's sync meeting as to how we should deal with pretty values - I think we should have them as the intention of the wizard is to allow people who don't work with SCT to easily queue up a job.

After your requested changes this is how form looks now:
image

image

fruch:
> "allow people who don't work with SCT to easily queue up a job"

that doesn't say anything about capital letter, or relabeling things differently the original values

we did talked about help texts that we might want to add to explain things, but I still think we shouldn't be hiding the original names of things, if they are bad names, we should change the in SCT.

fruch:
> > Overall it looks good.
> > Just I think we shouldn't try to prettify the values - cause it would cause confusion, we can add help text to values, but we need to see the actual names.
> 
> We can discuss this on today's sync meeting as to how we should deal with pretty values - I think we should have them as the intention of the wizard is to allow people who don't work with SCT to easily queue up a job.
> 
> After your requested changes this is how form looks now: image
> 
> image

backend list is a bit better

I think we can put them side by side, 

original name - extra descriptive name 

and some cases it's a bit silly, like: 

Scylla Version 
vs.
scylla_version

I don't think an extra help is needed in this case.writing both doesn't add anything

we should have both text, when it's adding some more description, otherwise the original value is enough.

k0machi:
Post discussion this is what we have now:

* AZ changed to string field
* `update_db_packages` description name changed, pending a checkbox to hide it if it's empty
* Email broken out of test config into a separate category at the bottom
* Select component now prevents changes if it encounters an unknown value - solution for multidc tests

image

Still validation left to do - in particular:
 * Validate that at least ONE installation related parameter is filled out
 * Validate its content - they're all free-form strings currently
 * Validate that email is filled out - if not, replace with qa email

Issue #341, opened 2024-03-03
Display issues with the new "Test-Error" status


roydahan:
When choosing the "TEST-ERROR" Status, there are several issues with it:
1. The status "severity" - currently when choosing TEST-ERROR its "Severity" order is "less" than PASSED and it means that it moves to be the least significant status so when you have more than one run with the same build number, TEST-ERROR moves to the most right and the others with "PASSED" status "wins" the general status (see screenshot below).
2. Duplicates the icons of the tabs when switching from FAILED TO TEST-ERROR and vice-versa (see screeshot below).
3. COLOR should be ORANGE or something different than FAILED.

![Screenshot 2024-03-03 at 14 26 45](https://github.com/scylladb/argus/assets/20959584/1e70b5d9-5d12-4f59-b8b7-eeca708c25b2)
![Screenshot 2024-03-03 at 14 26 56](https://github.com/scylladb/argus/assets/20959584/b16f8f4d-d721-4eae-9324-2f98249051ea)

roydahan:
The order in the status bar should also be changed to be right after the regular failure. (Green, Red, Orange, Black,...)

![Screenshot 2024-03-17 at 19 49 50](https://github.com/scylladb/argus/assets/20959584/ec4a570f-f2d9-46e1-af91-d5bdf5c883c8)

fruch:
@k0machi you have PRs for those fixes ?

PR #347, opened 2024-03-25
fix(TestRun.svelte): Use svelte-fa instead of <i> tag

k0machi:
This fixes an issue where repeated mount/dismount of component would
duplicate Font-Awesome icons in the markup, as they don't have same
cleanup procedures as svelte does.

Task: #341


fruch (approved):
LGTM

PR #348, opened 2024-03-25
fix(TestRuns.svelte): Re-enable background refresh for Workspace Runs

### LINKED ISSUE #343: Re-introduce "Auto Refresh" in Workspace area
When working in the workspace area on multiple jobs in parallel and rebuilding some jobs the auto-refresh feature is very helpful.
**Refreshing manually is not convenient** and it causing the user to miss jobs that may fail during rebuild.

k0machi:
This commit reverts the behaviour of TestRuns component where it will
now again refresh without tab being in focus

Fixes #343


fruch (approved):
LGTM

PR #349, opened 2024-04-11
feature(jenkins_service): Job cloning for Jenkins

k0machi:
This commits adds logic and frontend components to clone a job inside
jenkins. The main difference between rebuild and clone being
the target release and group selector, in addition to "Advanced Settings"
menu, which differs between job types. All 3 main jobs types are
supported.

Task: scylladb/qa-tasks#1653

### DIFF
--- a/argus/backend/service/jenkins_service.py
+++ b/argus/backend/service/jenkins_service.py
@@ -1,12 +1,17 @@
+import re
+import requests
 from typing import Any, TypedDict
+from uuid import UUID
 import xml.etree.ElementTree as ET
 import jenkins
 import logging
 
 from flask import current_app, g
 
-LOGGER = logging.getLogger(__name__)
+from argus.backend.models.web import ArgusGroup, ArgusRelease, ArgusTest, UserOauthToken
 
+LOGGER = logging.getLogger(__name__)
+GITHUB_REPO_RE = r"(?P<http>^https?:\/\/(www\.)?github\.com\/(?P<user>[\w\d\-]+)\/(?P<repo>[\w\d\-]+)(\.git)?$)|(?P<ssh>git@github\.com:(?P<ssh_user>[\w\d\-]+)\/(?P<ssh_repo>[\w\d\-]+)(\.git)?)"
 
 class Parameter(TypedDict):
     _class: str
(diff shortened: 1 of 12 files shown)


k0machi:
Demo: 

https://github.com/scylladb/argus/assets/7761415/8a1f59b4-3b14-4aba-a47c-6f1711936afb

roydahan:
Looks great.
A question about target release, when one chooses a "target release", where will it put the job?

IMO,In scylla-staging it should put it under "user name" but in any other release it should put it under "Reproducers".
The form should explain that somehow...

k0machi:
> Looks great. A question about target release, when one chooses a "target release", where will it put the job?
> 
> IMO,In scylla-staging it should put it under "user name" but in any other release it should put it under "Reproducers". The form should explain that somehow...

For now, it allows user to select a subfolder, currently filtered down to which ones have a valid jenkins url in them, I plan to add the `reproducer` filter, so that for releases it shows that (and only that folder) and a special 'staging` flag for release which will filter based on user's name/email/fullname and look for that.

roydahan:
Ok, do you plan to do it in this PR or a follow up PR?

k0machi:
> Ok, do you plan to do it in this PR or a follow up PR?

Reproducer filter added, I've decided against a flag for now, substring of a name containing staging will do for now.

k0machi:
@fruch @roydahan PR ready, waiting for review

fruch:
> @fruch @roydahan PR ready, waiting for review

what's in the demo from 2 weeks ago, is still relevant ? i.e. we'll need to see it before we can approve

roydahan:
I saw it during the weekly Argus call and it behaves well.
Same as above with some minor improvements of filtering destination and handling the first and second building.

roydahan (approved):

k0machi:
https://github.com/scylladb/argus/assets/7761415/9a8013a9-542b-421e-840f-bad86b61d8fb

Updated demo^

@fruch

fruch:
Few comment

* I'm not sure why we can't queue job 2, right away, and now waiting for 1 to finish 

* We need to not disable Jenkins output, about the quiet period, it is just confusing, the user has nothing to do with it

argus/backend/service/jenkins_service.py
```diff
@@ -15,9 +20,31 @@ class Parameter(TypedDict):
...
 
 
+class JenkinsServiceError(Exception):
+    pass
+
+
 class JenkinsService:
     RESERVED_PARAMETER_NAME = "requested_by_user"
 
+    SETTINGS_CONFIG_MAP = {
```
fruch, line 30: What's the point of this mapping ? If they are identical?

fruch (approved):
LGTM

argus/backend/service/jenkins_service.py
k0machi: I was thinking if they were to diverge in the future - since iirc a job xml allows for more than one configured repo for example, so while this isn't really necessary right now (as evidenced that handling function used is the same, it's here in case we'll have to adapt the job advanced settings page further.

k0machi:
> Few comment
> 
> * I'm not sure why we can't queue job 2, right away, and now waiting for 1 to finish
> * We need to not disable Jenkins output, about the quiet period, it is just confusing, the user has nothing to do with it

First one - trying to do that I get immediate error returned by the API, so I assume some kind of jenkins limitation? I could queue  "transparently" by firing off a timeout either client-side or server-side.

Second one - the message seen is what jenkins responds with when queried about the job queued - eventually it would change into actual job information (once started) or into other system messaged, e.g. "waiting in queue".

k0machi:
Going to change jenkins poll rate to 5 seconds to make user's life easier - build number 1 is usually complete in 10 seconds in case of sct

PR #351, opened 2024-04-25
fix(driver_matrix_tests): Make some fields optional in JUnit testcases

k0machi:
This commit adjusts which fields are required from a testcase, since
certain frameworks do not report classname in certain cases (e.g. rust
driver matrix)


juliayakovlev (approved):

juliayakovlev:
Example: skipped test case of rust driver does not report classname

roydahan (approved):

PR #355, opened 2024-05-14
improvement(plugins/sct): Scan for CoreDumpEvent on events submission

### LINKED ISSUE #326: Bug Template - Add coredump download links next to all other logs
As suggested by @avikivity here:
https://github.com/scylladb/scylla-enterprise/issues/3726#issuecomment-1871159838

It's basically going over the events, and taking the download instructions out

k0machi:
This change makes argus look for CoreDumpEvent events in submitted
events on test completion - parsing ones that contain links to coredumps
and storing them in the 'logs' container, which can be viewed on the
"Logs" tab in the frontend. Note that some cores might be lost if SCT
generates a lot of error events, shadowing last 100 it submits to argus.

Fixes #326

### DIFF
--- a/argus/backend/plugins/sct/service.py
+++ b/argus/backend/plugins/sct/service.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass
 from functools import reduce
 import logging
 import math
+import re
 from time import time
 from flask import g
 from argus.backend.models.web import ArgusEventTypes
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

Even though it might be confusing that the name isn't the name of the actual file

k0machi:
> LGTM
> 
> Even though it might be confusing that the name isn't the name of the actual file

The actual name is not very readable I feel, you'd have to cross reference which node it belongs to, while like this you see the node and the filename is in the url provided.

fruch:
@k0machi do we have a link to a job that demonstrate this ?

PR #357, opened 2024-05-14
fix(argus): update scylla-driver to 3.26.8

k0machi:
Updating to solve some performance issues with CQLEngine and such.


fruch (approved):
LGTM

PR #359, opened 2024-05-14
feature(TestRuns): Jenkins Job Configuration Modal

k0machi:
This commit adds Jenkins settings page to the TestRuns banner (while
also slightly redesigning the look and feel of said banner component).
The modal allows adjusting job settings same way they can be adjusted
during clone, but available without cloning the original job.

### DIFF
--- a/argus/backend/controller/testrun_api.py
+++ b/argus/backend/controller/testrun_api.py
@@ -6,6 +6,7 @@ from flask import (
 )
 
 from argus.backend.error_handlers import handle_api_exception
+from argus.backend.models.web import ArgusTest
 from argus.backend.service.jenkins_service import JenkinsService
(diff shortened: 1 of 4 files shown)


k0machi:
Demo:


https://github.com/scylladb/argus/assets/7761415/8d1fb0ac-915e-4bdb-8527-932650670800

roydahan:
LGTM, BUT!
Do we want to allow that in all jobs? even official ones?
IIUC it will do all changes using the Argus user, so if someone changes an official release job we won't know who did it and why.

Bottom line, It's a risk to add that in all jobs, but I think it's ok for staging.

k0machi:
Done, restricted button to staging and admins only for everything else

PR #362, opened 2024-05-20
fix(stats.py): Refactor row-fetching for stats.

k0machi:
This commit address performance bottlenecks in the stats fetching
module. Specifically it:
 - No longer uses release_id index, instead opting out to an IN query
   that includes relevant tests for the release
 - The 15 test limit is now achieved via PER PARTITION LIMIT clause in
   the stat query, instead of discarding indexed data
 - To avoid exceeding maximum cartesian product of an IN query a
   batching mechanism is introduced for reads, splitting stats query
   into N amount of 100 test queries
 - The queries themselves are using execute_async to execute in parallel
 - Most linear searches are now replaced with map lookups (trading
   multiple searches with an upfront map setup cost)

### DIFF
--- a/argus/backend/plugins/core.py
+++ b/argus/backend/plugins/core.py
@@ -1,5 +1,6 @@
 import logging
 from datetime import datetime
+from math import ceil
 from uuid import UUID
 from time import time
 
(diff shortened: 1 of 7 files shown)


fruch:
@k0machi this look o.k., but I would still suggest talking with @nyh seeing if there a better way

fruch (approved):
LGTM

roydahan (approved):

PR #366, opened 2024-05-23
improvement(plugins/sct): Use different naming scheme for coredump links

k0machi:
This change changes the current "COREDUMP-{whatever it is in the node
name in sct" into "core.scylla-{node_instance_name}.gz"


fruch:
I think timestamp would be helpful as part of the name, since there could be multiple cores on the same node.

k0machi:
@fruch Done

PR #369, opened 2024-05-28
improvement(ReleaseDashboard): UX adjustments for Quick Selector

k0machi:
This commit fixes and improves some of the usability issues related to
TestPopoutSelector.

* The button group now starts on top and follows cursor
* Double scroll has been removed
* A button to scroll to the top of the window has been added
* A "Remove All" button is now present alongside "Investigate Selected" button
* Quick filters are now explained better and contain additional buttons,
  such as: Select all failed and not investigated, Select all
  investigated without additional info.
* Version container has been adjusted slightly to contain its elements
  better and also highlighted from the background
* The quick select buttons inside the investigation breakdown are now
  indicated that are buttons better, using a chevron icon on each of
  them.
* The build number on quick investigate selector is now clickable and
  leads to the run page.
* Each row in quick investigate now shows if there's a comment or issue
  on the run.

### DIFF
--- a/frontend/Stats/NumberStats.svelte
+++ b/frontend/Stats/NumberStats.svelte
@@ -4,6 +4,7 @@
     import { titleCase, subUnderscores } from "../Common/TextUtils";
     import { Collapse } from "bootstrap";
     import { createEventDispatcher } from "svelte";
+    import { faChevronRight } from "@fortawesome/free-solid-svg-icons";
     export let displayNumber = false;
     export let displayPercentage = false;
     export let displayInvestigations = false;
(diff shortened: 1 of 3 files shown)


k0machi:
Investigation widget:
image

Version selector and quick filter:
image

Sidebar Test Viewer:
image

Only buttons now follow the viewport:
image

frontend/Stats/NumberStats.svelte
```diff
@@ -4,6 +4,7 @@
     import { titleCase, subUnderscores } from "../Common/TextUtils";
     import { Collapse } from "bootstrap";
     import { createEventDispatcher } from "svelte";
+    import { faChevronRight, faDotCircle } from "@fortawesome/free-solid-svg-icons";
```
fruch: maybe something like
`<i class="fa-regular fa-object-group"></i>`

would be better ?

the cheveron isn't exactly that clear in it's meaning

fruch (approved):
All of the changes looks good to me,

beside the icon inside the investigation widget, it doesn't express any meaning in regards of the selection operation its going to be doing.

roydahan:
The Chevron has some relation to expend so it’s kind of ok, not very intuitive but ok.

PR #373, opened 2024-05-29
feature(argus): Update argus dependencies (Flask 3.0.x)

k0machi:
This commit relaxes some of the poetry dependencies and additionally
updates Flask and other components to their most recent counterparts.
Due to certain deprecations, a new ArgusJSONProvider class is now
replacing previous serializer for Argus objects.


fruch (approved):
LGTM

fruch:
@k0machi 

How do we test such a change ?

I.e. we need a staging env for this one

And initial CI running unittest to the very minimum (pick up GitHub actions and a run of the unittests)

k0machi:
> @k0machi
> 
> How do we test such a change ?
> 
> I.e. we need a staging env for this one
> 
> And initial CI running unittest to the very minimum (pick up GitHub actions and a run of the unittests)

Yes, this one needs it, I'll set up some basic action to test it, just to verify the startup and then I'll proceed with the staging env.

fruch:
Also what is the reason for this update ? to support python 3.12 ?

I.e. what's the urgency of this ? is it blocking any task you are working on ?

fruch:
@k0machi 

for this change production would need to change to python3.12 ?

if yes, let's move it to use pyenv, so'll have better control on python versions we need.

k0machi:
> @k0machi 
> 
> for this change production would need to change to python3.12 ?
> 
> if yes, let's move it to use pyenv, so'll have better control on python versions we need. 

Production is already using pyenv so no changes needed there aside from installing 3.12 and changing startup script to run that

argus/backend/util/encoders.py
```diff
@@ -3,12 +3,14 @@
...
 from uuid import UUID
 
+from flask.json.provider import DefaultJSONProvider
 import cassandra.cqlengine.usertype as ut
 import cassandra.cqlengine.models as m
 
 
 LOGGER = logging.getLogger(__name__)
 
+# TODO: Deprecated.
```
fruch, line 13: can't we remove it ?

k0machi: Since it's an encoder class for `json.dump(s)` and not a provider class for `flask.Flask`, I'll keep it and remove the TODO marker as it is somewhat still used by some scripts and might be useful (would probably unify the default method however)

Issue #352, opened 2024-04-25
RFE - "User View" feature - ability to define a "view" that present a user only certain things


roydahan:
As there are different users to Argus and each cares about different things, I'm suggesting a new method of looking at things with regards to releases.
(In general it can be relevant to cross releases in cases like performance, but for now we can focus on views inside of releases).

The way I imagine that is an ability to pick a widget (e.g. "All issues", "Status Bar", "Jobs Last status" and more when we invent them), for each widget we will be able to define the way to filter, group and sort (E.g. filter only jobs I care when I work on a patch release, or group the jobs by type, assignee, etc).

Main use cases I see:
1. "ALL" (our current view under a release) where we have all widgets and all jobs, grouped by their Jenkins folder.
2. "Patch Release" - a view that filters only the jobs that should be auto triggered for a patch release.
3. "Release Engineering" - a view that filters only the jobs releng team cares about (e.g. their jobs + dtest + auto triggered jobs).
4. "Standard User" - only his staging jobs / reproducers jobs.

roydahan:
@k0machi let's have it discussed in our weekly meeting next week.

k0machi:
Work in progress PR #374

PR #374, opened 2024-06-05
feature(argus): Custom Views

k0machi:
This commit adds a new way to present tests inside argus - Views. Built
by users, they allow scoping tests, issues and statistics into a more
focused view. The View Manager allows users to build view from existing
tests, groups and releases, along with a new widget builder allowing
users to select what their view should contain widget wise (Currently
contains 3 widgets, copying the release dashboard structure).
Additionally, the Issue View was improved to allow filtering by issue
state and labels, as well as caching issue state to prevent extraneous
fetches.

Fixes https://github.com/scylladb/argus/issues/352

### DIFF
--- a/argus/backend/service/views.py
+++ b/argus/backend/service/views.py
@@ -1,0 +1,243 @@
+import datetime
+import logging
+import re
+from functools import partial, reduce
+from typing import Any, Callable, TypedDict
+from urllib.parse import unquote
+from uuid import UUID
+
+from cassandra.cqlengine.models import Model
+from argus.backend.models.web import ArgusGroup, ArgusRelease, ArgusTest, ArgusUserView, User
+from argus.backend.plugins.loader import all_plugin_models
+from argus.backend.util.common import chunk, current_user
+
+LOGGER = logging.getLogger(__name__)
+
+
+class UserViewException(Exception):
+    pass
+
+
+class ViewUpdateRequest(TypedDict):
+    name: str
+    description: str
+    display_name: str
+    tests: list[str]
+    widget_settings: str
+
+
+class UserViewService:
+    ADD_ALL_ID = UUID("db6f33b2-660b-4639-ba7f-79725ef96616")
+    def create_view(self, name: str, items: list[str], widget_settings: str, description: str = None, display_name: str = None) -> ArgusUserView:
+        try:
+            name_check = ArgusUserView.get(name=name)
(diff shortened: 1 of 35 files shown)


k0machi:
Picture demos:

* View Editor (currently only available to admins)
image

* View overview page
image

* View page itself
image

* Improved issue view
image

fruch (approved):
LGTM

soyacz:
what's the difference between 'view name' and 'display name for users'? Can we merge them?

roydahan (approved):

k0machi:
@soyacz good idea and yes, I forgot to implement that:

image

argus/backend/service/views.py
```diff
@@ -0,0 +1,232 @@
...
+    display_name: str
+    tests: list[str]
+    widget_settings: str
+
+
+class UserViewService:
+    ADD_ALL_ID = UUID("db6f33b2-660b-4639-ba7f-79725ef96616")
+    def create_view(self, name: str, items: list[str], widget_settings: str, description: str = None, display_name: str = None) -> ArgusUserView:
+        view = ArgusUserView()
+        view.name = name
```
soyacz, line 33: what happens if someone posts the same name? Won't it override?

k0machi: Hmm, currently the sluggified name is not really used - it won't override as the partition key is the `id` column. To be honest I'm not sure we really need the view name, maybe I could output as a helper for the view path e.g. `/views/<uuid>/<slug>` just to make it easier for user

soyacz: uuid's in path are not helping and are generally looking bad.
can't we enforce slug (name) to be unique and make /view/<slug> path?

k0machi: We can, will do that.

soyacz: gh, cut `<slug>` from my comment above (fixed), I hope you got it

soyacz: Looks great now.

😅 polish symbols match, how about Cyrillic?

soyacz (approved):
LGTM

argus/backend/service/views.py
k0machi: Non latin scripts just get stripped unfortunately, I think the library supports custom matching dictionaries but it's its default behaviour.

image

soyacz: no problem, anyway we don't expect using different language than english there

PR #378, opened 2024-06-25
feature(Notifications/Comment): Add inline reply to notifications

### LINKED ISSUE #203: Reply button on notifications screen doesn't work
When opening the notifications on Android chrome, replay button doesn't do anything.

k0machi:
This commit adds functionality required to do replies inside
notification viewer, skipping the need for going into the main source
body and sending a comment there.
        
Fixes #203

### DIFF
--- a/argus/backend/models/web.py
+++ b/argus/backend/models/web.py
@@ -180,6 +180,7 @@ class ArgusTestRunComment(Model):
     test_run_id = columns.UUID(required=True, index=True)
     user_id = columns.UUID(required=True, index=True)
     release_id = columns.UUID(required=True, index=True)
+    test_id = columns.UUID(required=True, index=True)
     posted_at = columns.Integer(
         required=True, clustering_order="desc", primary_key=True)
     message = columns.Text(min_length=1, max_length=65535)
--- a/frontend/Profile/NotificationCommentWrapper.svelte
+++ b/frontend/Profile/NotificationCommentWrapper.svelte
@@ -1,8 +1,134 @@
 <script>
     import Comment from "../Discussion/Comment.svelte";
+    import CommentEditor from "../Discussion/CommentEditor.svelte";
+    import { sendMessage } from "../Stores/AlertStore";
     import { userList } from "../Stores/UserlistSubscriber";
     let users = $userList;
     export let supportData;
+    let replying = false;
+    let fetching = false;
+    let replyUser = "";
+
+    const newCommentTemplate = {
+        id: "",
+        message: "",
+        release: "",
+        reactions: {},
+        mentions: [],
+        user_id: "",
+        release_id: "",
+        test_run_id: supportData.test_run_id,
+        posted_at: new Date(),
+    };
+
+    let newCommentBody;
+
+    const handleReplyButton = function (e) {
+        newCommentBody = Object.assign({}, newCommentTemplate);
+        let authorId = e.detail.author;
+        let authorUsername = users?.[authorId]?.username;
+        replyUser = authorUsername;
+        let quotedMessage = e.detail.message
+            .trim()
+            .replaceAll("&gt;", ">")
+            .replaceAll("&lt;", "<")
+            .split("\n")
+            .map((val) => `> ${val}`)
+            .join("\n");
+        newCommentBody.message = `${authorUsername ? `@${authorUsername} posted:` : ""}${
+            newCommentBody.message
+        }\n${quotedMessage}\n`;
+        replying = true;
+    };
+
+    const handleCommentSubmit = async function (e) {
+        let commentBody = e.detail;
+        fetching = true;
+        try {
+            let apiResponse = await fetch(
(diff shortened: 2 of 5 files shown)


argus/backend/models/web.py
```diff
@@ -180,6 +180,7 @@ class ArgusTestRunComment(Model):
     test_run_id = columns.UUID(required=True, index=True)
     user_id = columns.UUID(required=True, index=True)
     release_id = columns.UUID(required=True, index=True)
+    test_id = columns.UUID(required=True, index=True)
```
soyacz, line 183: @k0machi I'll be asking basic questions to understand more your PR's. Please be patient :)

So, the first one, is what's a difference between `test_run_id` and `test_id`?
Also, why comment.id is needed?  Couldn't we use test_run_id as partition key and posted_at clustering key

frontend/Profile/NotificationCommentWrapper.svelte
```diff
@@ -1,8 +1,132 @@
...
+        }\n${quotedMessage}\n`;
+        replying = true;
+    };
+
+    const handleCommentSubmit = async function (e) {
+        let commentBody = e.detail;
+        fetching = true;
+        try {
+            let apiResponse = await fetch(
+                `/api/v1/test/${supportData.test_id}/run/${supportData.test_run_id}/comments/submit`,
```
soyacz, line 47: why test_id is needed?

argus/backend/models/web.py
k0machi: The table for  comments is... weird, I'm not sure why I made it like this back then and it'll need a rework at some point - you are correct  that `id` is completely extraneous and `test_run_id` should be the partition key. `test_id` is the id of the test entity that hosts runs, i.e. `longevity-10gb-3h` for `longevity-10gb-3h#3`, it is needed to later be able to make the API call to post comment.

frontend/Profile/NotificationCommentWrapper.svelte
k0machi: The `test_id` is needed to resolve the plugin to which the `test_run_id` belongs to, without it we would have to query each plugin table - SCT, driver matrix, sirenada - to be able to find the run.

argus/backend/models/web.py
soyacz: So, `test_id` is like a bucket containing runs of a test. And `test_run_id` matches actual `test_id` in SCT?

As to refactor, we'll see if needed later.

frontend/Profile/NotificationCommentWrapper.svelte
soyacz: why we care about plugin in context of comments?

soyacz (commented):
general questions, not related to PR itself, just clarifying my understanding.
Can you post screenshots how it looks when replying?

frontend/Profile/NotificationCommentWrapper.svelte
k0machi: To build the url for the user to click inside notification in case they want to visit the run

argus/backend/models/web.py
k0machi: Yes, that is correct. `test_id` in sct is what in argus mostly called `run_id` with some exceptions like this one (because this is a very old table)

k0machi:
@soyacz 

https://github.com/scylladb/argus/assets/7761415/2341f391-8ee9-4e77-a7e4-f4216809a664

soyacz:
There's a problem with rendering previous `>` characters - converted to `&gt`. Can we do something about it?
Maybe we could switch to markdown when using comments to improve the look?

soyacz (approved):
One remark, but generally it fixes the original issue. LGTM

k0machi:
> There's a problem with rendering previous `>` characters - converted to `&gt`. Can we do something about it? Maybe we could switch to markdown when using comments to improve the look?

it's to do with who does the sanitization - currently the comment is sanitized before saving by the backend, we could instead sanitize it on the frontend when it is retrieved (since frontend's markdown renderer already does sanitization, this should be simple)

soyacz:
> > There's a problem with rendering previous `>` characters - converted to `&gt`. Can we do something about it? Maybe we could switch to markdown when using comments to improve the look?
> 
> it's to do with who does the sanitization - currently the comment is sanitized before saving by the backend, we could instead sanitize it on the frontend when it is retrieved (since frontend's markdown renderer already does sanitization, this should be simple)

please add a task for it, so this issue does not vanish (or fix here if it is so simple).

k0machi:
> > > There's a problem with rendering previous `>` characters - converted to `&gt`. Can we do something about it? Maybe we could switch to markdown when using comments to improve the look?
> > 
> > 
> > it's to do with who does the sanitization - currently the comment is sanitized before saving by the backend, we could instead sanitize it on the frontend when it is retrieved (since frontend's markdown renderer already does sanitization, this should be simple)
> 
> please add a task for it, so this issue does not vanish (or fix here if it is so simple).

Done in here, now the reply window will have `<>`

fruch (approved):
LGTM

PR #386, opened 2024-07-08
improvement(GithubIssues.svelte): Modal window for viewing attached runs

### LINKED ISSUE #157: Issue that was mentioned in many runs, spans in ugly way in the list of issues
Screen Shot 2022-07-27 at 14 28 40

I think we should show by default only last 10 runs, with an option to expand

k0machi:
This commit adds new button to the aggregated issue (an issue that
contains multiple runs, e.g. on release dashboard) that allows viewing
all the runs attached to it as a list. It also adds an endpoint to
resolve those runs into human readable strings to provide an overview.
Additionally, a new "Investigate selected" button is added on each modal
window to jump to workspace with **test** in question (each run is
clickable to get to the test+run page. Additionally, "Share" button on
workspace page is now fixed, Labels on issues are now individually
clickable (replicating "All labels" functionality) and the labels strip
is now hidden by default and is revealed by a button.

Fixes #157

### DIFF
--- a/argus/backend/controller/testrun_api.py
+++ b/argus/backend/controller/testrun_api.py
@@ -289,6 +289,22 @@ def ignore_jobs():
     }
 
 
+@bp.route("/get_runs_by_test_id_run_id", methods=["POST"])
+@api_login_required
+def get_runs_by_test_id_run_id():
+    payload: list[tuple[UUID, UUID]] = get_payload(request)
+    service = TestRunService()
+
+    result = service.resolve_run_build_id_and_number_multiple(payload)
+
+    return {
+        "status": "ok",
+        "response": {
+            "runs": result
+        }
+    }
+
+
 @bp.route("/jenkins/params", methods=["POST"])
 @api_login_required
 def get_jenkins_job_params():
--- a/argus/backend/service/testrun.py
+++ b/argus/backend/service/testrun.py
@@ -1,4 +1,6 @@
+from collections import defaultdict
 from datetime import datetime, timedelta
+from functools import reduce
 import json
 import logging
 import re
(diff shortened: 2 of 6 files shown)


k0machi:
Demo:

image
chrome_MkKF1xP4cF
image

fruch (approved):
LGTM

fruch:
> Demo:
> 
> image chrome_MkKF1xP4cF image

looks much more useful then before

argus/backend/controller/testrun_api.py
```diff
@@ -289,6 +289,22 @@ def ignore_jobs():
     }
 
 
+@bp.route("/resolve_runs", methods=["POST"])
+@api_login_required
+def resolve_runs():
```
soyacz, line 294: why not `get_runs_build_id_and_number`?
For me would be much easier to understand

argus/backend/controller/testrun_api.py
```diff
@@ -289,6 +289,22 @@ def ignore_jobs():
...
+    payload = get_payload(request)
```
soyacz, line 295: Maybe instead generic `payload` use `test_id_test_run_tuples_list` showing what payload should contain?

argus/backend/service/testrun.py
```diff
@@ -409,6 +411,25 @@ def get_github_issues(self, filter_key: str, filter_id: UUID, aggregate_by_issue
...
         return response
 
+    def resolve_run_build_id_and_number_multiple(self, runs: list[tuple[UUID, UUID]]) -> dict[UUID, dict[str, Any]]:
+        test_ids = [r[0] for r in runs]
+        all_tests: list = []
+        for id_slice in chunk(test_ids):
+            all_tests.extend(ArgusTest.filter(id__in=id_slice).all())
+
+        tests: dict[str, ArgusTest] = {str(t.id): t for t in all_tests}
+        buckets = reduce(lambda acc, val: acc[tests[val[0]].plugin_name].append(val[1]) or acc, runs, defaultdict(list))
```
soyacz, line 421: what is a bucket?

frontend/Github/GithubIssue.svelte
```diff
@@ -214,18 +244,65 @@
...
+                            class="btn btn-close"
+                            on:click={() => {
+                                showRuns = false;
+                            }}
+                        ></button>
+                    </div>
+                </div>
+                <div>
+                </div>
+                    {#await resolveRuns(issue.runs)}
```
soyacz, line 273: this is requested only when `showRuns` is true?

frontend/Github/GithubIssue.svelte
```diff
@@ -49,8 +52,32 @@
     export let aggregated = false;
 
     let deleting = false;
+    let showRuns = false;
     let issueState = issue.state;
 
+    let resolvedRuns;
+
+
+    const resolveRuns = async function(runs) {
```
soyacz, line 61: I didn't look into code, but how about other approach: when querying runs for the issue - why not to get required build_id and built_number along? So we don't need to have reslveRuns at all?

frontend/Github/GithubIssue.svelte
k0machi: Yes

frontend/Github/GithubIssue.svelte
k0machi: Will have to do a migration and adjust the table, which will require code like this anyway, so we can do that as a follow-up to speed this up.

soyacz: only if needed. I was curious only.

argus/backend/service/testrun.py
k0machi: In this context it is a bucket of run ids keyed by plugin they use, could use a better name

frontend/Github/GithubIssue.svelte
k0machi: Potentially the current approach can be slow, as we are doing N queries for N runs (as using indices for a query like this is actually slower and gets extra data), one thing we could do here is to use `execute_async` and resolve those futures after queuing all of them, but from what I've checked the current approach is fast enough for up to 100 runs per issue.

soyacz (approved):
LGTM

argus/backend/controller/testrun_api.py
k0machi: Done

argus/backend/controller/testrun_api.py
k0machi: Added a type hint since this one is just an array.

argus/backend/service/testrun.py
k0machi: Renamed to something better.

PR #387, opened 2024-07-09
refactor(plugins/driver_matrix_tests): Multi-stage pipeline for Drivers

### LINKED ISSUE #289: driver_matrix: Multi-stage pipeline tracking
Currently, driver matrix tests rely on filename parsing to determine which test they are running. In addition to that, the submission happens at the end of pipeline, losing some of the potentially useful information - like start time, overall pipeline health is guessed instead.

The replacement mechanism should function akin to SCT and provide following features:

- [x] Maintain a unique job id per matrix run
- [x] Create an empty entity when a run is started
- [x] Submit each test result separately as they complete
- [ ] Use pipeline health indicator to provide general status - Failed, Passed, etc
- [x] Pass the test information (kind, driver) to the API.

k0machi:
This commit reworks the way Driver Matrix Runs are submitted to Argus,
replacing previously one-shot architecture with a multi-stage pipeline
akin to SCT, where driver matrix run first submits itself, and then
submits results sequentially (either a pass or failure), finalizing with
a call that makes backend determine status based on resulting payload.
Patch failures are now supported. Entire logic was moved from the client
to the backend, removing the need to update the client in case of large
changes to the API / submitted data.

Fixes #289

### DIFF
--- a/argus/backend/plugins/driver_matrix_tests/model.py
+++ b/argus/backend/plugins/driver_matrix_tests/model.py
@@ -1,8 +1,12 @@
 from dataclasses import dataclass
 from datetime import datetime
 from functools import reduce
+import logging
+from pprint import pformat
 import re
+from typing import Literal, TypedDict
 from uuid import UUID
+from xml.etree import ElementTree
 from cassandra.cqlengine import columns
 from argus.backend.db import ScyllaCluster
 from argus.backend.models.web import ArgusRelease
@@ -11,6 +15,8 @@ from argus.backend.plugins.driver_matrix_tests.udt import TestCollection, TestSu
 from argus.backend.plugins.driver_matrix_tests.raw_types import RawMatrixTestResult
 from argus.backend.util.enums import TestStatus
 
+
+LOGGER = logging.getLogger(__name__)
 
 class DriverMatrixPluginError(Exception):
     pass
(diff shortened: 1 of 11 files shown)


argus/backend/plugins/driver_matrix_tests/model.py
```diff
@@ -78,6 +151,173 @@ def parse_driver_name(cls, raw_file_name: str) -> str:
...
+
+    def get_driver_info(self, xml_name: str, test_type: TestTypeType) -> dict[str, str]:
+        if test_type == "cpp":
+            filename_re = r"TEST-(?P<driver_name>[\w]*)-(?P<version>[\d\.]*)(\.xml)?"
+        else:
+            filename_re = r"(?P<name>[\w]*)\.(?P<driver_name>[\w]*)\.(?P<proto>v\d)\.(?P<version>[\d\.]*)(\.xml)?"
+
+        match = re.match(filename_re, xml_name)
+
+        return match.groupdict() if match else {}
```
k0machi, line 263: One small thing: In case of failure, we may not have the driver upstream, as it is derived from xml filename here, which means we need to either generate that name on the spot inside driver matrices or replace this logic with a dedicated driver_source parameter, e.g. `datastax` or `scylla`

k0machi: This calculation is relevant when determining run status - it is automatically failed if we're missing upstream + downstream for all drivers, HOWEVER, run is considered failed anyway if there's a patch failure, so might not be needed.

k0machi:
Tested with python submissions so far, needs testing with cpp, java, go and rust matrix results.

soyacz:
> Tested with python submissions so far, needs testing with cpp, java, go and rust matrix results.

link?

fruch:
is there an scylla-pkg PR that goes with this one ?

k0machi:
I will need to first try it locally with one of the matrices we have, as I still need to make them emit a `metadata.json` file for each run, after that I can run it via scylla-pkg and jenkins job.

PR #391, opened 2024-07-15
improvement(build_system_monitor): Use regex for folder scan

### LINKED ISSUE #390: job scanning should identify release folders, and scan them
we shouldn't be doing any manual work, for a new release of scylla in argus.

SCT jobs should be identified, and marked as SCT plug-in, it shouldn't be manual

k0machi:
This commit replaces exact substring match for release folder names with
a regex pattern search for releases, removing the need to manually add
release names into the scanner configuration. It also moves the scanner
configuration out of argus_web.yaml into module file itself, allowing
versioning of the scanner patterns.

Fixes #390

### DIFF
--- a/argus/backend/service/build_system_monitor.py
+++ b/argus/backend/service/build_system_monitor.py
@@ -2,6 +2,7 @@ import logging
 from abc import ABC, abstractmethod
 import jenkins
 import click
+import re
 from flask import current_app
 from flask.cli import with_appcontext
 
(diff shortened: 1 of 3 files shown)


argus/backend/service/build_system_monitor.py
```diff
@@ -68,17 +73,42 @@ def check_filter(self, group_name: str) -> bool:
 
 
 class JenkinsMonitor(ArgusTestsMonitor):
+
+    BUILD_SYSTEM_FILTERED_PREFIXES = [
+        "byo",
+        "releng",
+        "load-test",
+        "sct_repair",
```
fruch, line 81: why are those being filtered ?

argus/backend/service/build_system_monitor.py
```diff
@@ -68,17 +73,42 @@ def check_filter(self, group_name: str) -> bool:
 
 
 class JenkinsMonitor(ArgusTestsMonitor):
+
+    BUILD_SYSTEM_FILTERED_PREFIXES = [
+        "byo",
```
fruch, line 78: why filtering byo jobs ? it's can be quite useful to look at them with argus... (i.e. the SCT ones)

argus/backend/service/build_system_monitor.py
k0machi: I don't quite remember why, it's better to ask @roydahan about it.

argus/backend/service/build_system_monitor.py
k0machi: Ditto

soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #395, opened 2024-07-15
improvement(service/views): Add 'Type' facet filter

k0machi:
This commit adds a new filter type to the view searcher, called 'type'.
To use, specify it inside query like this: `release:5.4 type:group` or
`type:test` to further refine query


roydahan:
Perfect (for now).

roydahan (approved):

PR #396, opened 2024-07-15
improvement(service/views): Ignore disabled tests/groups/releases

k0machi:
This commit makes it so that any item returned by search that is
disabled is removed from the resulting set

### DIFF
--- a/argus/backend/service/views.py
+++ b/argus/backend/service/views.py
@@ -63,6 +63,15 @@ class UserViewService:
         return mapped
 
     def test_lookup(self, query: str):
+        def check_visibility(entity: dict):
+            if not entity["enabled"]:
+                return False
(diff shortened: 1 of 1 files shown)


argus/backend/service/views.py
```diff
@@ -63,6 +63,15 @@ def index_mapper(item: Model, type = "test"):
         return mapped
 
     def test_lookup(self, query: str):
+        def check_visibility(entity: dict):
+            if not entity["enabled"]:
+                return False
+            if entity.get("group") and not entity["group"]["enabled"]:
+                return False
+            if entity.get("release") and not entity["release"]["enabled"]:
```
roydahan, line 71: Good, I had this problem as well :)

roydahan (approved):

PR #397, opened 2024-07-16
refactor(ReleaseScheduler.svelte): Release Assignee Planner rework

k0machi:
This commit reworks Release Planner wholesale, completely redoing the
way the table is structured. It is now similar in structure to the
TestDashboard used by Release Dashboard page and includes several new
features:

* Ability to select a whole group
* Collapseable groups that can be hidden when not needed
* Ability to batch edit a comment
* Ability to edit individual tests on a schedule
* Ability to reassign whole schedule to a different assignee
* Backlinks to the release dashboard

### DIFF
--- a/argus/backend/controller/api.py
+++ b/argus/backend/controller/api.py
@@ -13,7 +13,7 @@ from argus.backend.controller.client_api import bp as client_bp
 from argus.backend.controller.testrun_api import bp as testrun_bp
 from argus.backend.controller.team import bp as team_bp
 from argus.backend.controller.view_api import bp as view_bp
-from argus.backend.service.argus_service import ArgusService
+from argus.backend.service.argus_service import ArgusService, ScheduleUpdateRequest
 from argus.backend.service.user import UserService, api_login_required
 from argus.backend.service.stats import ReleaseStatsCollector
 from argus.backend.models.web import ArgusRelease, ArgusGroup, ArgusTest, User, UserOauthToken
@@ -229,6 +229,8 @@ def release_schedules_submit():
         groups=payload["groups"],
         assignees=payload["assignees"],
         tag=payload["tag"],
+        comments=payload.get("comments"),
+        group_ids=payload.get("groupIds"),
     )
 
     return jsonify({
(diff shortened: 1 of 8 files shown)


k0machi:
Demo...

https://github.com/user-attachments/assets/8f961cd1-5704-4f5b-9ed6-c1fc8ca27e88

argus/backend/controller/api.py
```diff
@@ -253,6 +255,27 @@ def release_schedules_delete():
     })
 
 
+@bp.route("/release/schedules/update", methods=["POST"])
+@api_login_required
+def release_schedule_update():
+    payload = get_payload(request)
+    req = ScheduleUpdateRequest(**payload)
+    service = ArgusService()
+    update_result = service.update_schedule(
```
soyacz, line 264: why not using `ScheduleUpdateRequest` as input parameter to `update_schedule` method?
Also `ScheduleUpdateRequest` could convert to specific types, so later it's not needed to convert to sets and UUID's in method code.

argus/backend/service/argus_service.py
```diff
@@ -234,7 +246,7 @@ def assign_runs_for_scheduled_group(self, schedule: ArgusSchedule, group_id: UUI
                 row.save()
 
     def submit_new_schedule(self, release: str | UUID, start_time: str, end_time: str, tests: list[str | UUID],
-                            groups: list[str | UUID], assignees: list[str | UUID], tag: str) -> dict:
+                            groups: list[str | UUID], assignees: list[str | UUID], tag: str, comments: dict[str, str] | None, group_ids: dict[str, str] | None) -> dict:
```
soyacz, line 249: what are `groups` and `group_ids`?

argus/backend/service/argus_service.py
```diff
@@ -398,11 +423,52 @@ def update_schedule_comment(self, payload: dict) -> dict:
             "newComment": new_comment,
         }
 
+    def update_schedule(self, release_id: UUID | str, schedule_id: UUID | str, old_tests: list[UUID | str], new_tests: list[UUID | str], comments: dict[str, str], assignee: UUID | str):
```
soyacz, line 426: In case of 2 concurrent users updating the same schedule old_tests might be outdated and cause errors later in adding tests. I suppose

argus/backend/service/argus_service.py
k0machi: This method is used by both group and test planner,  so for test planner, `group_ids` are just hints to make it possible to create comments there (could be folded into single parameter, e.g. `comments`), `groups` on the other hand is only used when using Weekly Planner and that is an array of groups to schedule (much like `tests` is an array of tests to schedule).

argus/backend/service/argus_service.py
k0machi: the resulting set in this case will be whatever was removed from old tests which is fine, however `new_tests` could cause duplication, in which case it might be a good idea to first check if we are not creating a second test entity. This isn't that big of a deal, as it will just be removed when schedule is deleted and it doesn't affect logic of assigning.

argus/backend/controller/api.py
k0machi: Since default `@dataclass` can't do that I propose we leave it as is until later refactor

soyacz: ok

soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #399, opened 2024-07-16
fix(build_system_monitor): Handle situations where we can't create test

k0machi:


argus/backend/service/build_system_monitor.py
```diff
@@ -143,9 +143,12 @@ def collect(self):
                         except StopIteration:
                             LOGGER.warning("Test %s for release %s (group %s) doesn't exist, creating...",
                                            job["name"], saved_release.name, saved_group.name)
-                            saved_test = self.create_test(
+                            try:
+                                saved_test = self.create_test(
                                 saved_release, saved_group, job["name"], job["fullname"], job["url"])
```
soyacz, line 148: this could be indented

k0machi: Done

PR #412, opened 2024-07-23
fix(jenkins-triggering): pick the impersonate user according to email

fruch:
argus rebuild is using the github id/login, to trigger job in jenkins while SCT uses by default the user from the primary email if it's part of scylla emails.
this small diffrence is cause confusion for users which thier email is diffrent from thier github login, and caused cloud resources to be forgotten

this change change the why we select the emails out of the github accout to default to scylla emails, and also to user the user from the email when rebuild jobs, so it would be aligned with SCT

### DIFF
--- a/argus/backend/service/jenkins_service.py
+++ b/argus/backend/service/jenkins_service.py
@@ -220,7 +220,9 @@ class JenkinsService:
     def build_job(self, build_id: str, params: dict, user_override: str = None):
         queue_number = self._jenkins.build_job(build_id, {
             **params,
-            self.RESERVED_PARAMETER_NAME: g.user.username if not user_override else user_override
+            # use the user's email as the default value for the requested by user parameter,
+            # so it would align with how SCT default works, on runs not trigger by argus
+            self.RESERVED_PARAMETER_NAME: g.user.email.split('@')[0] if not user_override else user_override
         })
         return queue_number
 
--- a/argus/backend/service/user.py
+++ b/argus/backend/service/user.py
@@ -90,7 +90,10 @@ class UserService:
         except User.DoesNotExist:
             user = User()
             user.username = user_info.get("login")
-            user.email = email_info[-1].get("email")
+            # pick only scylladb.com emails
+            scylla_email = next(iter([email.get("email") for email in email_info if email.get["email"].endswith("@scylladb.com")]), None)
+            primary_email = next(iter([email.get("email") for email in email_info if email.get["primary"] and email.get["verified"]]), None)
+            user.email = scylla_email or primary_email
             user.full_name = user_info.get("name", user_info.get("login"))
             user.registration_date = datetime.utcnow()
             user.roles = ["ROLE_USER"]


argus/backend/service/user.py
```diff
@@ -90,7 +90,10 @@ def github_callback(self, req_code: str) -> dict | None:
         except User.DoesNotExist:
             user = User()
             user.username = user_info.get("login")
-            user.email = email_info[-1].get("email")
+            # pick only scylladb.com emails
+            scylla_email = next(iter([email.get("email") for email in email_info if email.get["email"].endswith("@scylladb.com")]), None)
```
soyacz, line 94: how about existing users? shouldn't we update their emails to conform this too?
E.g by verification if user.email ends with scylladb.com and if not try to reassign it like here? 
Btw. in case of no scylladb email, shouldn't we  abandon user?

argus/backend/service/jenkins_service.py
```diff
@@ -220,7 +220,9 @@ def clone_build_job(self, build_id: str, params: dict[str, str]):
     def build_job(self, build_id: str, params: dict, user_override: str = None):
         queue_number = self._jenkins.build_job(build_id, {
             **params,
-            self.RESERVED_PARAMETER_NAME: g.user.username if not user_override else user_override
+            # use the user's email as the default value for the requested by user parameter,
+            # so it would align with how SCT default works, on runs not trigger by argus
+            self.RESERVED_PARAMETER_NAME: g.user.email.split('@')[0] if not user_override else user_override
```
roydahan, line 225: Could it be that there are users that do not expose their emails?
maybe we should have a fallback to username if we couldn't get email.

k0machi: The users that hide their email will usually have a github bounce email, so that can be used as a fallback (or first email available as a fallback)

k0machi, line 225: I was actually under impression that it's one or the other,  has something changed?

k0machi: e.g. I'm pretty sure runs started by me from jenkins use my github username in the email reports

roydahan: for me it was roydahan instead of roy

fruch: I did a fallback to the email mentioned as primary, if couldn't find a scylla email

anyhow I'm good with people that have not email configured, they won't be able to rebuild.
i.e. no name for list on resources, no runs for you...

they can fix that on their own, and set the email in user preferences.

fruch: @k0machi notting change, it was wrong assumption to begin with, SCT was using email for long long time, this code was using the github login name, which for lots of people isn't the same.

roydahan: The question is what error they will have and if it'll be clear enough for them to understand that.
I'm pretty sure there are many users without any email exposed.

fruch: it can easily check, if we have such user in our DB...

argus/backend/service/user.py
fruch: I don't know if we should drop or not, but already check if they belongs to the github organization

anyhow we could check the current state of user, and see if fixes are needed or not

k0machi (approved):

PR #414, opened 2024-07-24
fix(argus/service): Remove implicit 10,000 entity limit on full scans

k0machi:
This commit changes two instances of full scans currently used in Argus
to instead request unlimited amount of rows, rather than defaulting to
standard 10,000. This fixes issues where tests would be missing in
search queries or during build system scanner.


fruch:
@k0machi 

can you elaborate which issue this is solving ?

k0machi:
> @k0machi
> 
> can you elaborate which issue this is solving ?

Not being able to find `longevity-gce-custom-d1-worklod2-hybrid-raid-test` on production for 2024.2

PR #415, opened 2024-07-24
fix(service/user): Check if user is None

k0machi:
Fixes an issue where server would return 500 for non-logged in users.


soyacz (approved):
LGTM

fruch (approved):

Issue #417, opened 2024-07-26
Broken results table when columns/rows are missing


soyacz:
When result table misses a column or row it breaks:
![image](https://github.com/user-attachments/assets/d336b66d-626e-49f9-b458-68fc798b2097)

In case results don't contain given column/row it should be not present.
In case only one value is missing (for given row&column pair) it should be grayed out.

PR #418, opened 2024-07-26
fix(results): fix results tables when missing data

soyacz:
In case some columns/rows are missing table breaks (wrong data in columns or looks just bad).

Fix by revamping the way we pass data to renderer: transform it to more organized data structure with known columns and rows that are present and properly handling missing values.

fixes: https://github.com/scylladb/argus/issues/417

### DIFF
--- a/frontend/TestRun/ResultsTab.svelte
+++ b/frontend/TestRun/ResultsTab.svelte
@@ -1,10 +1,70 @@
 <script>
     import {onMount} from "svelte";
+    import {ResultCellStatusStyleMap} from "../Common/TestStatus";
 
     let fetching = true;
     export let id = "";
     export let test_id = "";
     let results = [];
+
+    const transformToTable = (tableData) => {
+        // Transform table data to a more usable format for rendering: {table_name: {description, table_data, columns, rows}}
+        // where table_data: {row_name: {column_name: {value, status, type}}}
+        let transformedData = {};
+
+        tableData.forEach((table) => {
+            const tableName = table.meta.name;
+            const tableDescription = table.meta.description;
+            const columnTypesMap = table.meta.columns_meta.reduce((map, colMeta) => {
+                map[colMeta.name] = colMeta.type;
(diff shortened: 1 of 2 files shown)


soyacz:
before fixing:
image

after fixing:
fixed-table

and example with missing single value:
Screenshot from 2024-07-26 09-33-59

frontend/TestRun/ResultsTab.svelte
```diff
@@ -43,18 +100,15 @@
...
 
-    const getColumnType = (result, columnName) => {
-        return result.meta.columns_meta.find((col) => col.name === columnName).type;
-    };
-
     onMount(() => {
         fetchResults();
     });
 
     const styleMap = {
```
k0machi, line 107: Const values ideally should be uppercase split by underscore: `STYLE_MAP`. Also you can move this into `Common/TestStatus.js` and import it here, that's where a lot of those helper objects are.

frontend/TestRun/ResultsTab.svelte
```diff
@@ -6,17 +6,74 @@
     export let test_id = "";
     let results = [];
 
+    const transform_to_table = (table_data) => {
```
k0machi, line 9: should be `transformToTable` here (and `tableData`)

k0machi (commented):
LGTM, aside from minor style comments

frontend/TestRun/ResultsTab.svelte
soyacz: python style 😆

frontend/TestRun/ResultsTab.svelte
soyacz: fixed

frontend/TestRun/ResultsTab.svelte
soyacz: fixed

k0machi (approved):

PR #426, opened 2024-08-02
feature(argus/client): Endpoint for retrieving the run

k0machi:
This commit adds a new endpoint and a client method to retrieve
currently instantiated run data. This is useful in case client wants to
check/validate run before doing something to it.

Task: scylladb/scylla-cluster-tests#7483


fruch (approved):
LGTM

soyacz (approved):

PR #427, opened 2024-08-08
improvement(test-data): download full releases data

soyacz:
Added possibility to download whole release data and import it without errors. Previous script didn't work if release/group/job was not imported previously.

### DIFF
--- a/argus/backend/util/config.py
+++ b/argus/backend/util/config.py
@@ -10,7 +10,7 @@ LOGGER = logging.getLogger(__name__)
 class Config:
     CONFIG = None
     CONFIG_PATHS = [
-        Path("./config/argus_web.yaml"),
+        Path(__file__).parents[3] / "config" / "argus_web.yaml",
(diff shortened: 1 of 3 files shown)


argus/backend/util/config.py
```diff
@@ -12,6 +12,7 @@ class Config:
     CONFIG_PATHS = [
         Path("./config/argus_web.yaml"),
         Path("argus_web.yaml"),
+        Path("../config/argus_web.yaml"),
```
fruch, line 15: wouldn't it be easy to just usg globbing of something like that ?

fruch (approved):
LGTM

argus/backend/util/config.py
soyacz: I didn't think much here, just added a line when testing.
Your'e right, this could be done better - I'll do it later today.

fruch (approved):

PR #433, opened 2024-08-14
fix(results): Fix results ordering and visibility

soyacz:
In case we have multiple tables, their order is random.
Added ordering based on which table values were inserted first.

Results were having to little margins and looking too messy.
Added filtering capabilities, increased margins and shown a vertical bar
on the left side of each table with color based on results (red if any
result was failed).

### DIFF
--- a/argus/backend/service/testrun.py
+++ b/argus/backend/service/testrun.py
@@ -13,7 +13,7 @@ from flask import g
 from cassandra.query import BatchStatement, ConsistencyLevel
 from cassandra.cqlengine.query import BatchQuery
 from argus.backend.db import ScyllaCluster
-from argus.backend.models.result import ArgusGenericResultMetadata, ArgusGenericResultData
+from argus.backend.models.result import ArgusGenericResultMetadata
 
 from argus.backend.models.web import (
     ArgusEvent,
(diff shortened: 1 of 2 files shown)


soyacz:
image

fruch (approved):
LGTM

argus/backend/service/testrun.py
```diff
@@ -307,23 +307,27 @@ def get_run_events(self, run_id: UUID):
         }
         return response
 
-    def fetch_results(self, test_id: UUID, run_id: UUID) -> dict:
-        query_fields = ["column", "row", "value", "status"]
+    def fetch_results(self, test_id: UUID, run_id: UUID) -> list[dict]:
+        session = ScyllaCluster.get_session()
```
k0machi, line 311: you should probably use `ScyllaCluster.prepare` to get a `PreparedStatement` statement instead

k0machi (changes requested):

argus/backend/service/testrun.py
soyacz: true, fixed, please take  a look again

soyacz:
when merged we can deploy it (along with graphs support)

argus/backend/service/testrun.py
k0machi: Ah, I meant this [helper method](https://github.com/scylladb/argus/blob/master/argus/backend/db.py#L70-L75) as it also saves this statement so you don't have to on every query.

soyacz: ok, I see, fixed.

k0machi (approved):

PR #436, opened 2024-08-27
fix(results-table): pass 0 value to formatValue instead string

soyacz:
When result table contains 0 value, it breaks because empty string is passed to `formatValue` function in this case.
This causes failure to conversion to string (and anyway is wrong).

### DIFF
--- a/frontend/TestRun/ResultsTab.svelte
+++ b/frontend/TestRun/ResultsTab.svelte
@@ -96,6 +96,9 @@
     };
 
     const formatValue = (value, type) => {
+        if (value === null || value === undefined) {
+            return "N/A";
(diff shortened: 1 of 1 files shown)


soyacz:
@k0machi can you also take a look at this one - it's a bug preventing showing results in perf-simple-query tests

k0machi:
> @k0machi can you also take a look at this one - it's a bug preventing showing results in perf-simple-query tests

You have a merge conflict here

soyacz:
> > @k0machi can you also take a look at this one - it's a bug preventing showing results in perf-simple-query tests
> 
> You have a merge conflict here

fixed

PR #441, opened 2024-08-29
fix(graphs): hide button if no graphs available

soyacz:
In case graphs for given test are not available, showing graph button is misleading. This may cause people get used to missing graphs and never look there.

Fix is about showing graphs button only when there are results. Results are verified by metadata tables being present for given test.


soyacz:
See artifact test missing graph icon:
image

k0machi (approved):

fruch (approved):

PR #442, opened 2024-08-30
fix(graphs): adjust filters similar to result tables

soyacz:
Graph filters are adjusting according to selections - this behavior differs from one in results tables and confuses users. Also it's harder to switch between different graphs (as requires deselect previous filter to show the other ones).

Adjusted graphs filters to work similarly like in results tables. Graphs thou have another level of filtering: columns (as each table column is a separate graph). These filters are not exclusive and users can select multiple of them - to show interesting multiple columns only.

### DIFF
--- a/frontend/TestRun/ResultsGraph.svelte
+++ b/frontend/TestRun/ResultsGraph.svelte
@@ -22,6 +22,7 @@
                 }
             }
         };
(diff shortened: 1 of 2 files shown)


k0machi (approved):

fruch (approved):
LGTM

PR #444, opened 2024-09-04
fix(results): fix sut_date for upgrade tests

soyacz:
In upgrade tests results should be mapped to upgraded scylla version instead of base one.

Fixed by first getting `scylla-server-upgraded` package date and only fallback to `scylla-server` when not available.
Also fixed issue with timestamp stored in db on local env.


fruch (approved):
LGTM

PR #447, opened 2024-09-09
improvement(notification_manager): Add "Assigned to a run" notification

### LINKED ISSUE #398: People should get Email/Notification when they are assigned to jobs
right now we need to assign people, and write something for them on the discussion, to them to notice it.

It's hard to do when it's a mass of test, every assignment on a run (not a job) should send a notification/email

k0machi:
This commit adds both internal and email notification for assignments to
runs within argus. Manual assignments done by a user on a test run card
send a notification/email from him.

Fixes #398

### DIFF
--- a/argus/backend/service/client_service.py
+++ b/argus/backend/service/client_service.py
@@ -1,4 +1,5 @@
 from uuid import UUID
+
 from argus.backend.db import ScyllaCluster
 from argus.backend.models.result import ArgusGenericResultMetadata, ArgusGenericResultData
 from argus.backend.plugins.core import PluginModelBase
@@ -25,6 +26,7 @@ class ClientService:
     def submit_run(self, run_type: str, request_data: dict) -> str:
         model = self.get_model(run_type)
         model.submit_run(request_data=request_data)
+
         return "Created"
     
     def get_run(self, run_type: str, run_id: str):
--- a/argus/backend/service/notification_manager.py
+++ b/argus/backend/service/notification_manager.py
@@ -93,7 +93,8 @@ class NotificationSenderBase:
 
 class ArgusDBNotificationSaver(NotificationSenderBase):
     CONTENT_TEMPLATES = {
-        ArgusNotificationTypes.Mention: lambda p: render_template("notifications/mention.html.j2", **p if p else {})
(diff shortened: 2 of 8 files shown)


argus/backend/service/client_service.py
```diff
@@ -24,7 +30,24 @@ def get_model(self, run_type: str) -> PluginModelBase:
 
     def submit_run(self, run_type: str, request_data: dict) -> str:
         model = self.get_model(run_type)
-        model.submit_run(request_data=request_data)
+        run = model.submit_run(request_data=request_data)
+        if run.assignee and run.test_id:
+            self.notification_manager.send_notification(
+                receiver=run.assignee,
+                sender=g.user.id,
```
soyacz, line 37: what is g.user.id in the scope of ArgusClient?

argus/backend/service/client_service.py
```diff
@@ -14,6 +19,7 @@ class ClientService:
     PLUGINS = {name: plugin.model for name, plugin in AVAILABLE_PLUGINS.items()}
 
     def __init__(self) -> None:
+        self.notification_manager = NotificationManagerService()
```
soyacz, line 22: why not to create it in submit_run method if needed only?

k0machi: Could be useful for other methods in the future.

argus/backend/service/client_service.py
k0machi: It's the ClientService scope of argus application and `g` is the global flask context variable.

Anyway I'll remove this notification in particular, as now I think it's a bit too excessive to have a notification created on *every single* run submitted to argus.

soyacz: yes, might be too excessive and people will ignore it. Please remove unrelated imports then too.
I think would be good to send once a day a reminder about jobs to investigate.

k0machi: Done

roydahan (approved):
LGTM

soyacz (approved):
LGTM

fruch (approved):
I would suggest to update description and commit message that it's only doing it when update is happening

k0machi:
> I would suggest to update description and commit message that it's only doing it when update is happening

Forgot to update the description on PR, already updated commit message.

PR #456, opened 2024-09-17
improvement(results): track best and validate results

soyacz:
Results to make difference need to change test status depending if are passing or failing. At first idea was to force clients to adjust test status. This commit adds possibility to Argus to track best results for given test and validate incoming results based on rules defined by Argus clients.

List of improvements:
1. Argus tracks best result for each column&row pair in tables, including `run_id` and `result_date` for future reference. Evaluated only when `higher_is_better` column metadata is set. Stored in separate table to track the changes (and correctly show validation rules calculated limits for given time)
2. Argus client may provide validation rules for each column. Supported validations (can set multiple of them):
- fixed_value - verifying if value is better than it
- best_pct - verifying if value is better than margin percent from best value
- best_abs - vefifying if value is better than margin absolute from best value Validation rules are tracked for future reference.
3. Argus evaluates each submitted cell against current validation rules and best result for given sut_timestamp. Marks cell as ERROR if rule finds its value below any rule limit. Validation is skipped if cell status is other than `UNSET`.

refs: https://github.com/scylladb/qa-tasks/issues/1765

### DIFF
--- a/argus/backend/models/result.py
+++ b/argus/backend/models/result.py
@@ -1,12 +1,21 @@
+import math
+from datetime import datetime, timezone
+
 from cassandra.cqlengine import columns
 from cassandra.cqlengine.models import Model
 from cassandra.cqlengine.usertype import UserType
 
+class ValidationRules(UserType):
+    valid_from = columns.DateTime()
+    best_pct = columns.Double()  # max value limit relative to best result in percent unit
+    best_abs = columns.Double()  # max value limit relative to best result in absolute unit
+    fixed_limit = columns.Double()  # fixed limit
 
 class ColumnMetadata(UserType):
     name = columns.Ascii()
     unit = columns.Text()
     type = columns.Ascii()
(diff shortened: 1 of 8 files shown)


soyacz:
Ready for review, in the meantime I'll try to work on unit tests for that as it has many edge cases.

argus/backend/models/result.py
```diff
@@ -1,12 +1,23 @@
+import math
+from datetime import datetime, timezone
+
 from cassandra.cqlengine import columns
 from cassandra.cqlengine.models import Model
 from cassandra.cqlengine.usertype import UserType
+import logging
 
+logger = logging.getLogger(__name__)
```
k0machi, line 9: For consistency, constants should be all uppercase.

k0machi (approved):
Looks good, aside from a minor style issue.

fruch (approved):
LGTM

PR #460, opened 2024-09-24
fix(results): display and submit results fixes

soyacz:
Perf Simple Query results were not displayed due Cell.svelte error on `null` value. Also fixed table error status when there's UNSET status cell value.
And most important, fixing regression in submit_results when submitting tables with text values.


soyacz:
@k0machi sorry for merging without approve, but needed for hot-fixing my previous PR. 
Let me know if something needs to be changed here.

PR #463, opened 2024-09-25
fix(client): don't require ValidationRules in results tables

soyacz:
ValidationRules are optional and should not break the code when not provided.


fruch (approved):
LGTM

PR #469, opened 2024-09-27
fix(results): fix ordering of best results table clustering keys

soyacz:
In future we're going to need to fetch best results for given point in time. Currently it's not efficient as would require reading rows that happened after requested point in time. This is because 'result_date' where clause restriction must happen along with 'key' restriction.

To tackle that problem, need to reorder 'key' and 'result_date' columns. Better to do it early, before we have data there, so we don't need to migrate anything.

### DIFF
--- a/argus/backend/models/result.py
+++ b/argus/backend/models/result.py
@@ -129,10 +129,10 @@ class ArgusGenericResultData(Model):
     status = columns.Ascii()
 
 class ArgusBestResultData(Model):
-    __table_name__ = "generic_result_best_v1"
+    __table_name__ = "generic_result_best_v2"
     test_id = columns.UUID(partition_key=True)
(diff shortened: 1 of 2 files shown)


argus/backend/models/result.py
```diff
@@ -129,10 +129,10 @@ class ArgusGenericResultData(Model):
     status = columns.Ascii()
 
 class ArgusBestResultData(Model):
-    __table_name__ = "generic_result_best_v1"
+    __table_name__ = "generic_result_best_v2"
```
fruch, line 132: This means we need to migrate current data ?

soyacz: teoretically yes, practicaly no - client was not updated and 'higher_is_better' param not set (thus no best result evaluation done).

k0machi (approved):

Issue #468, opened 2024-09-27
Add and execute unit (integration) tests


soyacz:
Current unit-tests in Argus repo are outdated, referencing broken code.
We need to gradually add unit tests to protect from introducing regressions.
Also to make it effective, we need to have CI for them and execute for each PR.

PR #471, opened 2024-10-04
improvement(tests): added unit tests for generic results

soyacz:
Ressurrecting unit-tests for backend. At first, introduced tests for generic results feature. Found few bugs and unneeded code along.

Tests are relying on containerized scylladb cluster (1 node) and reuse it for next tests runs for speed testing. Tests are designed in a way that can be run on the same db (without truncation). Although, there's an unused `truncate_all_tables` function in case it's needed (it adds few secs of overhead, so better not to use it).

refs: https://github.com/scylladb/argus/issues/468

### DIFF
--- a/argus/backend/cli.py
+++ b/argus/backend/cli.py
@@ -15,6 +15,9 @@ LOGGER = logging.getLogger(__name__)
 @click.command('sync-models')
 @with_appcontext
 def sync_models_command():
+    sync_models()
+
+def sync_models():
     cluster = ScyllaCluster.get()
     cluster.sync_core_tables()
     LOGGER.info("Synchronizing plugin types...")
--- a/argus/backend/models/result.py
+++ b/argus/backend/models/result.py
@@ -1,15 +1,16 @@
-import math
 from datetime import datetime, timezone
 
 from cassandra.cqlengine import columns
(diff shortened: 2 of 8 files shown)


k0machi (approved):
LGTM, but I think we still need to clean up old tests / conftest setup

soyacz:
> LGTM, but I think we still need to clean up old tests / conftest setup

Yes, are there any relevant test cases that could stay?

k0machi:
> > LGTM, but I think we still need to clean up old tests / conftest setup
> 
> Yes, are there any relevant test cases that could stay?

No, all of them can be removed. As I mentioned before, they all test `argus.db` module, which is no longer used.

soyacz:
> > > LGTM, but I think we still need to clean up old tests / conftest setup
> > 
> > 
> > Yes, are there any relevant test cases that could stay?
> 
> No, all of them can be removed. As I mentioned before, they all test `argus.db` module, which is no longer used.

Ok, will remove it with other PR that enables tests in CI

soyacz:
@k0machi did you try running them

k0machi:
> @k0machi did you try running them?

No, I haven't

PR #476, opened 2024-10-08
improvement(results): add limit lines to graphs

soyacz:
Added limit lines to graphs based on best results and validation rules. When creating graphs, for each point limit is calculated and plotted (also visible in the tooltip).
Have in mind that limits might differ from ones in "Results" tab as client might ignore rules (which most of our test cases currently do).

closes: https://github.com/scylladb/qa-tasks/issues/1763

### DIFF
--- a/argus/backend/service/results_service.py
+++ b/argus/backend/service/results_service.py
@@ -2,6 +2,7 @@ import copy
 import logging
 import math
 import operator
+from collections import defaultdict
 from datetime import datetime, timezone
 from functools import partial
 from typing import List, Dict, Any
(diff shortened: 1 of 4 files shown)


soyacz:
our most complex example*:
image

*It won't be visible yet like that on prod, due `ValidationRules` are not yet applied to latency decorator results or perf simple query results.
Also it should vary when new best result appears (or validation rule) but it is not working on my test data as they're imported (and best result date - `now` - is always greater than `SUT date`)

soyacz:
As per discussion today with @fruch , I made it to print only one limit line if 'fixed limit' is set (as it is the same for all the series).

soyacz:
@fruch @k0machi ping

fruch (approved):
LGTM

k0machi (approved):

PR #477, opened 2024-10-09
improvement(CI): add GH action to run tests

soyacz:
Run unit tests in GH action for each PR.

closes: https://github.com/scylladb/argus/issues/468


soyacz:
`succeeded now in 1m 21s`

k0machi (approved):

.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
fruch, line 28: didn't we moved to python 3.12 ?

soyacz: I'm not aware of it and `pyproject.toml` has 3.10 specified

fruch: > Production now runs 3.12 yes, toml currently specifies minimum as 3.10. We should probably test with 3.12 and maybe bump that version as well.

Not probably, we should strive for that code and tests in the repo are done with the same environment as production.

Not doing so, means we can easily break production in some code path, that we won't see, even if there is a test that covers it.

soyacz: ok, I updated it, with minimal dependency changes.

fruch (approved):
LGTM

PR #478, opened 2024-10-09
improvement(frontend): Add frontend endpoint for plugin/run_id

k0machi:
This commit adds a new endpoint which replaces hard to use
/test/test_id/runs?additionalRuns[]=run_id endpoint into a much more
simpler /tests/plugin_name/run_id endpoint, allowing to build a link to
the test run without knowing test_id in advance and simplifying
sharing/searching for a run.

Part of Task: scylladb/qa-tasks#1490

### DIFF
--- a/frontend/TestRun/TestRun.svelte
+++ b/frontend/TestRun/TestRun.svelte
@@ -133,7 +133,7 @@
     <div class="d-flex px-2 py-2 mb-1 border-bottom bg-white ">
         <div class="p-1">
             {#if testRun}
-                <a class="link-dark" href="/test/{testRun.test_id}/runs?additionalRuns[]={testRun.id}">
+                <a class="link-dark" href="/tests/{testInfo.test.plugin_name}/{testRun.id}">
                     {testRun.build_id}#{buildNumber}
                 </a>
             {/if}
--- a/argus/backend/controller/main.py
+++ b/argus/backend/controller/main.py
@@ -9,6 +9,7 @@ from argus.backend.controller.notifications import bp as notifications_bp
 from argus.backend.controller.team_ui import bp as teams_bp
 from argus.backend.service.argus_service import ArgusService
 from argus.backend.models.web import WebFileStorage
+from argus.backend.service.testrun import TestRunService
 from argus.backend.service.user import UserService, login_required
 from argus.backend.service.views import UserViewService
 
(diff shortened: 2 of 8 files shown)


k0machi:
Demo
image

soyacz (approved):
what happens if two runs are selected?
Besides, I can't wait to see it merged :)
LGTM

k0machi:
> what happens if two runs are selected? Besides, I can't wait to see it merged :) LGTM

In both cases it wasn't possible to directly link to two runs in frontend. While the old endpoint supports array declaration `/test/test_id/runs?additionalRuns[]=run_A&additionalRuns[]=run_B`, you could build those manually but the functionality for that on the frontend UI side never existed.

frontend/TestRun/TestRun.svelte
```diff
@@ -133,7 +133,7 @@
     <div class="d-flex px-2 py-2 mb-1 border-bottom bg-white ">
         <div class="p-1">
             {#if testRun}
-                <a class="link-dark" href="/test/{testRun.test_id}/runs?additionalRuns[]={testRun.id}">
+                <a class="link-dark" href="/tests/{testInfo.test.plugin_name}/{testRun.id}">
```
vponomaryov, line 136: Why the `/{testInfo.test.plugin_name}` part is needed?
I guess it will be much more useful to have release name than just testing tool name.

For example:

`tests/releases/enterprise-2024.2/%test-id%` says clearly about the belonging of a test run to a release.
So it will be easy to dostinguish from the following:
`tests/staging/valerii/%test-id%`.

k0machi: > Why the `/{testInfo.test.plugin_name}` part is needed? I guess it will be much more useful to have release name than just testing tool name.
> 
> For example:
> 
> `tests/releases/enterprise-2024.2/%test-id%` says clearly about the belonging of a test run to a release.
> 
> So it will be easy to distinguish it from the following: `tests/staging/valerii/%test-id%`.
> 
> Which I personally find useful - we get info not even opening the link.

Testing tool is needed (or test-id, as it was previously) to determine which result table we query for those results, so it would have to stay. The release marker can be introduced, as an alternative URL, but release name is not something which can be guessed from testing frameworks themselves, aside from making them query the run. For argus, we can add another route which would act similarly to this one, like `/release/scylla-master/tests/scylla-cluster-tests/$id` or `/tests/scylla-cluster-tests/scylla-master/folder-name/test-name/$id`

fruch: For the use case of getting from emails or elasticsearch data into Argus run

What's in this PR fits best, introducing more data into the url, complicates how we can build it where needed.

As suggested we can have more options for better sharing out of Argus, but we still need a way to get from test-id to Argus page

PR #479, opened 2024-10-14
fix(results): show filters when there are at least 2 in level

soyacz:
When there's only 2 levels (groups) of filtering, filters are not shown. This is wrong assumption, e.g. In gradual throughput test when we have multiple result tables for each throughput with only latencies results.

Show filters when one filter level contains more than 1 filter.

### DIFF
--- a/frontend/TestRun/ResultsTab.svelte
+++ b/frontend/TestRun/ResultsTab.svelte
@@ -15,6 +15,7 @@
     let selectedFilters = [];
     let filteredTables = [];
     let selectedScreenshot = "";
+    let showFilters = false;
 
     const tableStyleToColor = {
         "table-success": "green",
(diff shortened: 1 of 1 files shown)


juliayakovlev (approved):
I have no opinion about correctness. I looks OK for me

frontend/TestRun/ResultsTab.svelte
```diff
@@ -177,14 +178,20 @@
                 fltrs.get(level).add(part.trim());
             });
         });
-        filters = Array.from(fltrs.entries()).sort((a, b) => a[0] - b[0]).map(entry => ({
-            level: entry[0],
-            items: Array.from(entry[1])
-        }))
+
+        filters = Array.from(fltrs.entries())
+            .sort((a, b) => a[0] - b[0])
```
k0machi, line 183: Small tip here, you can use `([left], [right]) => left - right` form to destructure array into named values

frontend/TestRun/ResultsTab.svelte
```diff
@@ -177,14 +178,20 @@
...
+            .map(entry => ({
```
k0machi, line 184: Same here, `.map(([level, items]) => { level, items: Array.from(items) })`

frontend/TestRun/ResultsTab.svelte
```diff
@@ -300,7 +307,7 @@
 
 {#if !fetching }
     <div class="p-2">
-        <div class="filters-container {Object.keys(filters).length < 2 ? 'd-none' : ''}">
+        <div class="filters-container {showFilters ? '' : 'd-none'}">
```
k0machi, line 310: This can be a svelte class declaration instead: `class:d-none={!showFilters}`

k0machi (changes requested):
Overall looks good, just some small comments

soyacz:
thanks @k0machi , fixed your comments

k0machi (approved):

Issue #462, opened 2024-09-25
Add ability to exclude result


soyacz:
Sometimes result sent to argus might be wrong, and break how graphs look (introducing unnecessary noise) or spoil "best result" and affecting validation.
We need a method to remove such results easily.
Open questions:
1. Should it be per run (e.g. removing/disabling whole run result) or single one?
2. Shall it revalidate all the next results (if 'best_result' was affected)?

My proposal is to add a button in "results" tab to mark all tables as "spoiled" (something like "Disable" button with graph icon).

soyacz:
@fruch , @juliayakovlev proposed to use `IGNORED` job status to exclude result.
I think it's good idea. WDYT?
Also, I'll need to push https://github.com/scylladb/argus/pull/476 forward as it messes with the same parts of code (and some will be reused) to make this one happen.

juliayakovlev:
> [@fruch](https://github.com/fruch) , [@juliayakovlev](https://github.com/juliayakovlev) proposed to use `IGNORED` job status to exclude result. I think it's good idea. WDYT? Also, I'll need to push [#476](https://github.com/scylladb/argus/pull/476) forward as it messes with the same parts of code (and some will be reused) to make this one happen.

Actually it's Israel's idea )

fruch:
> > [@fruch](https://github.com/fruch) , [@juliayakovlev](https://github.com/juliayakovlev) proposed to use `IGNORED` job status to exclude result. I think it's good idea. WDYT? Also, I'll need to push [#476](https://github.com/scylladb/argus/pull/476) forward as it messes with the same parts of code (and some will be reused) to make this one happen.
> 
> Actually it's Israel's idea )

it was @roydahan idea, we are just passing along the message :)

fruch:
I think matches the reason of what is ignore meant to be, and not we have a practical usage of it.

it solve the UX around this one, kind of nicely.
and yet again introduce a reason for us to load the runs data. (we have plenty of those reasons)

PR #481, opened 2024-10-15
improvement(results): exclude ignored results from graphs

soyacz:
Sometimes results might be wrong (e.g. wrong job configuration) leading to results that outstand and being irrelevant or updating best result value causing validation to break.

If job investigation status is set to `IGNORED` make Argus to skip plotting it and exclude from best results map.

closes: https://github.com/scylladb/argus/issues/462

### DIFF
--- a/argus/backend/service/results_service.py
+++ b/argus/backend/service/results_service.py
@@ -4,13 +4,14 @@ import math
 import operator
 from collections import defaultdict
 from datetime import datetime, timezone
-from functools import partial
+from functools import partial, cache
 from typing import List, Dict, Any
 from uuid import UUID
 
 from dataclasses import dataclass
 from argus.backend.db import ScyllaCluster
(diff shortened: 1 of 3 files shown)


argus/backend/service/results_service.py
```diff
@@ -197,6 +198,17 @@ class ResultsService:
     def __init__(self):
         self.cluster = ScyllaCluster.get()
 
+    @cache
+    def _ignored_results(self, test_id: UUID) -> list[str]:
```
fruch, line 202: When is this cache gonna be cleared ?

When Argus restarts ?

soyacz: no, it's per request - will be needed when we merge PR with drawing limit lines, where best results are also used, to avoid double query db.

k0machi (approved):
LGTM

PR #482, opened 2024-10-16
improvement(results): respond with error when result failed

soyacz:
In case there is a cell with 'ERROR' status, then respond with `error` status, so client may take proper action based on it (e.g. marking test as failed).

refs: https://github.com/scylladb/qa-tasks/issues/1765

### DIFF
--- a/argus/backend/service/client_service.py
+++ b/argus/backend/service/client_service.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from uuid import UUID
 
 from argus.backend.db import ScyllaCluster
+from argus.backend.error_handlers import DataValidationError
 from argus.backend.models.result import ArgusGenericResultMetadata, ArgusGenericResultData
 from argus.backend.plugins.core import PluginModelBase
 from argus.backend.plugins.loader import AVAILABLE_PLUGINS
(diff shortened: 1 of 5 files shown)


soyacz:
This requires a change in SCT to handle possible error and act accordingly (raise error event, so later test is failed).

soyacz:
this one should be merged after https://github.com/scylladb/scylla-cluster-tests/pull/9011

argus/backend/service/client_service.py
```diff
@@ -111,12 +111,24 @@ def submit_results(self, run_type: str, run_id: str, results: dict) -> dict[str,
...
+                result_failed = True
             ArgusGenericResultData(test_id=run.test_id,
                                    run_id=run.id,
                                    name=table_name,
                                    sut_timestamp=sut_timestamp,
                                    **asdict(cell)
                                    ).save()
-        return {"status": "ok", "message": "Results submitted"}
+        if result_failed:
+            response = {"status": "error",
```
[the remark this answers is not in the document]
soyacz: done. See I adjusted a bit what we log in that situation: when error is a subclass of `APIException` then log with info level without traceback.
Possibly we'll need to update in relevant places (as I don't think we need traceback for most of them).

argus/backend/service/client_service.py
```diff
@@ -109,12 +110,19 @@ def submit_results(self, run_type: str, run_id: str, results: dict) -> dict[str,
...
             ArgusGenericResultData(test_id=run.test_id,
                                    run_id=run.id,
                                    name=table_name,
                                    sut_timestamp=sut_timestamp,
                                    **asdict(cell)
                                    ).save()
-        return {"status": "ok", "message": "Results submitted"}
+        if result_failed:
+            raise DataValidationError()
+        else:
```
k0machi, line 126: Just to clean this up, you can now remove the `else:` and just leave `return {...}` statement

soyacz: done

Issue #480, opened 2024-10-15
Stop including backend source code in client library


soyacz:
Client library contain references to backend code. We should remove them and make build process include only client code.

PR #484, opened 2024-10-16
improvement(cleanup): remove backend from argus client

soyacz:
Some cleanup work had to be done to better structure argus client for less problematic management:
* Removed old db stuff, that cause confusion when searching for db/models related things. Along with old migration and prod dump scripts.
* moved common modules for backend and client to `common` module next to `backend` and `client` modules
* excluded `backend` module from client package. Including backend might cause import error due backend-only required packages
* created unit tests for argus client package - build, install, basic import

closes: https://github.com/scylladb/argus/issues/480


soyacz:
@k0machi WDYT?

k0machi (approved):

Issue #473, opened 2024-10-08
Filter graphs data by release


soyacz:
In performance regression tests we mix release version with 'master' versions. We'd like to filter per release/master all the points to see regressions for specified release.

soyacz:
I got 2 ideas:
1. One series for all release versions:
* Each point on chart will also contain version info (e.g. 2024.1.10, dev, etc.). 
* new filter buttons will be added to filter 'major' releases e.g. `2024.1` `2024.2` `dev` (for master). rc versions also will be put under major release. Selecting filter will hide points accordingly.
* each point marker for 'official' release will get different point marker (e.g `x`) (I didn't confirm yet if can be done) while 'dev' ones will be marked by circle
pros: will have less lines on graphs
cons: hard to compare version to version as all points belong to one line

2. One series per each release:
* Series will be segregated by version (e.g. 2024.1, 2024.2, dev), basically adding new series for each release.
* no filtering needed, on graphs user may click on series legend to hide it
* distinction by point markers (chartjs have options: circle', 'rect', 'triangle', 'star', 'cross' - so max 5 versions on one graph) , colors the same
pros: nicer comparison version to version (both visible on graph as separate lines), 
cons: for graphs where we have 3 series (e.g. 3 cycles) each version will duplicate that so can get messy after having 3 releases on one graph (9 data series + 3 limit series). also hiding will be more problematic as will require to click on each graph (unless we create also filters for bulk 'hide'/'show')

@israel I'm opting for option 2, but wanted to know your opinion and if I understand the problem&solution correctly.

fruch:
seems like options 1 is simpler to achieve, maybe coloring differently the dot ?

and option 2 can get quite messy quite fast, and we will need lots handy quick filter to help us tame it to our needs.

I would suggest starting with option one, and let talk with @roydahan tomorrow about this one

PR #489, opened 2024-10-23
Support multiple releases on graphs

soyacz:
Performance job can run against different scylla release. Because it was
not taken into account, graph lines don't differentiate between them
making impossible to distinguish different releases.

This commit adds support for mixing different SUT releases in one job
and make them visible as separate series on graphs. Additional
'releases' filters were added to easily toggle releases that should be
shown.

closes: https://github.com/scylladb/argus/issues/473

### DIFF
--- a/argus/backend/service/results_service.py
+++ b/argus/backend/service/results_service.py
@@ -10,10 +10,14 @@ from uuid import UUID
 
 from dataclasses import dataclass
 from argus.backend.db import ScyllaCluster
-from argus.backend.models.result import ArgusGenericResultMetadata, ArgusGenericResultData, ArgusBestResultData
+from argus.backend.models.result import ArgusGenericResultMetadata, ArgusGenericResultData, ArgusBestResultData, ColumnMetadata
+from argus.backend.plugins.sct.udt import PackageVersion
 from argus.backend.service.testrun import TestRunService
 
 LOGGER = logging.getLogger(__name__)
+
+type RunId = str
+type ReleasesMap = dict[str, list[RunId]]
 
 
 @dataclass
@@ -58,6 +62,12 @@ class Cell:
             self.status = "ERROR"
 
 
+@dataclass
+class RunsDetails:
+    ignored: list[RunId]
+    packages: dict[RunId, list[PackageVersion]]
+
+
 default_options = {
     "scales": {
         "y": {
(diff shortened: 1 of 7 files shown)


soyacz:
It's not visible that good as I have randomized test data.  Note the graph legend on each graph, also there's point shape difference between releases. I left colors the same, but I'm not sure, maybe we should change hue or something to additionally differentiate.
Without filters:
image

With filter on:
image

argus/backend/service/results_service.py
```diff
@@ -250,21 +297,64 @@ def calculate_graph_ticks(graphs: List[Dict]) -> dict[str, str]:
...
+            if package.name == main_package:
+                if "dev" in package.version:
+                    major_version = 'dev'
+                else:
+                    major_version = '.'.join(package.version.split('.')[:2])
+                releases_map[major_version].append(run_id)
+    return releases_map
+
+
+def create_chartjs(table: ArgusGenericResultMetadata, data: list[ArgusGenericResultData], best_results: dict[str, List[BestResult]],
```
k0machi, line 322: Regarding charts in general - could you try enabling cubic interpolation for the line, `rect` style for the points (or just make them tiny dots) and `filled` style for the line itself? (with some opacity thrown in for the color) 

The random data kinda showcases worst case scenario and it looks completely unreadable, especially with a lot of data points. Suggestions above could help with that, I feel.

argus/backend/service/results_service.py
```diff
@@ -250,21 +297,64 @@ def calculate_graph_ticks(graphs: List[Dict]) -> dict[str, str]:
...
         plugin_query = self.cluster.prepare("SELECT id, plugin_name FROM argus_test_v2 WHERE id = ?")
         plugin_name = self.cluster.session.execute(plugin_query, parameters=(test_id,)).one()['plugin_name']
         plugin = TestRunService().get_plugin(plugin_name)
-        ignored_runs_query = self.cluster.prepare(f"SELECT id, investigation_status FROM {plugin.model.table_name()} WHERE test_id = ?")
-        ignored_runs = [run["id"] for run in self.cluster.session.execute(ignored_runs_query, parameters=(test_id,)).all()
-                        if run["investigation_status"].lower() == "ignored"]
-        LOGGER.debug(f"Ignored runs for test {test_id}: {ignored_runs}")
-        return ignored_runs
+        ignored_runs_query = self.cluster.prepare(
+            f"SELECT id, investigation_status, packages FROM {plugin.model.table_name()} WHERE test_id = ?")
```
k0machi, line 353: CQLengine supports sparse queries, like so:

```python
model.filter(test_id=test_id).only(["id", "investigation_status", "packages"]).all()
```

I haven't really been using them but in this case they are perfect (as long as you don't need asyncio).

argus/backend/service/results_service.py
```diff
@@ -274,6 +364,14 @@ def _get_tables_metadata(self, test_id: UUID) -> list[ArgusGenericResultMetadata
         tables_meta = self.cluster.session.execute(query=query, parameters=(test_id,))
         return [ArgusGenericResultMetadata(**table) for table in tables_meta]
 
+    def _get_tables_data(self, test_id: UUID, table_name: str, ignored_runs: list[RunId]) -> list[ArgusGenericResultData]:
+        query_fields = ["run_id", "column", "row", "value", "status", "sut_timestamp"]
+        raw_query = (f"SELECT {','.join(query_fields)}"
```
k0machi, line 369: Ditto

k0machi (commented):
Also should probably look into making a nicer color scheme - should probably avoid `255` for anything, check bootstrap main color scheme that we use for something more easy on the eyes.

argus/backend/service/results_service.py
soyacz: I use it mostly to reuse 'prepared' queries, can we do it with your proposed solution?

argus/backend/service/results_service.py
soyacz: I try to play a bit with it and with grid uniformity too.

argus/backend/service/results_service.py
k0machi: Unfortunately CQLEngine seems to always do a `SimpleStatement`, so you're trading some potential performance for convenience.

soyacz: I think convenience is not worth it - I don't find it much more complex than model method.

k0machi (approved):

soyacz:
@fruch after adjusting to @k0machi feedback and testing it on staging, we decided to merge it - if you have more feedback, let me know.

Issue #474, opened 2024-10-08
Filter graph data points by time range


soyacz:
When running test against old SUT, graphs X axis is too broad and disturbs the view.
Add filters: `SUT start date` - `SUT end date` with simple auto fill buttons like: last month, last 3 months, last 6 months

PR #491, opened 2024-10-25
feature(graphs): filter data points by date filters

soyacz:
In order to narrow down graphs range, date filters were added (by start and end date) with ability to quickly switch between 1,3 and 6 recent months. Last 3 months filter is on by default.

closes: https://github.com/scylladb/argus/issues/474

### DIFF
--- a/argus/backend/controller/api.py
+++ b/argus/backend/controller/api.py
@@ -1,4 +1,5 @@
 import logging
+from datetime import datetime, timezone
 from uuid import UUID
 import requests
 from flask import (
@@ -386,13 +387,21 @@ def test_info():
 @api_login_required
 def test_results():
     test_id = request.args.get("testId")
+    start_date_str = request.args.get("startDate")
+    end_date_str = request.args.get("endDate")
+
     if not test_id:
         raise Exception("No testId provided")
+
+    start_date = datetime.fromisoformat(start_date_str).astimezone(timezone.utc) if start_date_str else None
+    end_date = datetime.fromisoformat(end_date_str).astimezone(timezone.utc) if end_date_str else None
+
     service = ResultsService()
     if request.method == 'HEAD':
         exists = service.is_results_exist(test_id=UUID(test_id))
         return Response(status=200 if exists else 404)
-    graphs, ticks, releases_filters = service.get_test_graphs(test_id=UUID(test_id))
+
+    graphs, ticks, releases_filters = service.get_test_graphs(test_id=UUID(test_id), start_date=start_date, end_date=end_date)
 
     return {
         "status": "ok",
--- a/argus/backend/service/results_service.py
+++ b/argus/backend/service/results_service.py
@@ -295,7 +295,10 @@ def calculate_graph_ticks(graphs: List[Dict]) -> dict[str, str]:
                 min_x = first_x
             if max_x is None or last_x > max_x:
                 max_x = last_x
+    if not max_x or not min_x:
+        return {}  # no data
     return {"min": min_x[:10], "max": max_x[:10]}
+
 
 
 def _identify_most_changed_package(packages_list: list[PackageVersion]) -> str:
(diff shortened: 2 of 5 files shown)


soyacz:
It's deployed to staging.
In general, it looks like this:
image

frontend/TestRun/ResultsGraphs.svelte
```diff
@@ -59,6 +61,21 @@
...
 
+    const setDefaultDateRange = () => {
+        const now = new Date();
+        endDate = now.toISOString().split('T')[0];
+        const pastDate = new Date(now.setMonth(now.getMonth() - 3));
+        startDate = pastDate.toISOString().split('T')[0];
+    };
+
+    const setDateRange = (months) => {
+        const now = new Date();
```
k0machi, line 72: You can use `dayjs` package for these date calculations.

frontend/TestRun/ResultsGraphs.svelte
```diff
@@ -12,17 +12,19 @@
...
 
     const dispatch_run_click = (e) => {
-        dispatch("runClick", {runId: e.detail.runId});
+        dispatch("runClick", { runId: e.detail.runId });
     };
 
     const fetchTestResults = async function (testId) {
         try {
-            let res = await fetch(`/api/v1/test-results?testId=${testId}`);
+            let res = await fetch(`/api/v1/test-results?testId=${testId}&startDate=${startDate}&endDate=${endDate}`);
```
k0machi, line 27: `query-string` package can be used here, optional at this stage, but it does save you from potential certain footguns later.

argus/backend/controller/api.py
```diff
@@ -386,13 +387,21 @@ def test_info():
 @api_login_required
 def test_results():
     test_id = request.args.get("testId")
+    start_date_str = request.args.get("startDate")
+    end_date_str = request.args.get("endDate")
+
     if not test_id:
         raise Exception("No testId provided")
+
+    start_date = datetime.fromisoformat(start_date_str) if start_date_str else None
```
k0machi, line 396: Unspecified timezone.

k0machi (changes requested):
Overall looks good, just one comment about timezones and some recommendations about packages available.

argus/backend/service/results_service.py
```diff
@@ -365,12 +368,25 @@ def _get_tables_metadata(self, test_id: UUID) -> list[ArgusGenericResultMetadata
...
+
+        if start_date:
+            raw_query += " AND sut_timestamp >= ?"
+            parameters.append(start_date)
+        if end_date:
+            raw_query += " AND sut_timestamp <= ?"
+            parameters.append(end_date)
+
+        if start_date or end_date:
+            raw_query += " ALLOW FILTERING"
```
fruch, line 387: do we have some estimation of the performance of those queries ? before/after ?

fruch:
> It's deployed to staging. In general, it looks like this: image

the layout looks good to me, 
I think we are using a bit too much coloring for button, it becomes a bit less clear what each one represents.

fruch:
> 
> the layout looks good to me, I think we are using a bit too much coloring for button, it becomes a bit less clear what each one represents.

maybe it's time to do a session with a designer, about the UI of those ?

argus/backend/service/results_service.py
soyacz: No, but in general should be faster as there will be less data sent to the backend.
Possibly this whole table could be improved by making SUT_TIMESTAMP a static field - as we don't support different SUT versions for one test run. I'd consider doing this in separate PR. WDYT?

k0machi:
> > the layout looks good to me, I think we are using a bit too much coloring for button, it becomes a bit less clear what each one represents.
> 
> maybe it's time to do a session with a designer, about the UI of those ?

I can look into the UI/UX after this PR, we've discussed some changes in #489.

argus/backend/controller/api.py
```diff
@@ -386,13 +387,21 @@ def test_info():
 @api_login_required
 def test_results():
     test_id = request.args.get("testId")
+    start_date_str = request.args.get("startDate")
+    end_date_str = request.args.get("endDate")
+
     if not test_id:
         raise Exception("No testId provided")
+
+    start_date = datetime.fromisoformat(start_date_str).astimezone(timezone.utc) if start_date_str else None
```
k0machi, line 396: What does the input string here look like? My concern is potential shift of the requested data +/- offset from UTC if input string doesn't have `Z` at the end or timezone specified.

k0machi (approved):
LGTM, other than one comment

argus/backend/controller/api.py
soyacz: input string is just the date e.g. 2024-10-30

soyacz:
> > > the layout looks good to me, I think we are using a bit too much coloring for button, it becomes a bit less clear what each one represents.
> > 
> > 
> > maybe it's time to do a session with a designer, about the UI of those ?
> 
> I can look into the UI/UX after this PR, we've discussed some changes in #489.

Yes, I tried improving UI/UX but I'd rather do it in separate PR to discuss it there out of this context.
May I merge it?

PR #493, opened 2024-10-28
fix(templates/base.html.j2): Do not parse JSON provided by the backend

k0machi:
This commit fixes an issue where some users would not be able to use
argus because parts of their usernames contained unescaped json
sequences, which would fail JSON.parse. As a replacement, templates now
receive JSON as a JS object literal.


roydahan (approved):

soyacz (approved):

fruch (approved):
LGTM

PR #495, opened 2024-10-30
improvement(graphs-ux): graphs filters reorganization

soyacz:
Some improvements around graphs filtering: less colors, more organization and labels.
image

### DIFF
--- a/frontend/TestRun/ResultsGraphs.svelte
+++ b/frontend/TestRun/ResultsGraphs.svelte
@@ -16,12 +16,17 @@
     let filteredGraphs = [];
     let startDate = "";
     let endDate = "";
+    let dateRange = 3;
+    let showCustomInputs = false;
     let width = 500;  // default width for each chart
     let height = 300;  // default height for each chart
+
+    $: setDateRange(dateRange);
+
     const dispatch = createEventDispatcher();
 
     const dispatch_run_click = (e) => {
-        dispatch("runClick", { runId: e.detail.runId });
+        dispatch("runClick", {runId: e.detail.runId});
     };
 
     const fetchTestResults = async function (testId) {
@@ -72,10 +77,17 @@
     };
 
     const setDateRange = (months) => {
+        if (months === -1) {
+            startDate = "";
+            endDate = "";
+            showCustomInputs = true;
+            return;
+        }
         const now = dayjs();
         endDate = now.format('YYYY-MM-DD');
         const pastDate = now.subtract(months, 'month');
         startDate = pastDate.format('YYYY-MM-DD');
+        showCustomInputs = false;
         fetchTestResults(test_id);
     };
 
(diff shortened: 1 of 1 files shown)


frontend/TestRun/ResultsGraphs.svelte
```diff
@@ -172,48 +173,81 @@
...
+        </button>
+        <button class="btn btn-outline-primary btn-sm {startDate === dayjs().subtract(3, 'month').format('YYYY-MM-DD') ? 'active' : ''}"
+                on:click={() => {setDateRange(3); showCustomInputs = false;}}>
+            Last 3 Months
+        </button>
+        <button class="btn btn-outline-primary btn-sm {startDate === dayjs().subtract(6, 'month').format('YYYY-MM-DD') ? 'active' : ''}"
+                on:click={() => {setDateRange(6); showCustomInputs = false;}}>
+            Last 6 Months
+        </button>
+        <button class="btn btn-outline-primary btn-sm {showCustomInputs ? 'active' : ''}"
```
k0machi, line 190: This is a good candidate to utilize Svelte reactivity pattern:

```svelte
<script>

     let dateRange = 3;
     $: setDateRange(dateRange);
</script>

<button class:active={dateRange === 3} on:click={() => dateRange == 3}>Last 3 Months</button>
```

frontend/TestRun/ResultsGraphs.svelte
```diff
@@ -172,48 +173,81 @@
...
-        <button
-                on:click={() => toggleColumnFilter(filter)}
-                class:selected={selectedColumnFilters.some(f => f === filter)}
-                style="background-color: #a3e2cc"
-        >
-            {filter}
+    </div>
+</div>
+
+<label>Tables Filters</label>
```
k0machi, line 211: `<label>` should not be used like that, use either heading `hN` (e.g. `h4`) or a normal `div`

k0machi (changes requested):

frontend/TestRun/ResultsGraphs.svelte
```diff
@@ -172,48 +173,81 @@
     });
 </script>
 <div class="filters-container">
-    <div class="input-group input-group-sm mx-2">
-        <input type="date" class="form-control date-input" bind:value={startDate} on:change={() => fetchTestResults(test_id)} />
-        <input type="date" class="form-control date-input" bind:value={endDate} on:change={() => fetchTestResults(test_id)} />
-        <button class="btn btn-info btn-sm" on:click={() => setDateRange(1)}>Last Month</button>
-        <button class="btn btn-info btn-sm" on:click={() => setDateRange(3)}>Last 3 Months</button>
-        <button class="btn btn-info btn-sm" on:click={() => setDateRange(6)}>Last 6 Months</button>
+    <label class="mt-1">Time range:</label>
```
k0machi, line 176: Ditto

fruch:
I think the table/column filter separation/titles is a bit unclear for the common user. what we mean by column here is, and what we mean by table, is not exactly intuitive 

Maybe "group filters" and "metric filter" ? Or something in that direction?

Also it would be really helpful to have tooltips with a longer description for each button we show there

frontend/TestRun/ResultsGraphs.svelte
soyacz: I replaced it with <span>

frontend/TestRun/ResultsGraphs.svelte
soyacz: ditto

frontend/TestRun/ResultsGraphs.svelte
soyacz: thanks, improved

soyacz:
image

k0machi (approved):

fruch:
> > I think the table/column filter separation/titles is a bit unclear for the common user. what we mean by column here is, and what we mean by table, is not exactly intuitive
> > Maybe "group filters" and "metric filter" ? Or something in that direction?
> 
> I replaced with just `Filters` and `Metrics` - after clicking it should become more intuitive.

Much better

> 
> > Also it would be really helpful to have tooltips with a longer description for each button we show there
> 
> I validated idea and it doesn't play well with filters. E.g. descriptions: for `read - _mgmt_repair_cli - stalls - REACTOR_STALLED` is `REACTOR_STALLED event counts` or table `read - _mgmt_repair_cli - latencies` has `read workload - Scylla-Manger repair`. I'd rather see descriptions in graphs itself, under the titles.

Logicly I would argue that it's best to be visible when you are selecting what to look at.
and not after browsing all of the graphs.
nevertheless, we should be able to see those descriptions for anything that has a description.

fruch (approved):
LGTM

soyacz:
> Logicly I would argue that it's best to be visible when you are selecting what to look at. and not after browsing all of the graphs. nevertheless, we should be able to see those descriptions for anything that has a description.

https://github.com/scylladb/argus/issues/496

Issue #490, opened 2024-10-24
Automatically annotate performance graphs with dependency changes


soyacz:
### Problem:
Changes in important dependencies such as driver versions, load tool versions, kernel versions, or Scylla Manager versions can introduce performance regressions. These changes are often difficult to detect during the manual performance review process, as there is no immediate visual indication correlating dependency changes with observed performance metrics.

### Solution:
Automatically annotate performance graphs when a dependency change is detected to help correlate changes in performance with system alterations. This will be achieved as follows:
1. **Automatic Detection:** During data retrieval from the database, the system will pull information about dependencies (e.g., driver version, kernel version) associated with each data point. When a change in any dependency is detected, an annotation will be generated.
   
2. **Visual Annotation:** A horizontal line will be placed on the graph at the point where the dependency change occurred. The line will represent the time of the change and will not extend across the graph. A new line will be placed for each detected change.

3. **Tooltip Information:** When a user hovers over the annotation, a tooltip will display detailed information about the change, including:
   - The name of the package (e.g., driver, kernel, Scylla Manager).
   - The specific version that the package was updated to.

PR #502, opened 2024-11-13
improvement(graphs): Mark dependency changes

soyacz:
In performance tests it's important to note changes of dependencies,
like kernel, drivers, stress tools and SUT version.

This commit changes the way point is displayed:
when there's package change (except SUT change) then the point gets
a white background.
Hovering over the point shows tooltip with all packages changes
(including SUT).

Tooltip position was adjusted to be show above the point so mouse is not
covering text.

closes: https://github.com/scylladb/argus/issues/490

image

### DIFF
--- a/argus/backend/service/results_service.py
+++ b/argus/backend/service/results_service.py
@@ -127,13 +127,37 @@ colors = [
 shapes = ["circle", "triangle", "rect", "star", "dash", "crossRot", "line"]
 
 
-def get_sorted_data_for_column_and_row(data: List[ArgusGenericResultData], column: str, row: str) -> List[Dict[str, Any]]:
-    return sorted([{"x": entry.sut_timestamp.strftime('%Y-%m-%dT%H:%M:%SZ'),
+def get_sorted_data_for_column_and_row(data: List[ArgusGenericResultData], column: str, row: str,
+                                       runs_details: RunsDetails, main_package: str) -> List[Dict[str, Any]]:
+    points =  sorted([{"x": entry.sut_timestamp.strftime('%Y-%m-%dT%H:%M:%SZ'),
                     "y": entry.value,
-                    "id": entry.run_id}
+                    "id": entry.run_id,
+                    }
                    for entry in data if entry.column == column and entry.row == row],
                   key=lambda point: point["x"])
-
+    if not points:
+        return points
+    packages = runs_details.packages
+    prev_versions = {pkg.name: pkg.version + (f" ({pkg.date})" if pkg.date else "") for pkg in packages.get(points[0]["id"], [])}
+    points[0]['changes'] = [f"{main_package}: {prev_versions.pop(main_package, None)}"]
+    points[0]['dep_change'] = False
+    for point in points[1:]:
+        changes = []
+        mark_dependency_change = False
+        current_versions = {pkg.name: pkg.version + (f" ({pkg.date})" if pkg.date else "") for pkg in packages.get(point["id"], [])}
+        main_package_version = current_versions.pop(main_package, None)
+        for pkg_name in current_versions.keys() | prev_versions.keys():
+            curr_ver = current_versions.get(pkg_name)
+            prev_ver = prev_versions.get(pkg_name)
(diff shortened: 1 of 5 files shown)


fruch:
@soyacz 

can we add the the full scylla release as well ?

soyacz:
> @soyacz
> 
> can we add the the full scylla release as well ?

We can. See also that kernel changes almost every week, I wonder if we shouldn't mark only big changes like major version updates only (I mean still change will be shown, but point won't be marked).

fruch:
> > @soyacz
> > can we add the the full scylla release as well ?
> 
> We can. See also that kernel changes almost every week, I wonder if we shouldn't mark only big changes like major version updates only (I mean still change will be shown, but point won't be marked).

before I want to notice the change, we want to see them.
then we can indicate there was a change on the inside the popup
and last we can decide what to indicate on the graph

one more thing, not for this PR, we should be able to change the x axis to something else like driver version, or just plain data of the run, cause we don't have version change or date in the drivers CI (at least not yet, we didn't thought of it yet)

soyacz:
> before I want to notice the change, we want to see them. then we can indicate there was a change on the inside the popup and last we can decide what to indicate on the graph
> 

Summarizing:
Ok, any change in dependencies (except SUT) will be marked with different point color. 
All the changes, including SUT will be shown in tooltip.
If this won't fit us well in the wild, we can adjust.

> one more thing, not for this PR, we should be able to change the x axis to something else like driver version, or just plain data of the run, cause we don't have version change or date in the drivers CI (at least not yet, we didn't thought of it yet)

Generally, SUT is detected automatically based on which dependency changes the most often. If driver version will be changing the most often then it becomes the SUT. This feature is used e.g. in SM tests. We can adjust it in followup issues if we need something else.

fruch:
> > before I want to notice the change, we want to see them. then we can indicate there was a change on the inside the popup and last we can decide what to indicate on the graph
> 
> Summarizing: Ok, any change in dependencies (except SUT) will be marked with different point color. All the changes, including SUT will be shown in tooltip. If this won't fit us well in the wild, we can adjust.
> 
> > one more thing, not for this PR, we should be able to change the x axis to something else like driver version, or just plain data of the run, cause we don't have version change or date in the drivers CI (at least not yet, we didn't thought of it yet)
> 
> Generally, SUT is detected automatically based on which dependency changes the most often. If driver version/build date will be changing the most often then it becomes the SUT. This feature is used e.g. in SM tests. We can adjust it in followup issues if we need something else.

not sure automatic would fit the best, consider a case we run CI with multiple release of scylla, let say OSS and enterprise ?

soyacz:
> run CI with multiple release of scylla, let say OSS and enterprise

package name for both is `scylla-server` so it works (just both anyway are separated for different releases on the graph)

fruch:
> > run CI with multiple release of scylla, let say OSS and enterprise
> 
> package name for both is `scylla-server` so it works (just both anyway are separated for different releases on the graph)

in driver CI, we care to see the driver version change / commit, not necessarily scylla, in my example both driver, and scylla might be changing across time

soyacz:
> > > run CI with multiple release of scylla, let say OSS and enterprise
> > 
> > 
> > package name for both is `scylla-server` so it works (just both anyway are separated for different releases on the graph)
> 
> in driver CI, we care to see the driver version change / commit, not necessarily scylla, in my example both driver, and scylla might be changing across time

I doubt in testing performance while changing both dependencies at the same pace, anyway, we can adjust to whatever we need later.

soyacz:
@fruch I adjusted the code, please see PR description to see the changes.

soyacz:
@fruch I adjusted things we agreed yesterday: SUT version only, error threshold instead 'limit'. Also I did various fixes for tooltip: enlarge font, make it be above mouse cursor (so it does not cover text).

soyacz:
@k0machi this one is unrelated to highlights (discussion) view widget and is relatively small. I propose to review it first so it can be merged and used.

fruch (approved):
LGTM

k0machi (approved):

PR #503, opened 2024-11-14
fix(ViewTypes.js): Add a fallback widget in case of undefined widgets

k0machi:
This fix addresses a potential issue where a widget created on a newer
version of argus will cause an error on an older one when it is loaded
from the database.

### DIFF
--- a/frontend/AdminPanel/ViewsManager.svelte
+++ b/frontend/AdminPanel/ViewsManager.svelte
@@ -4,13 +4,14 @@
     import Fa from "svelte-fa";
     import { faPlus, faTrash } from "@fortawesome/free-solid-svg-icons";
     import ViewWidget from "./ViewWidget.svelte";
-    import { ADD_ALL_ID, Widget } from "../Common/ViewTypes";
+    import { ADD_ALL_ID, Widget, WIDGET_TYPES } from "../Common/ViewTypes";
     import queryString from "query-string";
     import * as urlSlug from "url-slug";
     import Select from "svelte-select";
     import ViewSelectItem from "./ViewSelectItem.svelte";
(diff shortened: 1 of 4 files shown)


soyacz (changes requested):
need also to add support in View edit

k0machi:
> need also to add support in View edit

I think it would be for the best if we restrict view editing if an unsupported widget is found in the view definition. Show the error to the user and stop the edit process.

k0machi:
Demo:

image

soyacz:
> > need also to add support in View edit
> 
> I think it would be for the best if we restrict view editing if an unsupported widget is found in the view definition. Show the error to the user and stop the edit process.

Cant we make it to be ignored? so we still can edit the others or remove unsupported one?

k0machi:
> > > need also to add support in View edit
> > 
> > 
> > I think it would be for the best if we restrict view editing if an unsupported widget is found in the view definition. Show the error to the user and stop the edit process.
> 
> Cant we make it to be ignored? so we still can edit the others or remove unsupported one?

I can add an "Edit anyway" button that will drop the unsupported widget

soyacz:
> > > > need also to add support in View edit
> > > 
> > > 
> > > I think it would be for the best if we restrict view editing if an unsupported widget is found in the view definition. Show the error to the user and stop the edit process.
> > 
> > 
> > Cant we make it to be ignored? so we still can edit the others or remove unsupported one?
> 
> I can add an "Edit anyway" button that will drop the unsupported widget

Yes, if can't just "ignore" it then better to have possibility to fix view in case it's needed.

k0machi:
@soyacz added the "Edit anyway" button.

image

soyacz (approved):
LGTM

PR #504, opened 2024-11-14
fix(stats.py): Add a fallback for old assignments and stats

k0machi:
This commit fixes an issue where old schedules would not be counted for
tests that have not been run (but have been scheduled). This restores
both the assignee icon and NOT RUN status.


roydahan (approved):

soyacz (approved):
LGTM

PR #519, opened 2024-11-20
feature(frontend): Add possibility to clone empty jobs

### LINKED ISSUE #379: Ability to clone jobs that doesn't have runs
For testing purposes people should be able to clone job into staging, from job that doesn't have any runs.

for example from the individual nemesis folder, where we don't keep running all of them, but still we want people to be able to clone those and give it the first run.

(as a worked around, I have to run on master until I got one run that start, and then I could clone, but that's waste of time and resources)

k0machi:
This commit adds a new button alongside "Configure" button on the job
collapse, allowing user to clone a job without having to run one first.
In this case, the parameter retrieval is skipped initially, and the
parameters are retrieved from the new job. In case of pressing this
button on a job with runs, the flow is largely the same as normal clone,
the parameters are taken from the latest job. Additionally, for
usability, last stage now gives user possibility to instantly add newly
created job to the workspace.

Fixes #379

### DIFF
--- a/frontend/TestRun/Jenkins/BuildSuccessPlaceholder.svelte
+++ b/frontend/TestRun/Jenkins/BuildSuccessPlaceholder.svelte
@@ -93,8 +93,16 @@
                     <div>
                         This is a first build, which will require a re-run to actually start.
                         <div class="my-2 p-2">
-                            <button class="w-100 btn btn-success" on:click={() => dispatch("exit", { firstBuildRestart: true })}>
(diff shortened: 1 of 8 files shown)


k0machi:
Button placement:

image

The flow is exactly the same as normal clone, so refer to #349 for a demo.

fruch:
can we also trigger them ?

k0machi:
> can we also trigger them ?

I can add that in as well, it would use the same logic to retrieve latest build (or start from zero)

Issue #505, opened 2024-11-15
Add possibility to 'enlarge' graph


soyacz:
When analyzing graphs it would be convenient to enlarge them for better view.
This also will improve experience in online meetings to draw attention to specific graph.

PR #522, opened 2024-11-26
improvement(graphs): enlarge graph button

soyacz:
Show enlarge graph button in right upper corner of each graph to open modal with it on full screen.

closes: https://github.com/scylladb/argus/issues/505

clicking top right corner button will enlarge to full screen in modal (with x to close or pressing esc on keyboard) 
image


fruch (approved):
LGTM

it would be nice seeing it in action (so it can be shared)

k0machi (approved):

Issue #486, opened 2024-10-18
Feature Request: Performance Summary View Widget


soyacz:
### Problem:
In Argus View, identifying failed test results from the current week is challenging due to the large number of tests and runs. Users must also navigate into test run details to view failed result tables, making it difficult to grasp failures at a glance. Additionally, important tested versions can be buried in the results.

### Proposed Solution:
A widget will be developed to simplify the view of weekly test results, emphasizing failures and presenting failure details upfront, without requiring the user to open test run details. It will also provide better visibility into both passed and failed tests.

### Widget Features:

1. **One widget per release**:
   - The widget will configured to summarize only one release.
   - Summary will be provided per Scylla version in given release (e.g. dev, 2024.2, 6.2 etc.)
   - User will be able to go back to previous versions (select from list containing exact version and build date or with arrow buttons to go back to previous/next ones). Ordered by Scylla build date.

2. **Highlighting missed tests**:
   - Tests that were not executed for given version or failed to start (e.g., due to infrastructure issues like AWS capacity) will be highlighted at the top of the widget (when there's no metric results).
   - These tests will be treated as failures.

3. **Test Run Lists**:
   - Widget will contain test runs segregated into two collapsible lists:
     - **Failed List**: Displays all failed test runs and their failed result tables (expanded by default). 
     - **Passed List**: Displays all successfully passed test runs (collapsed by default).
   - tests may use different test methods - each test method results should be shown in own column (3 maximum) 
   - Each test run in both lists will be clickable, allowing users to navigate to the detailed test run page if needed.
   - Each test run entry will show:
     - values of selected metrics (`key metrics`) in tabular form, with columns: result table name, and rest will be selected metrics names
     - The **count of error events** associated with that test run.
     - **Failed result tables** will be displayed (collapsed by default for passed tests, visible for failed tests).
of the error count and the failed result tables (collapsed by default).

4. **Run Investigation Status**:
   - Any test runs marked as "IGNORED" in their investigation status will be excluded from the analysis to avoid clutter.
   - Test runs that have failed result tables or error events but are manually marked as passed will still be listed in the passed list, with an indication 

5. **Widget Configuration**:
   - Users can configure the widget by:
     - Setting a name (e.g. summary - vnodes based tests, summary - tablets based tests)
     - Setting release to analyze (e.g. 2024.2, 6.2)
     - Selecting which tests to monitor from the available test groups.
     - selected metric values (e.g. p99's, duration of repair etc.)
[...]

PR #530, opened 2024-12-03
feature(summary-widget): summary widget

soyacz:
View widged showing tests statuses and results per Scylla version. Each test can be expanded to see results (tables) with key metrics. Initial features:
1. Results summary by scylla version (date, scm revision)
2. passed tests are collapsed by default
3. not run tests marked as warning
4. hiding tables with all cells green ... many more but small ones.

caveats:
1. works only with scylla-server package
2. tests filtering based on release (oss vs enterprise) implementation is poor as will work only with scylla-master and scylla-enterprise argus releases
3. key metrics are hardcoded, followup issue: https://github.com/scylladb/qa-tasks/issues/1821

Did a refactor of ResultsTab - extracted ResultTable to be able to reuse it across different parts of Argus (like Summary Widget)

refs: https://github.com/scylladb/argus/issues/486

### DIFF
--- a/argus/backend/controller/view_api.py
+++ b/argus/backend/controller/view_api.py
@@ -6,6 +6,7 @@ from flask import (
     request,
 )
 from argus.backend.controller.views_widgets.highlights import bp as highlights_bp
+from argus.backend.controller.views_widgets.summary import bp as summary_bp
 from argus.backend.error_handlers import handle_api_exception
 from argus.backend.models.web import User
 from argus.backend.service.stats import ViewStatsCollector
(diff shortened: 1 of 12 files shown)


soyacz:
preview on staging - see Performance Regression Tests View

fruch:
filter button, are not collapsing with the job
image

fruch:
I think the ability to copy the version string (including git sha), now it's displayed only on the dropdown, and it's not easy to copy anything from it.

soyacz:
I think this one could be reviewed, @k0machi please check when you have the time

k0machi (approved):
LGTM

PR #536, opened 2024-12-16
feature(service/planner_service): Trigger jenkins jobs by plans

### LINKED ISSUE #377: Auto triggering tests bases on planning
ability to trigger multiple jobs base on the version specified in their planning.

for example if I have multiple jobs marked as planned for RC1, we should have a place to trigger all of those jobs with `scylla_version` as example: `2024.1.rc1`

```[tasklist]
### Tasks
- [x] Refactoring - new Planner in place
- [x] SCT - Add job metadata
- [ ] Triggering by API - https://github.com/scylladb/argus/pull/536
- [ ] start calling the api from scylla-pkg
```

k0machi:
This commit adds a new feature to the Release Planning API - allowing
API consumers to trigger jobs from within argus in Jenkins. The endpoint
accepts target version (for example "6.2.0") in which case it will
trigger all plans with this version (release independent), release name
(will trigger all plans within that relase) or specific plan id (will
only trigger that plan). The API client provides the parameters in two
parts: as a set of common parameters intended for each job (useful for
things like provision_type) and an additional per job
type/backend/region set that each job will require in order to trigger.
Currently, only longevity is supported.

Fixes #377

### DIFF
--- a/argus/backend/service/planner_service.py
+++ b/argus/backend/service/planner_service.py
@@ -6,14 +6,17 @@ from collections import defaultdict
 from copy import deepcopy
 from dataclasses import dataclass
 from functools import reduce
-from typing import Any, Optional
+from typing import Any, Optional, TypedDict
 from uuid import UUID
+from flask import g
 from slugify import slugify
 
 from argus.backend.models.plan import ArgusReleasePlan
 from argus.backend.models.web import ArgusGroup, ArgusRelease, ArgusTest, ArgusUserView, User
+from argus.backend.service.jenkins_service import JenkinsService
 from argus.backend.service.test_lookup import TestLookup
 from argus.backend.service.views import UserViewService
+from argus.backend.util.common import chunk
 
 
 LOGGER = logging.getLogger(__name__)
@@ -59,6 +62,13 @@ class CopyPlanPayload:
     replacements: dict[str, str]
     targetReleaseId: str
     targetReleaseName: str
+
+class PlanTriggerPayload(TypedDict):
+    plan_id: str | None
+    release: str | None
+    version: str | None
+    common_params: dict[str, str]
+    params: list[dict[str, str]]
 
 
 class PlannerServiceException(Exception):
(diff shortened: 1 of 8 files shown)


fruch (changes requested):
Let's add some docs and example for this API

argus/backend/service/planner_service.py
```diff
@@ -438,3 +448,88 @@ def resolve_plan(self, plan_id: str | UUID) -> list[dict[str, Any]]:
...
+        match condition_set:
+            case (True, False, False):
+                release = ArgusRelease.get(name=release_name)
+                filter_expr = { "release_id__eq": release.id }
+            case (False, True, False):
+                filter_expr = { "id__eq": plan_id }
+            case (False, False, True):
+                filter_expr = { "target_version__eq": version }
+            case _:
+                raise PlannerServiceException("No version, release name or plan id specified.", payload)
```
soyacz, line 469: this will be raised also when multiple are selected (like version and release). Is it intended? maybe it should be validated within `PlanTriggerPayload`?

argus/backend/service/planner_service.py
```diff
@@ -438,3 +448,88 @@ def resolve_plan(self, plan_id: str | UUID) -> list[dict[str, Any]]:
...
+        params = payload.get("params", [])
+        test_ids = [test_id for plan in plans for test_id in plan.tests]
+        group_ids = [group_id for plan in plans for group_id in plan.groups]
+
+        tests = []
+        for batch in chunk(test_ids):
+            tests.extend(ArgusTest.filter(id__in=batch).all())
+
+        for batch in (chunk(group_ids)):
+            tests.extend(ArgusTest.filter(group_id__in=batch).allow_filtering().all())
```
soyacz, line 487: won't it contain duplicates?

argus/backend/service/planner_service.py
```diff
@@ -438,3 +448,88 @@ def resolve_plan(self, plan_id: str | UUID) -> list[dict[str, Any]]:
...
+
+        service = JenkinsService()
+        failures = []
+        successes = []
+        for test in tests:
+            try:
+                latest_build_number = service.latest_build(test.build_system_id)
+                if latest_build_number == -1:
+                    failures.append(test.build_system_id)
+                    continue
```
soyacz, line 499: In case we have fresh release, won't jobs be without first runs?
how about triggering it in that case, wait and try again?

argus/client/generic/cli.py
```diff
@@ -39,8 +41,23 @@ def finish_run(api_key: str, base_url: str, id: str, status: str, scylla_version
     client.finalize_generic_run(run_id=id, status=status, scylla_version=scylla_version)
 
 
+@click.command("trigger-jobs")
+@click.option("--api-key", help="Argus API key for authorization", required=True)
+@click.option("--base-url", default="https://argus.scylladb.com", help="Base URL for argus instance")
+@click.option("--job-info-file", required=True, help="JSON file with trigger information (see detailed docs)")
```
soyacz, line 47: looks like detailed docs are missing

argus/client/generic/cli.py
```diff
@@ -39,8 +41,23 @@ def finish_run(api_key: str, base_url: str, id: str, status: str, scylla_version
...
+def trigger_jobs(api_key: str, base_url: str, job_info_file: str):
```
soyacz, line 48: can we make job-info-file consist only jenkins params and things like version, plan_id, release be specified in cli?
This way maybe we could commit job-info-files and reuse across releases?

argus/backend/service/planner_service.py
k0machi: You are correct, after discussion we decided to also support (version, release) pair, but yes, previously that would be an exception and it was intended.

argus/backend/service/planner_service.py
k0machi: Good catch, will fix

argus/backend/service/planner_service.py
k0machi: So in theory SCT starts the first build when it syncs the release tree, I could also try firing off a job without parameters to just get it past `#1` and executed on next call.

argus/client/generic/cli.py
k0machi: Yes, adding them soon

argus/client/generic/cli.py
k0machi: Hmm, good idea, will do it that way.

argus/backend/service/planner_service.py
k0machi: Done

argus/backend/service/planner_service.py
k0machi: Done

argus/client/generic/cli.py
k0machi: Done

argus/client/generic/cli.py
k0machi: Added docs

k0machi:
Added documentation and examples to `api_usage.md`

docs/api_usage.md
```diff
@@ -44,3 +44,77 @@ Accepts following parameters:
...
+POST /api/v1/planning/plan/trigger
+```
+
+Accepts following payload:
+
+Type: application/json
+
+| Parameter | Type | Description |
+| --------- | ---- | ------------|
+| release_name          | string     | Release name to trigger plans in (can be mixed vith version to narrow filter)            |
```
soyacz, line 58: `release` instead of `release_name`

docs/api_usage.md
```diff
@@ -44,3 +44,77 @@ Accepts following parameters:
...
+    ]
+  }, 
+  "status": "ok"
+}
+```
+
+Additionally, this endpoint is available inside `argus-client-generic` executable, as follows:
+
+```bash
+argus-client-generic trigger-jobs --api-key $key --version $version --plan_id $id --release_name $release_name --job-info-file $file_path
```
soyacz, line 117: `--release-name` is in CLI, shouldn't we switch to `release` like in api calls?

soyacz (commented):
small remarks, otherwise LGTM

docs/api_usage.md
k0machi: Fixed

docs/api_usage.md
k0machi: Fixed

soyacz (approved):
LGTM

PR #537, opened 2024-12-16
improvement(TestRunInfo.svelte): Add popup hints for copy buttons

### LINKED ISSUE #431: Confusing buttons text: "Hydra Show Monitor". The button seems to have no effect
Argus users expect the buttons of "Hydra Show Monitor" and "Hydra Show Logs" to do something visible.
A user has no way to understand, that unlike other buttons, this only copy a command to clipboard.
image

Two possible solutions could be:

1. Add a tooltip to tell user about it.
2. Rename button text like: "Hydra Show Monitor"  => "Copy Hydra-Show-Monitor command-line"

k0machi:
This commit adds success indicators when user does copy-to-clipboard
actions, making it less confusing.

Fixes #431

### DIFF
--- a/frontend/TestRun/ResourcesInfo.svelte
+++ b/frontend/TestRun/ResourcesInfo.svelte
@@ -9,6 +9,7 @@
         faCopy,
         faPlay,
     } from "@fortawesome/free-solid-svg-icons";
+    import { sendMessage } from "../Stores/AlertStore";
     let sortHeaders = {
         creationTime: ["instance_info", "creation_time"],
         terminationTime: ["instance_info", "termination_time"],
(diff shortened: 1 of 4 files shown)


soyacz (approved):
LGTM

frontend/TestRun/ResourcesInfo.svelte
```diff
@@ -88,6 +89,7 @@
                 navigator.clipboard.writeText(
                     `${regions} hydra clean-resources --backend ${backend} --test-id ${run_id}`
                 );
+                sendMessage("success", `\`${regions} hydra clean-resources --backend ${backend} --test-id ${run_id}\` has been copied to your clipboard`);
```
fruch, line 92: can we put the text into variable, and not create it twice ? it's a recipe of getting it out of sync, and would be a bit confusing.

fruch (approved):
LGTM

PR #540, opened 2024-12-16
fix(plugins/sct): Support SCT runner for resources table

k0machi:
This commit adds logic to add SCT runner to the resource table and also
adds client support for its name. Additionally, it adds a workaround for
older client versions, where SCT runner will be submitted without a
name.

Fixes scylladb/scylla-cluster-tests#9543

### DIFF
--- a/argus/backend/plugins/sct/controller.py
+++ b/argus/backend/plugins/sct/controller.py
@@ -40,7 +40,8 @@ def sct_set_runner(run_id: str):
         public_ip=payload["public_ip"],
         private_ip=payload["private_ip"],
         region=payload["region"],
-        backend=payload["backend"]
+        backend=payload["backend"],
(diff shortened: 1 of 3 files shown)


argus/backend/plugins/sct/controller.py
```diff
@@ -40,7 +40,8 @@ def sct_set_runner(run_id: str):
         public_ip=payload["public_ip"],
         private_ip=payload["private_ip"],
         region=payload["region"],
-        backend=payload["backend"]
+        backend=payload["backend"],
+        name=payload.get("name")
```
soyacz, line 44: maybe here `name=payload.get("name") or 'sct-runner'`?

soyacz (approved):
small comment (not sure if valid). 
LGTM

PR #541, opened 2024-12-16
fix(sct/testrun): Check for duplicate entries on log submit

### LINKED ISSUE #532: Adding logs manually may duplicate records without possibility to delete redundant ones
CI job: [scylla-staging/valerii/vp-longevity-aws-custom-d2-workload1-multidc-big#15](https://argus.scylladb.com/tests/scylla-cluster-tests/266b471d-200b-4375-8aa5-812a3f8e0acb)
Originally the CI job failed to collect logs.
I triggered appropriate command manually.
Started adding links myself one by one.
After pressing "submit" button for the fifth log link all the log link records got duplicated:

Image

So, at first, it must be smart enough to filter out duplicated.
And, secondly, we should be able to remove, at least, manually added log links any time.

k0machi:
Fixes #532

### DIFF
--- a/argus/backend/plugins/sct/testrun.py
+++ b/argus/backend/plugins/sct/testrun.py
@@ -225,6 +225,8 @@ class SCTTestRun(PluginModelBase):
 
     def submit_logs(self, logs: list[dict]):
         for log in logs:
+            if any(existing[0] == log["log_name"] for existing in self.logs):
+                continue
             self.logs.append((log["log_name"], log["log_link"]))
 
     def add_screenshot(self, screenshot_link: str):


argus/backend/plugins/sct/testrun.py
```diff
@@ -225,6 +225,11 @@ def finish_run(self, payload: dict = None):
 
     def submit_logs(self, logs: list[dict]):
         for log in logs:
+            try:
+                next(existing for existing in self.logs if existing[0] == log["log_name"])
+                raise Exception(f"Log {log['log_name']} already exists!", log["log_name"])
+            except StopIteration:
+                pass
```
soyacz, line 232: how about:
```
def submit_logs(self, logs: list[dict]):
    for log in logs:
        if any(existing[0] == log["log_name"] for existing in self.logs):
            continue
        self.logs.append((log["log_name"], log["log_link"]))
```
simpler and it should not break when there's one duplicate across multiple files - better to skip them

k0machi: Done

soyacz (approved):
LGTM

Issue #500, opened 2024-11-12
Notify assignee from Action Item in Hightlights Widget


soyacz:
In https://github.com/scylladb/argus/pull/499 we didn't include notifications after setting assignee to Action Item.
This is followup issue for adding support for that.

PR #558, opened 2024-12-20
improvement(highlights): send notification to action item assignee

soyacz:
When assigning action item in Highlights Widget,
assignee will get Argus Notification with
link to the view and action details.

closes: https://github.com/scylladb/argus/issues/500


soyacz:
example notification:
image

fruch (approved):
LGTM

Issue #550, opened 2024-12-19
Automatic SUT selection for graphs is not working properly


soyacz:
There are cases (e.g. https://argus.scylladb.com/test_runs?state=WyI5YzUzYTc5NS0zODFiLTQ4ZWQtOTYwYS02ZDE4NGRhY2FiYTciXQ) where SUT is not chosen properly (selects kernel instead of Scylla).
Add ability to manualy select SUT package name (and save it for the future)

soyacz:
@fruch I implemented quick workaround so mentioned graphs work (without breaking SM graphs).
In future will be easy to extend it for driver tests in case it was needed.
In the meantime, after implementing https://github.com/scylladb/argus/issues/551 we should have full solution. 
Can we close this one in favor of https://github.com/scylladb/argus/issues/551?

soyacz:
btw. it's on staging for testing

PR #559, opened 2024-12-20
fix(graphs): automatic SUT issue workaround

soyacz:
Before implementing proper fix with ability to SUT package for graphs, introducing quick issue workaround to enable work with some tests.

refs: https://github.com/scylladb/argus/issues/550


k0machi (approved):

fruch (approved):
LGTM

Issue #510, opened 2024-11-18
Markdown support for Highlights widget


soyacz:
We want to support markdown for highlights widget to be able to:
1. mention people
2. link issues
3. Style content, multiline entries
4. show pictures

PR #560, opened 2024-12-23
improvement(highlights): add markdown support

soyacz:
Added support for highlights widget.
Reused 'comments' editor known from job discussions. Thus implementing all that is supported by this component: titles, links, multi-line content, code blocks etc.

One caveat: mentioning does not trigger notifications yet. This is going to be done in followup task.

closes: https://github.com/scylladb/argus/issues/510

image


fruch (approved):
LGTM

Issue #554, opened 2024-12-20
Graphs Views for fast graphs filtering


soyacz:
### Problem Statement

The current Argus graphing functionality lacks a structured way to manage and focus on specific sets of graphs. Navigating through numerous graphs in the "Graphs window," where all graphs are presented, can be overwhelming. This makes it difficult for users to concentrate on relevant data without applying filters repeatedly. This inefficiency affects usability and slows down workflows, especially for recurring tasks requiring specific graph collections.

### Proposed Solution

Introduce a "Graph View" feature that allows users to group and persist specific graphs for reuse. This feature will enable users to create, manage, and activate Graph Views seamlessly while ensuring the selected graphs are readily accessible in subsequent sessions. The functionality will be scoped to the "Graphs window" where all graphs are displayed.

#### Key Features

1. **Create and Name Graph Views**:
   - Users can create a new Graph View using an "Add View" button.
   - Clicking "Add View" opens a modal where the user assigns a name to the Graph View.
   - The Graph View is immediately saved upon creation.

2. **Assign Graphs to a Graph View**:
   - Each graph in the "Graphs window" will have an action button in its top-left corner. Clicking this button allows the user to assign the graph to any previously created Graph View.
   - Graphs are added to the selected Graph View instantly without requiring additional saving.

3. **Activate Graph Views**:
   - Graph Views are represented by buttons labeled with their names.
   - Clicking a Graph View button displays only the graphs assigned to that view within the "Graphs window."
   - Graphs filtered by other means (e.g., search filters) will appear below the Graph View graphs, with a visible separator.

4. **Remove Graphs from a Graph View**:
   - Graphs in a Graph View will include a "Remove" button to unassign them from the current view.

5. **Reorder Graphs in a Graph View**:
   - Users can reorder graphs within a Graph View to prioritize their display order. This reordering is specific to the active Graph View and does not affect other contexts.

6. **Persistent Storage**:
   - All modifications to Graph Views (e.g., adding or removing graphs, reordering) will persist automatically, ensuring the Graph Views are accessible across sessions.

7. **Integration with Argus Views**:
   - Graph Views can be manually linked to Argus Views during the creation of a Graph Widget. Users will select which Graph Views to include in the widget.

8. **Shareable Links**:
   - Each Graph View will have a unique URL that opens with the specific Graph View active, simplifying collaboration and reuse.

9. **Future Extensibility**:
[...]

PR #566, opened 2025-01-03
feature(graphs-views): show predefined list of graphs

soyacz:
Users are allowed to create 'Graph Views' which can store specific
graphs. Created Graph Views are visible in test's graph window and can
be selected for quick look on saved graphs list. This way saving time
for playing with filters.

Graph Views may have a name and description.

closes: https://github.com/scylladb/argus/issues/554

Main View with all the graphs:
image

When View is activated:
image

Adding new view:
image

Graphs may be added/removed by left top corner button (with tooltip).
Filters are created for each Graph View and work within it.

### DIFF
--- a/frontend/TestRun/Components/Filters.svelte
+++ b/frontend/TestRun/Components/Filters.svelte
@@ -1,87 +1,87 @@
 <script lang="ts">
-    import {onMount} from 'svelte'
+    import {onMount} from 'svelte';
 
-    export let graphs: any[] = []
-    export let filteredGraphs: any[] = []
-    let tableFilters: { level: number; items: string[] }[] = []
-    let columnFilters: string[] = []
-    let selectedTableFilters: { name: string; level: number }[] = []
-    let selectedColumnFilters: string[] = []
+    export let graphs: any[] = [];
+    export let filteredGraphs: any[] = [];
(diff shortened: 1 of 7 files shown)


frontend/TestRun/Components/Filters.svelte
```diff
@@ -0,0 +1,144 @@
...
+
+    function getTableFilterColor(level: number) {
+        const colors = ["#B8EFFF", "#FECBA1", "#D6B3E6", "#FFE699", "#F0A5C5", "#C4C8CA"]
+        return colors[(level - 1) % colors.length]
+    }
+
+    onMount(() => {
+        extractTableFilters()
+        extractColumnFilters()
+        filterTables()
```
k0machi, line 83: Missing semicolons (in most of the script block). While possible in JS due to legacy reasons, this should not be done as ASI might insert semicolon in places you might not expect them to be.

k0machi (changes requested):

frontend/TestRun/Components/Filters.svelte
soyacz: fixed

k0machi (approved):

fruch (approved):
LGTM

Issue #507, opened 2024-11-18
Support Notifications when mentioning user in Highlights widget


soyacz:
Followup for https://github.com/scylladb/argus/issues/510

PR #568, opened 2025-01-03
improvement(highlights): support mentions in highlights widget

soyacz:
Highlights (with action items) now support mentions in content.

closes: https://github.com/scylladb/argus/issues/507


k0machi (approved):
LGTM, we should also update the normal mention template to the similar template you've added for view mention.

PR #569, opened 2025-01-08
improvement(graphs): add show graph button for each column in results

soyacz:
When viewing results tables users now can view graph for selected column to see values from the past/future.

closes: https://github.com/scylladb/qa-tasks/issues/1822

image
after clicking one:
image

clicking on the point opens new browser window (popup, without navigation bar) with selected test run.
Works also in summary view widget.

### DIFF
--- a/frontend/Common/ModalWindow.svelte
+++ b/frontend/Common/ModalWindow.svelte
@@ -1,12 +1,19 @@
 <script>
     import { createEventDispatcher } from "svelte";
+    import { onMount, onDestroy } from "svelte";
 
     const dispatch = createEventDispatcher();
 
     export let widthClass = "h-50";
 
+    function handleKeyDown(event) {
+        if (event.key === "Escape") {
+            dispatch("modalClose");
+        }
+    }
 </script>
 
+<svelte:window on:keydown={handleKeyDown} />
 <div class="modal-window">
     <div class="d-flex align-items-center justify-content-center p-4">
         <div class="rounded bg-white p-4 {widthClass}">
--- a/frontend/TestRun/Components/ResultTable.svelte
+++ b/frontend/TestRun/Components/ResultTable.svelte
@@ -1,22 +1,41 @@
 <script>
-    import { faMarkdown } from "@fortawesome/free-brands-svg-icons";
-    import { faInfoCircle } from "@fortawesome/free-solid-svg-icons";
-    import { ResultCellStatusStyleMap } from "../../Common/TestStatus";
+    import {faMarkdown} from "@fortawesome/free-brands-svg-icons";
+    import {faInfoCircle, faChartLine} from "@fortawesome/free-solid-svg-icons";
+    import {ResultCellStatusStyleMap} from "../../Common/TestStatus";
     import Cell from "./Cell.svelte";
     import Fa from "svelte-fa";
-    import { sendMessage } from "../../Stores/AlertStore";
+    import {sendMessage} from "../../Stores/AlertStore";
+    import ResultsGraph from "../ResultsGraph.svelte";
+    import queryString from "query-string";
+    import GraphFilters from "./GraphFilters.svelte";
+    import ModalWindow from "../../Common/ModalWindow.svelte";
 
     export let table_name;
     export let result;
     export let selectedScreenshot;
+    export let test_id;
 
     let showDescription = false;
+    let showGraphModal = false;
+    let graph = null;
+    let ticks = {};
+    let releasesFilters = {};
+    let selectedColumnName = "";
+    let dateRange = 6;
+    let startDate = "";
+    let endDate = "";
+    let redraw = 0;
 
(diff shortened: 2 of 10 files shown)


frontend/TestRun/Components/ResultTable.svelte
```diff
@@ -100,37 +195,90 @@
...
                             {#each result.columns as col}
-                                {#key result.table_data[row][col.name]}
-                                    <td class="{ResultCellStatusStyleMap[result.table_data[row][col.name]?.status || 'NULL']}">
-                                        <Cell cell={result.table_data[row][col.name]} bind:selectedScreenshot/>
-                                    </td>
-                                {/key}
+                                <th>
+                                    <div class="column-header">
+                                        <span
+                                            >{col.name}
```
fruch, line 208: that's a weird formatting for html

k0machi: In this case it is justified because `span` is an inline element and adding newline after `>` would cause it to get rendered as an extra space. However since line is short, this can be shortened to 
`<span>{col.name}</span>`.

frontend/TestRun/Components/ResultTable.svelte
```diff
@@ -100,37 +195,90 @@
...
+                                            <span class="unit">{col.unit ? `[${col.unit}]` : ""}</span></span
+                                        >
+                                        {#if col.unit}
+                                            <button
```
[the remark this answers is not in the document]
k0machi: Usually tables should probably have the whole cell clickable for simplicity

frontend/TestRun/Components/ResultTable.svelte
k0machi: Nevermind, it's a span within a span, I'd say the first span could be omitted

frontend/TestRun/Components/ResultTable.svelte
```diff
@@ -83,14 +106,86 @@
...
+    async function handleReleaseChange(event) {
+        releasesFilters = event.detail.releasesFilters;
+        redraw++;
+    }
+
+    function handleKeydown(event) {
+        if (event.key === "Escape" && showGraphModal) {
+            closeGraphModal();
+        }
+    }
```
k0machi, line 172: This function should probably be moved into `ModalWindow.svelte` component itself probably (and just dispatch closeEvent)

frontend/TestRun/Components/ResultTable.svelte
```diff
@@ -19,6 +35,11 @@
         "table-secondary": "gray",
     };
 
+    const dispatch_run_click = (e) => {
```
k0machi, line 38: Wrong function naming convention

k0machi (changes requested):
Overall LGTM, some small comments here and there.

frontend/TestRun/Components/ResultTable.svelte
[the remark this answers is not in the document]
soyacz: how about that?
image

frontend/TestRun/Components/ResultTable.svelte
soyacz: moved

frontend/TestRun/Components/ResultTable.svelte
soyacz: it's natural for a python programmer :grin:

frontend/TestRun/Components/ResultTable.svelte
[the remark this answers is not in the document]
fruch: Maybe a bit more explanation ?
"Show this metric on job history graph"

(Or ask chatGPT to rephrase it)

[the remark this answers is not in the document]
soyacz: maybe 'show metric history'?

frontend/Common/ModalWindow.svelte
```diff
@@ -1,10 +1,24 @@
...
+    }
+
+    onMount(() => {
+        window.addEventListener("keydown", handleKeydown);
+    });
+
+    onDestroy(() => {
+        window.removeEventListener("keydown", handleKeydown);
+    });
 </script>
```
k0machi, line 22: Should be possible to use
```svelte
<svelte:window on:keydown={handleKeyDown} />
```
In the markup instead of manually doing this in the onmount/ondestroy

frontend/TestRun/Components/ResultTable.svelte
soyacz: additional span removed, I made also whole header cell to be clickable, not just the text

frontend/TestRun/Components/ResultTable.svelte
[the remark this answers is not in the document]
soyacz: I adjusted to 'show metric history'

k0machi (approved):

Issue #508, opened 2024-11-18
Improve graph tooltip


soyacz:
See if can be done:
1. copy text from the tooltip
2. idea: make it stick after clicking. When doing that, maybe there's a possibility to add buttons like: copy, go to run (as currently clicking on tooltip is reserved for going into run, which is not clear for users)

PR #570, opened 2025-01-14
improvement(graph): sticky tooltips

soyacz:
Graph tooltips now can be clicked to make it stick - eases screenshot, enables copying content. Also has clear button ("open") for opening related test. Tests are open in new popup window, so user doesn't lose focus from graph and can open multiple windows with tests.

closes: https://github.com/scylladb/argus/issues/508

image

### DIFF
--- a/frontend/TestRun/ResultsGraph.svelte
+++ b/frontend/TestRun/ResultsGraph.svelte
@@ -250,7 +250,6 @@
             },
         };
         graph.options.scales.x.min = ticks["min"];
-        graph.options.scales.x.max = ticks["max"];
         graph.options.scales.x.ticks = {
             minRotation: 45,
             callback: function (value, index) {
--- a/frontend/TestRun/ResultsGraph.svelte
+++ b/frontend/TestRun/ResultsGraph.svelte
@@ -1,8 +1,8 @@
 <script>
-    import {Chart, registerables, Tooltip} from "chart.js";
-    import {createEventDispatcher, onDestroy, onMount} from "svelte";
+    import { Chart, registerables, Tooltip } from "chart.js";
+    import { createEventDispatcher, onDestroy, onMount } from "svelte";
     import "chartjs-adapter-date-fns";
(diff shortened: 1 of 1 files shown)


soyacz:
> @soyacz you have it in staging ?

Deployed on staging

frontend/TestRun/ResultsGraph.svelte
```diff
@@ -94,56 +108,145 @@
...
+
+                        // Add corresponding changes
+                        const changes = tooltipModel.body[i]?.lines || [];
+                        changes.forEach((line) => {
+                            innerHtml += `<div class="tooltip-body">${line}</div>`;
+                        });
+
+                        // Add Open button for this point
+                        const runId = context.chart.data.datasets[dataPoint.datasetIndex].data[dataPoint.dataIndex].id;
+                        innerHtml += `<button class="tooltip-button" onclick="document.getElementById('graph-${test_id}-${index}').dispatchEvent(new CustomEvent('openRun', {detail: '${runId}'}))">Open</button>`;
```
fruch, line 206: Nitpick: can we make this button small, and a bit less taking focus?

fruch:
Also its still hard to click any dot that is at the boundary of the graph

Like the last result 

But maybe that for a different PR ?

frontend/TestRun/ResultsGraph.svelte
soyacz: how about this?
image

soyacz:
> Also its still hard to click any dot that is at the boundary of the graph
> 
> Like the last result
> 
> But maybe that for a different PR ?

yes, let's do it in different PR

frontend/TestRun/ResultsGraph.svelte
fruch: It's better, maybe it should go on the right side ? maybe the expend icon ?

soyacz: external link button added and moved it to right:
image

k0machi (approved):

PR #571, opened 2025-01-16
fix(build_system_monitor): Enable releng-testing

k0machi:


argus/backend/service/build_system_monitor.py
```diff
@@ -75,7 +75,7 @@ def check_filter(self, group_name: str) -> bool:
 class JenkinsMonitor(ArgusTestsMonitor):
 
     BUILD_SYSTEM_FILTERED_PREFIXES = [
-        "releng",
+
```
soyacz, line 78: won't it be too much?

Issue #496, opened 2024-11-04
Graphs descriptions


soyacz:
Result tables have description available, we should show it somehow on graphs.
Ideas:
1. Under graph titles
2. after hovering title/graph

PR #573, opened 2025-01-17
improvement(graphs): add subtitle from table description

soyacz:
Graphs got subtitle based on result table's description.

closes: https://github.com/scylladb/argus/issues/496

example:
image


fruch (approved):
LGTM

PR #574, opened 2025-01-20
improvement(ViewDashboard): Per-widget item filtering

### LINKED ISSUE #393: Views - Per Widget job filter
Add a per-widget setting to filter the jobs selected for a view, e.g. One test dashboard widget for rolling upgrades and one for longevity.

cc @roydahan @fruch

k0machi:
This commit implements per-widget entitity filtering for views,
allowing users to select which tests/groups/releases they want shown on
the widget in a specific view. To facilitate this, stat fetching has
been modified to accept widget id (position) of a view and then filter
tests accordingly. On the frontend side to facilitate conflict-free stat
fetching (for example from two TestDashboards) a hashmap has been
introduced to handle multiple stat objects and to dispatch them
correctly to each widget.

Caveats:
* Release Planner views in general are not supported and contain
  several unwanted interactions:
    * If test dashboard is modified to filter anything, release
      planner stat bar will be gray (since there's no more anchor
      for it to get the stats from)
    * Anything version related will conflict
* TestDashboard has a setting for a specific ScyllaVersion filter
  preset - this is not compatible and will make that filter set only
  output that version, which might be undesirable.
* Aside from TestDashboard, currently no other widgets implement
  stat fetch, so filtering for them might not be interesting.
  However the widget will now receive filtered set of tests, which
  should make the performance summaries automatically support this
  feature.

Fixes #393

### DIFF
--- a/frontend/Common/ViewTypes.js
+++ b/frontend/Common/ViewTypes.js
@@ -20,9 +20,12 @@ export class Widget {
         this.position = position;
         this.type = type;
         this.settings = settings;
+        this.filter = [];
     }
 }
 
+// sha1("")
+export const GLOBAL_STATS_KEY = "da39a3ee5e6b4b0d3255bfef95601890afd80709";
 
 export const WIDGET_TYPES = {
     UNSUPPORTED: {
@@ -47,6 +50,12 @@ export const WIDGET_TYPES = {
                 default: true,
                 help: "Include No Version results for stat fetches",
                 displayName: "Include No Version"
+            },
+            flatView: {
+                type: CheckValue,
+                default: false,
+                help: "Do not group tests by group",
+                displayName: "Flat View"
             },
             productVersion: {
                 type: StringValue,
--- a/frontend/Views/ViewDashboard.svelte
+++ b/frontend/Views/ViewDashboard.svelte
@@ -1,13 +1,20 @@
 <script>
     import Fa from "svelte-fa";
-    import { WIDGET_TYPES } from "../Common/ViewTypes";
     import { sendMessage } from "../Stores/AlertStore";
     import { faLink } from "@fortawesome/free-solid-svg-icons";
+    import { GLOBAL_STATS_KEY, WIDGET_TYPES } from "../Common/ViewTypes";
+    import sha1 from "js-sha1";
     export let view;
-    export let stats = undefined;
+    export let stats = {};
     export let productVersion;
     export let embedded = false;
     let clickedTests = {};
+    let resolvedTests = [];
+    const versionDispatch = {
+        GLOBAL_STATS_KEY: productVersion,
+    };
+
+    console.log(versionDispatch);
 
     const handleTestClick = function (detail) {
         if (detail.start_time == 0) {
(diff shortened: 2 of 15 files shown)


k0machi:
UI demo:

image

frontend/Views/ViewDashboard.svelte
```diff
@@ -45,6 +82,26 @@
         });
     };
 
+    const calculateWidgetStatsKey = function (widget) {
+        return sha1((widget.settings.filter ?? []).join(""));
+    };
+
+    const filterViewForWidget = async function (widget) {
+        if (!widget.settings.filter) {
```
soyacz, line 90: isn't in js that anempty array (`filter=[]`) evaluates to `true` (not like in the python)?

frontend/Views/ViewDashboard.svelte
```diff
@@ -45,6 +82,26 @@
...
+        tests = tests.reduce((acc, test) => {
+            acc[test.id] = test;
+            return acc;
+        }, {});
+        let viewCopy = structuredClone(view);
+        viewCopy.tests = viewCopy.tests.filter((item) => {
+            const test = tests[item.id] ?? {};
+            return ["id", "release_id", "group_id"].some((key) => widget.settings.filter.includes(test[key]));
+        });
+        return viewCopy;
```
soyacz, line 103: this way we return view object in case of no filters and view copy in case of filters present.
Maybe we should align it to not cause future issues?
Or even to add new `filteredTests` attribute which would be used in widgets?

soyacz (commented):
few comments

frontend/Views/ViewDashboard.svelte
k0machi: Fixed

frontend/Views/ViewDashboard.svelte
k0machi: Yes, I was deliberating whether or not always return a copy, so I think for safety we should always return a copy here. Fixed

soyacz:
I tested it locally with summary widget - it does not filter tests there (but at least does not break anything)

k0machi:
This PR will need some additional adjustments before it's ready:

- [ ] TestDashboard widget should have an option to display filtered objects as flat list
- [ ] Some additional testing to see if we have any problems with filtering.
- [ ] Stats fetching for views that exclusively have filters might have issues with overview stat widget, as it might not get fetched - need to account for that.

One thing that's going to be on the widget and not on this change is the follow up keep-up with the filtered set of tests - the frontend does the filtering based on the provided settings object and passes the filtered set into the widget, if widget ignores that and retrieves the view data again from some other endpoint - unfiltered, that's on the widget to account for.

k0machi:
Will probably block #666 as changes from here are quite extensive.

frontend/Common/ViewTypes.js
```diff
@@ -20,9 +20,11 @@ export class Widget {
         this.position = position;
         this.type = type;
         this.settings = settings;
+        this.filter = [];
     }
 }
 
+export const GLOBAL_STATS_KEY = "da39a3ee5e6b4b0d3255bfef95601890afd80709";
```
soyacz, line 27: can you add a docstring how to get it?

soyacz (approved):
Generally looks good, only one question.

PR #575, opened 2025-01-20
improvement(EventsTab): Multi-nemesis events

### LINKED ISSUE #470: `Events` tab only showing at most one disruption for any event, even when there are multiple ones
For example let's take https://argus.scylladb.com/test/5cd4bfe3-231b-4f48-99bc-f21ebd1b5efe/runs?additionalRuns%5B%5D=e330a762-61b1-4318-a55c-0b072d4df26e

we have some events there, e.g.:
Image
the event shows that there was a disruption: `disrupt_remove_node_then_add_node` at the time of the event (`2024-09-10 16:59:35.814`)

but from `Nemesis` tab we can learn that we had multiple disruptions running during this time:
Image
both `disrupt_remove_node_then_add_node` and `disrupt_create_index` started before `2024-09-10 16:59:35.814` and finished after that.

So the event should show both disruptions. I think the implementation assumes there's at most one disruption at any given moment, but SCT has parallel nemeses.

cc @temichus

k0machi:
Some events happen during multiple nemeses. This commit updates the
logic to find all nemeses that were active during a specific event,
instead of just looking for one like previously.

Fixes #470

### DIFF
--- a/frontend/TestRun/StructuredEvent.svelte
+++ b/frontend/TestRun/StructuredEvent.svelte
@@ -29,11 +29,9 @@
     <div class="event-header d-flex align-items-start flex-wrap">
         <div class="ms-2 mb-2 bg-dark text-light rounded px-2">{event.eventType}</div>
         <div class="ms-2 mb-2 rounded px-2 severity-{event.severity.toLowerCase()}">{event.severity}</div>
-        {#if event.nemesis}
-            <div class="ms-2 mb-2 rounded px-2 status-{event.nemesis.status.toLowerCase()}">{event.nemesis.name}</div>
-        {:else}
-            <div class="ms-2 mb-2 rounded px-2 status-skipped">no nemesis</div>
-        {/if}
+        {#each event.nemesis ?? [] as nemesis}
+            <div class="ms-2 mb-2 rounded px-2 status-{nemesis.status.toLowerCase()}">{nemesis.name}</div>
+        {/each}
         <div class="ms-auto mb-2 rounded px-2">
             <button
                 class="btn btn-light"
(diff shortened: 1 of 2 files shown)


k0machi:
image

fruch (approved):
LGTM

frontend/TestRun/StructuredEvent.svelte
```diff
@@ -29,11 +29,9 @@
     <div class="event-header d-flex align-items-start flex-wrap">
         <div class="ms-2 mb-2 bg-dark text-light rounded px-2">{event.eventType}</div>
         <div class="ms-2 mb-2 rounded px-2 severity-{event.severity.toLowerCase()}">{event.severity}</div>
-        {#if event.nemesis}
-            <div class="ms-2 mb-2 rounded px-2 status-{event.nemesis.status.toLowerCase()}">{event.nemesis.name}</div>
-        {:else}
-            <div class="ms-2 mb-2 rounded px-2 status-skipped">no nemesis</div>
-        {/if}
+        {#each event.nemesis ?? [] as nemesis}
+            <div class="ms-2 mb-2 rounded px-2 status-{nemesis.status.toLowerCase()}">{nemesis.name}</div>
```
soyacz, line 33: we lose the note about no nemesis

soyacz (approved):
one remark (I'm not sure if important). LGTM

frontend/TestRun/StructuredEvent.svelte
k0machi: Do you think we need to have that? If there's no note I feel it would serve the same purpose.

soyacz: I don't mind, just highlighting it. We can try without it.

fruch: I don't think anyone was really reading it, and would expect it

PR #576, opened 2025-01-21
fix(docs): status page lambda deploy

soyacz:
Deploying status page lambdas require to log in to AWS as `DevOpsAccessRole`. Added note about it.


soyacz:
btw. I fixed ip's to dns names in lambda and status page works again.

k0machi (approved):

PR #582, opened 2025-01-27
improvement(db-scripts): improve import/export scripts

soyacz:
Import/export scripts missed graphs data which is needed for debugging localy issues on prod. Added support for graphs data.

### DIFF
--- a/dev-db/export_data.py
+++ b/dev-db/export_data.py
@@ -1,33 +1,95 @@
 import json
 from pathlib import Path
+from collections import defaultdict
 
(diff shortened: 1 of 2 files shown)


soyacz:
> LGTM, maybe for a bit better usability we should also allow user to have a constantly ready prod config, so that they don't have change their dev config to download things

Yes, that could be done - in some next improvement. Basically, we need prod config for exporting and local config for importing - so we could actually hardcode it in these scripts. Anyway, prod config cannot be shared in GH.

PR #584, opened 2025-01-28
fix(plugins/sct): Fix serialization of ResourceState enum

k0machi:
The default value of state field was not being serialized properly due
to cqlengine being unable to handle Enum classes, fixed by getting the
Enum value in the default handler. This fixes an issue where SCT runner
would not be set with a vague "SyntaxException" error due to not
providing ResourceState.


fruch (approved):
LGTM

soyacz (approved):
LGTM

PR #586, opened 2025-02-10
fix(service/plans): Update attached views instead of recreating them

k0machi:
This commit fixes an issue where a view would be dropped and re-created,
removing user customizations to it. Instead, the view is updated and
user customizations are kept.


fruch (approved):
LGTM

roydahan (approved):

soyacz (approved):
LGTM

PR #587, opened 2025-02-10
fix(plugin/driver_matrix_tests): Use status from pipeline

k0machi:
This commit changes the way driver results are interpreted on argus
side. Instead of trying to guesstimate the overall status of the test
the status received from pipeline is now used.


fruch (approved):
LGTM

roydahan (approved):

soyacz (approved):
There are legacy parts that still use `_determine_run_status` and can be removed in a followup.
LGTM

PR #600, opened 2025-03-03
improvement(plugins/sct): Use 'target' packages to set version

### LINKED ISSUE #533: On Release dashboard a multi-run test (e.g. rolling upgrade) is marked as not planned, even though it was scheduled and started, but all runs are failed
A multi-run test (e.g. rolling-upgrade) is marked as not-run on Release dashboard, if all runs of the test are failed.

When investigating test results of a release it may seem that such a test was not planned. The assignee would notice that the test was actually executed (and failed) only when checking My Jobs page.

Example of such a not-run test on a release dashboard:
Image

And actual rub for the release in question:
Image

k0machi:
This commit adds additional logic to package/version submission APIs,
changing the way scylla_version field is set on an SCTTestRun (which
gets later used by TestDashboard widget to filter runs by version).
Primarily, the change is here to address rolling-upgrade runs being
hidden by version filters inside said dashboard when they fail without
upgrading to target version. Because of this, the logic now does the
following:
* A new field has been added to SCTTestRun class, named 'version_source'
    * Said field is used to restrict certain APIs from further changing
      the version based on certain conditions
* submit_product_version will now not set the version if version_source
  has any content (we ignore updates to it from SCT when it installs
  scylla)
* submit_packages will now check if a package has `target` in its name
  and do the following flow:
    * set the version_source and scylla_version if version_source is
      unset
    * if we are setting it again and the package is `upgrade-target`,
      override the scylla_version
    * if we are setting it again and the package is
      `scylla-server-target` while the `version_source` is set to
      `scylla-server-upgrade-target` we don't do anything.

Fixes #533

### DIFF
--- a/argus/backend/plugins/sct/service.py
+++ b/argus/backend/plugins/sct/service.py
@@ -63,6 +63,8 @@ class SCTService:
             run: SCTTestRun = SCTTestRun.get(id=run_id)
             for package_dict in packages:
                 package = PackageVersion(**package_dict)
+                if "target" in package.name:
+                    SCTService.process_target_version(run, package)
                 if package not in run.packages:
                     run.packages.append(package)
             run.save()
@@ -72,6 +74,13 @@ class SCTService:
 
         return "added"
 
+
+    @staticmethod
+    def process_target_version(run: SCTTestRun, package: PackageVersion):
+        if "upgrade-target" in run.version_source and package.name == "scylla-server-target":
+            return
+        run.version_source = package.name
+        run.scylla_version = package.version
 
     @staticmethod
     def set_sct_runner(run_id: str, public_ip: str, private_ip: str, region: str, backend: str, name: str = None):
--- a/argus/backend/plugins/sct/testrun.py
+++ b/argus/backend/plugins/sct/testrun.py
@@ -68,6 +68,7 @@ class SCTTestRun(PluginModelBase):
     config_files = columns.List(value_type=columns.Text())
     packages = columns.List(value_type=columns.UserDefinedType(user_type=PackageVersion))
     scylla_version = columns.Text()
+    version_source = columns.Text()
     yaml_test_duration = columns.Integer()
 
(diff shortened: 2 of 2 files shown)


fruch (approved):
LGTM

argus/backend/plugins/sct/service.py
```diff
@@ -63,6 +63,8 @@ def submit_packages(run_id: str, packages: list[dict]) -> str:
             run: SCTTestRun = SCTTestRun.get(id=run_id)
             for package_dict in packages:
                 package = PackageVersion(**package_dict)
+                if "target" in package.name:
+                    SCTService.process_target_version(run, package)
```
soyacz, line 67: we have scylla-server-target and scylla-server-upgrade-target packages. Depending which is last in the list it will decide - is it good enough?

argus/backend/plugins/sct/testrun.py
```diff
@@ -215,7 +216,8 @@ def get_events(self) -> list[EventsBySeverity]:
         return self.events
 
     def submit_product_version(self, version: str):
-        self.scylla_version = version
+        if not self.version_source:
```
soyacz, line 219: so basically we ignore if version_source is set?

soyacz (commented):
Few questions, otherwise LGTM

argus/backend/plugins/sct/testrun.py
k0machi: Yes.

argus/backend/plugins/sct/service.py
k0machi: So that's handled by `process_target_version`, the `upgrade-target` should always take priority, so the function checks if `version_source` has already been set to `upgrade-target` and denies `scylla-server-target` the update to the version. This should handle the submission in arbitrary order.

soyacz (approved):

PR #602, opened 2025-03-06
Find similar events

soyacz:
This PR brings new service for calculating embeddings of error/critical events
and finding similar events in the past test runs.
Basic flow is:
1. Fetch new events (from sct test runs table)
2. sanitize messages (by removing moving parts like file names, dates and redundant things that not bring benefits for similarity search)
3. generate embeddings -> store them in ScyllaDb (as list of floats, 384 dimensions) and in Chromadb (vector one)
4. find similar events and store it along with embedding in ScyllaDB

This works in a loop processing new tests soon after events are published.
If service is restarted, it fetches embeddings from ScyllaDB and processes these which have `None` similarities (processed ones, that could not find similars are marked as empty dict).
Then it get's events that were added after last processed event in embeddings db.

On Argus UI side, each event displays similar ones count inside a button `view XX similar events`.
When clicked, table with test runs with similar event will be shown along with job location, scylla version, date and attached issue list

Can be reviewed at: https://argus-staging.lab.dbaas.scyop.net/workspace (currently covering jobs from last 90 days)


soyacz:
@fruch to address issue with searching related event (after clicking related run, especially when there's many error events), maybe we should add expandable place to show related event? So user could quickly asses if it's false positive without opening run.

fruch:
> @fruch to address issue with searching related event (after clicking related run, especially when there's many error events), maybe we should add expandable place to show related event? So user could quickly asses if it's false positive without opening run.

Also an option, but I would really like the possibility to have links to specific events. (Regardless of this feature)

soyacz:
> > @fruch to address issue with searching related event (after clicking related run, especially when there's many error events), maybe we should add expandable place to show related event? So user could quickly asses if it's false positive without opening run.
> 
> Also an option, but I would really like the possibility to have links to specific events. (Regardless of this feature)

https://github.com/scylladb/argus/issues/604
Anyway, we don't store info about event index in related run id, so I need to add this info.

k0machi:
I will take a look at the PR in detail soon as it is quite large, but first question that comes to mind - why is it a subproject managed by a different package manager (uv vs poetry)? Is there something poetry can't do? 

And regarding disconnections - I see code that tries to manage the session/cluster object - this should not be necessary, python driver does it by itself (you can see argus does nothing aside from initial ScyllaCluster.get() at the start of the application and it runs no problem) - You should be able to get `session` from `db` object returned by the `.get()` method.

soyacz:
> I will take a look at the PR in detail soon as it is quite large, but first question that comes to mind - why is it a subproject managed by a different package manager (uv vs poetry)? Is there something poetry can't do?
> 

Actually I didn't try with poetry. With uv it's easy as even does not require venv to be created before and it bootstraps project super fast.

> And regarding disconnections - I see code that tries to manage the session/cluster object - this should not be necessary, python driver does it by itself (you can see argus does nothing aside from initial ScyllaCluster.get() at the start of the application and it runs no problem) - You should be able to get `session` from `db` object returned by the `.get()` method.

This was my initial approach - see 1st commit only (the second one is for tackling this issue). Eventually driver reconnects, but it may take several minutes - I was hoping to fix it.
But after today's talk with Israel, we want to push it quickly to prod as even with current shape it brings a lot of benefits and we can tackle problems in follow up commits.
Two things before merging is to:
1. add event index to similarities so we can in future link directly to referenced event (not just run id)
2. lower similarity threshold as after sanitization we can see some false positives

soyacz:
@k0machi I marked this PR as ready for review.

Current approach does not have issue with db disconnection. Probably it was due missing paging configuration (now fetch only 10 per page, I believe default could be too big to fit page memory causing this issue - but that's my guess only).
Anyway I did some improvements about db connection stability.
Now I process again all the events on staging env (temporarily we might not see similar events thou) - because I changed a bit sanitization and wanted to test db stability.

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,266 @@
+import os
+from pathlib import Path
+import chromadb
+from threading import Thread, Event
+import time
+import logging
+import datetime
+import uuid
+
+from chromadb.utils.embedding_functions.onnx_mini_lm_l6_v2 import ONNXMiniLM_L6_V2
```
swasik, line 10: Would be nice to have one sentence with description of this model (when it was trained, based on what data, quality, etc.)

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,266 @@
...
+import itertools
+from argus.backend.models.argus_ai import ErrorEventEmbeddings, CriticalEventEmbeddings
+from argus.backend.plugins.sct.testrun import SCTTestRun
+from argusAI.utils.event_message_sanitizer import MessageSanitizer
+from argusAI.utils.scylla_connection import ScyllaConnection
+
+SLEEP_INTERVAL = 300  # 5 minutes
+SIMILARITY_THRESHOLD = 0.90
+MAX_SIMILARS = 20
+DAYS_BACK = int(os.getenv('DAYS_BACK', 90))
```
swasik, line 22: Why do we have different default (120) in argusAI/deployment/argusai_event_similarity_processor.service? Would be great if we can somehow synchronize those files.

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,266 @@
...
+stop_event = Event()
+
+class EventType:
+    ERROR = "ERROR"
+    CRITICAL = "CRITICAL"
+
+class EventProcessor:
+    def __init__(self):
+        """Initialize the EventProcessor with ChromaDB client and collections."""
+        self.client = chromadb.Client(Settings(anonymized_telemetry=False))
```
swasik, line 50: I would name it chromadb_client as in the same class we also store connection to Scylla (in self.db) and it can be non-trivial to understand in the code which client it is.

argusAI/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[project]
+name = "windsurf-project"
+version = "0.1.0"
+description = "Add your description here"
```
swasik, line 4: Would be nice to have at least few relevant words here.

argusAI/utils/event_message_sanitizer.py
```diff
@@ -0,0 +1,164 @@
+import re
+from datetime import datetime
+
+from typing import Pattern, List, Callable
+from uuid import UUID
+
+
+class MessageSanitizer:
```
swasik, line 8: Good work on this! I see that you cover plenty of cases. Of course it will be crucial to work on the feedback from users to tune it later but as a starting point looks good.

argusAI/utils/event_message_sanitizer.py
```diff
@@ -0,0 +1,171 @@
...
+    def sanitize(self, message: str) -> str:
+        result = message
+        for sanitizer in self.sanitizers:
+            result = sanitizer(result)
+        self.sanitized_messages_fp.write(result + "_" + '\n')
+        return result
+
+    # New method to remove Disk Size and Message
+    def remove_disk_size_and_message(self, text: str) -> str:
+        text = self.disk_size_pattern.sub('', text)
```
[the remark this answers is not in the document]
soyacz: fixed

argusAI/event_similarity_processor.py
soyacz: I added sentence - but in a different place where I actually used different model

argusAI/event_similarity_processor.py
soyacz: fixed

argusAI/event_similarity_processor.py
soyacz: fixed

soyacz:
ping @k0machi

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,274 @@
...
+    MODEL_NAME = "bge-small-en-v1.5"
+    DOWNLOAD_PATH = Path.home() / ".cache" / "chroma" / "onnx_models" / MODEL_NAME
+    EXTRACTED_FOLDER_NAME = "onnx"
+    ARCHIVE_FILENAME = "onnx.tar.gz"
+    MODEL_DOWNLOAD_URL = (
+        "https://scylla-qa-public.s3.us-east-1.amazonaws.com/ArgusAI/bge-small-en-v1.5/onnx.tar.gz"
+    )
+    _MODEL_SHA256 = "e7d1743b0c08f55c687cff6af696683398682f9ab4fb3cad1be644ee5553a72d"
+
+logger = logging.getLogger(__name__)
```
k0machi, line 48: Should be uppercase

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,274 @@
...
+stop_event = Event()
```
k0machi, line 49: Ditto

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,274 @@
...
+            embedding_entry (dict): Event data including run_id, event_index, and embedding
+            similarity_threshold (float): Minimum similarity score for matches
+            max_similars (int): Maximum number of similar events to find
+        """
+        collection = self.collections[event_type]
+        run_id = embedding_entry.run_id
+        event_index = embedding_entry.event_index
+        embedding = embedding_entry.embedding
+
+        logger.info(f"Finding similarities for run_id: {run_id}, event_index: {event_index}, type: {event_type}")
```
k0machi, line 84: f-string instead of `%s` formatter

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,274 @@
...
+            if 1 - distance >= similarity_threshold:
+                metadata = similar["metadatas"][0][j]
+                similar_map.update({uuid.UUID(metadata["run_id"]): metadata["idx"]})
+                self.similars_found += 1
+                if len(similar_map) >= max_similars:
+                    break
+
+        table = ErrorEventEmbeddings if event_type == EventType.ERROR else CriticalEventEmbeddings
+        try:
+            self.db.execute(
```
k0machi, line 104: Maybe a prepare statement here

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        Continuously process new events from test runs.
+
+        Args:
+            last_processed_time (datetime): Timestamp of the last processed event
+        """
+        while not stop_event.is_set():
+            logger.info(f"Starting event processing cycle; last processed time: {last_processed_time}")
+            try:
+                query = f"SELECT id, start_time, events FROM {SCTTestRun.table_name()} WHERE start_time > %s ALLOW FILTERING BYPASS CACHE"
```
k0machi, line 172: Prepared statement

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,274 @@
...
+                        for event_idx, embedding in enumerate(embeddings):
+                            event_id = f"{test_run.id}_{idx}_{event_idx}"
+                            metadata = {"run_id": str(test_run.id), "idx": event_idx}
+                            self.collections[event_type].upsert(
+                                ids=[event_id],
+                                embeddings=[embedding],
+                                metadatas=[metadata],
+                            )
+
+                            query = f"INSERT INTO {table.__table_name__} (run_id, event_index, embedding, start_time) VALUES (%s, %s, %s, %s)"
```
k0machi, line 206: Should probably be prepared here as well

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,274 @@
...
+from argus.backend.plugins.sct.testrun import SCTTestRun
+from argusAI.utils.event_message_sanitizer import MessageSanitizer
+from argusAI.utils.scylla_connection import ScyllaConnection
+
+SLEEP_INTERVAL = 300  # 5 minutes
+SIMILARITY_THRESHOLD = 0.90
+MAX_SIMILARS = 20
+DAYS_BACK = int(os.getenv('DAYS_BACK', 120))
+
+logging.basicConfig(
```
k0machi, line 25: You could use the logging module we have on the backend so that message format would be the same between those two.

argusAI/utils/event_message_sanitizer.py
```diff
@@ -0,0 +1,197 @@
+import re
+from datetime import datetime
+
+from typing import Pattern, List, Callable
+from uuid import UUID
+
+
+class MessageSanitizer:
```
k0machi, line 8: I think it may be a good idea to provide a raw event and a fully sanitized event example here, just to get a quick glance of how events looks pre and post.

frontend/TestRun/StructuredEvent.svelte
```diff
@@ -19,22 +22,74 @@
...
+                    if (data.status === "ok") {
+                        similarRunsInfo = data.response;
+                    } else {
+                        console.error("Error fetching similar runs info:", data);
+                    }
+                } else {
+                    console.error("Failed to fetch similar runs info:", response.statusText);
+                }
+            } catch (error) {
+                console.error("Error fetching similar runs info:", error);
```
k0machi, line 53: Missing user facing message

k0machi (changes requested):
Overall looks good, just some minor things here and there.

argusAI/event_similarity_processor.py
fruch: I think f-strings are just fine

and we want something like that enforce, it's with a linter (in pre-commit hooks)

otherwise it's just it's just a waste of energy/mental shares

argusAI/event_similarity_processor.py
```diff
@@ -0,0 +1,274 @@
...
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s] %(message)s",
+    datefmt="%Y-%m-%d %H:%M:%S",
+)
+
+EmbeddingEntry = namedtuple('EmbeddingEntry', ['run_id', 'event_index', 'embedding', 'start_time'])
+
+class BGE_SMALL(ONNXMiniLM_L6_V2):
```
fruch, line 33: what's BGE ?

argusAI/event_similarity_processor.py
k0machi: Okay

soyacz:
@k0machi thanks for your feedback.
Still I found one issue - long tests might be omitted and I need to fix it.

argusAI/utils/event_message_sanitizer.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+        A class used to sanitize event messages by removing or replacing sensitive information
+        such as IP addresses, URLs, file paths, and other identifiable data.
+
+        Sanitized messages are written to a file for further analysis.
+    """
+    def __init__(self):
+        self.event_pattern: Pattern = re.compile(r'Severity\.(ERROR|CRITICAL|WARNING|INFO)\)')
+        self.event_id_pattern: Pattern = re.compile(r'event_id=[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}')
```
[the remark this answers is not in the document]
soyacz: right, fixed

frontend/TestRun/StructuredEvent.svelte
soyacz: fixed

argusAI/utils/event_message_sanitizer.py
soyacz: sanitized events are collected (with run id info) for analysis on host running it
Would be worth to add unit tests for that.
@fruch shall I do it within this PR or followup?

k0machi (approved):
LGTM, just one small comment about the button icon.

frontend/TestRun/StructuredEvent.svelte
```diff
@@ -19,22 +22,75 @@
...
+            >
+        </div>
         {#each event.nemesis ?? [] as nemesis}
             <div class="ms-2 mb-2 rounded px-2 status-{nemesis.status.toLowerCase()}">{nemesis.name}</div>
         {/each}
-        <div class="ms-auto mb-2 rounded px-2">
+        <div class="ms-auto mb-2 rounded px-2 d-flex gap-2">
+            {#if (event.severity === "ERROR" || event.severity === "CRITICAL") && similars.length > 0}
+                <button class="btn btn-warning btn-sm d-flex align-items-center gap-2" on:click={toggleSimilars}>
+                    <Fa icon={faExpand} /><span>View {similars.length} Similar Events</span>
```
k0machi, line 88: I think `faSearchengin` might look better here.

k0machi: or `faMagic`

soyacz: I decided to go with faLink - as magic wand is not intuitive either. 
If it's a problem we can discuss it and prepare a quick fix.

PR #605, opened 2025-03-10
fix(ViewTypes.js): Rename duplicate key

k0machi:
This fixes missing ReleaseStats widget


soyacz:
my bad. How was it surfacing?

PR #606, opened 2025-03-17
fix(ReleasePlanner): Allow searching assignees by Full Name

### LINKED ISSUE #603: Search by person according their display name as well
Currently, if you search for a specific person in assignee field, you need to enter their github username and not their actual name. This makes it hard to search for users who do not have the same github name and their real (display) name.

We should allow to searching real names as well

k0machi:
Some assignee selectors were missing filter function that would allow
them to be filtered by real name of the user.

Fixes #603


pehala (approved):

fruch (approved):
LGTM

soyacz (approved):
LGTM

PR #608, opened 2025-03-17
improvement(WorkArea): Add Test Execute button

### LINKED ISSUE #542: Job execute button
We should add a button similar to the global clone button, except for executing the job (or rebuilding).

k0machi:
This commits adds a new button to the test runs within argus, allowing
user to either start a build from scratch or rebuild a latest run
without having to open a run to do so. The from scratch implementation
will try to retrieve previous build parameters (In case previous build
wasn't submitted to argus) and if that's not possible, will offer user
to start a build without parameters to be able to retrieve the build on
subsequent call.

Fixes #542

### DIFF
--- a/argus/backend/service/jenkins_service.py
+++ b/argus/backend/service/jenkins_service.py
@@ -50,8 +50,18 @@ class JenkinsService:
                                         username=current_app.config["JENKINS_USER"],
                                         password=current_app.config["JENKINS_API_TOKEN"])
 
-    def retrieve_job_parameters(self, build_id: str, build_number: int) -> list[Parameter]:
-        job_info = self._jenkins.get_build_info(name=build_id, number=build_number)
+    def retrieve_job_parameters(self, build_id: str, build_number: int | None) -> list[Parameter]:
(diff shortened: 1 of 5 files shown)


argus/backend/service/jenkins_service.py
```diff
@@ -51,7 +51,16 @@ def __init__(self) -> None:
                                         password=current_app.config["JENKINS_API_TOKEN"])
 
     def retrieve_job_parameters(self, build_id: str, build_number: int) -> list[Parameter]:
-        job_info = self._jenkins.get_build_info(name=build_id, number=build_number)
+        if build_number == -1:
```
soyacz, line 54: I believe -1 is just not known/non existent - better use `build_number: Optional[int]` and use `None`

k0machi: Done

k0machi:
Added additional logic to handle executed jobs without parameters (first build as done by SCT by executing without parameters)

soyacz (approved):
LGTM

PR #613, opened 2025-03-20
Add option to pass on extra headers into Argus clients

fruch:
for purpose of identificaton of the code part sending out results into argus, we should be able to configure as needed the headers the client are sending out.

this change introducs extra argument to clients, and to thier CLIs

### DIFF
--- a/argus/client/base.py
+++ b/argus/client/base.py
@@ -34,14 +34,19 @@ class ArgusClient:
         FETCH_RESULTS = "/testrun/$type/$id/fetch_results"
         FINALIZE = "/testrun/$type/$id/finalize"
 
-    def __init__(self, auth_token: str, base_url: str, api_version="v1") -> None:
+    def __init__(self, auth_token: str, base_url: str, api_version="v1", extra_headers: dict | None = None) -> None:
         self._auth_token = auth_token
         self._base_url = base_url
         self._api_ver = api_version
+        self._extra_headers = extra_headers or {}
 
     @property
     def auth_token(self) -> str:
         return self._auth_token
+
+    @property
+    def extra_headers(self) -> dict:
+        return self._extra_headers
 
     def verify_location_params(self, endpoint: str, location_params: dict[str, str]) -> bool:
         required_params: list[str] = re.findall(r"\$[\w_]+", endpoint)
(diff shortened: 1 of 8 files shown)


soyacz:
how we'll be getting these headers? I believe anyway we won't be willing to pass them in commandline due security concerns.
Maybe instead of additional argument, these could take headers from file?

k0machi:
> how we'll be getting these headers? I believe anyway we won't be willing to pass them in commandline due security concerns. Maybe instead of additional argument, these could take headers from file?

This might be only relevant to the CLI client, otherwise I think it's up to the client user to implement header passing to the function.

fruch:
> how we'll be getting these headers? I believe anyway we won't be willing to pass them in commandline due security concerns. Maybe instead of additional argument, these could take headers from file?

there are multiple places where we already pass down the token, some of it are via command-line.
those won't be much different in that sense

anyhow using command line or using environment variables, we can easily switch to environment variables

I'll add support for it into the commands as well.

fruch:
V2:

add environment variables for the option introduced in this PR
so user of it would be able not to pass those in commandline arguments

fruch:
@soyacz @k0machi can you give it one more look, this is the first in-line of a few one we need todo (i.e. all of the clients using it)

soyacz:
> @soyacz @k0machi can you give it one more look, this is the first in-line of a few one we need todo (i.e. all of the clients using it)

I think its ok.
Remember to update `version` in pyproject.toml

fruch:
> > @soyacz @k0machi can you give it one more look, this is the first in-line of a few one we need todo (i.e. all of the clients using it)
> 
> I think its ok. Remember to update `version` in pyproject.toml

who uploads them into pypi ?

k0machi (approved):
LGTM

k0machi:
> > > @soyacz @k0machi can you give it one more look, this is the first in-line of a few one we need todo (i.e. all of the clients using it)
> > 
> > 
> > I think its ok. Remember to update `version` in pyproject.toml
> 
> who uploads them into pypi ?

Me and Lukasz usually do it, but if you have write access to the package (I think you do), you should be able to upload using poetry:
```
poetry publish --build
```

PR #615, opened 2025-03-20
fix(ReleasePlannerGridView): Null-check group stats and force fetch

k0machi:
This fix fixes an issue where GridView would crash if a release was set
to dormant (so stat fetches don't happen) as well as it forces the stat
fetch to occur even if release is dormant.


fruch (approved):
LGTM

soyacz (commented):
LGTM

PR #616, opened 2025-03-20
improvement(ReleasePlanCreator): Copy assigned user into expanded group

### LINKED ISSUE #612: Release Plans - Owner isn't being moved into an expanded group
Currently the Plan editor does not add the assignee to the expanded group.

k0machi:
This commit makes it so that if a user was assigned to a group before
expansion that user is copied to all tests that will be added to the
plan.

Fixes #612


fruch (approved):

soyacz (approved):
LGTM

Issue #619, opened 2025-03-25
argus-alm can't be installed into python 3.10


fruch:
since:
https://github.com/scylladb/argus/commit/d8442760a05dc908cc0fbf4ed6a39182dba8fb51

we can't install argus-alm into hydra instance

we might need it in siren-tests, and we are currently blocked casue of it.

PR #620, opened 2025-03-26
fix(installation): allowed using python 3.10 and updated packages

soyacz:
Because of restriction enforced in pyproject.toml it is not possible to install argus client in 3.10 envirionments.
Loosen that restriction and updated packages.

fixes: https://github.com/scylladb/argus/issues/619

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -27,25 +27,26 @@ files = [
 
 [package.dependencies]
 lazy-object-proxy = ">=1.4.0"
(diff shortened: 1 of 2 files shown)


fruch (approved):
LGTM

fruch:
@soyacz 
lets' have https://github.com/scylladb/argus/pull/621 in, before the next release

k0machi (approved):

PR #631, opened 2025-03-27
fix(driver_matrix_test.cli): add missing extra_headers arguments

fruch:
in the finish-run command the `extra_headers` argumnet was missing while it was define as click option, and when it would be passed to the function it would fail

this introduce the argument, and pass it to the ArgusClient


fruch:
and we'll need a new client version for this one.

soyacz (approved):

soyacz:
> and we'll need a new client version for this one.

can you bump also version in this PR then?

fruch:
> > and we'll need a new client version for this one.
> 
> can you bump also version in this PR then?

now I did

soyacz:
> > > and we'll need a new client version for this one.
> > 
> > 
> > can you bump also version in this PR then?
> 
> now I did

ugh, I didn't check and merged - you didn't :sweat_smile:

soyacz:
bumped and pushed to PyPi

Issue #625, opened 2025-03-26
Test Stats Widget - similar to Nemesis Widget


soyacz:
Nemesis widget does not show full picture, while current stats shown in views are very limited.
There's idea to create similar Test Stats Widget that is similar to Nemesis Widget covering:
1. Pie chart with overall tests stats (passed, failed, not run(?) etc.)
2. Hours spend on testing
3. Unique tests run
4. Per test bar chart with number of pass/failures (like Nemesis widget but x axis will show tests)
5. drill down to test (?)

Basic version filtering like in Nemesis Widget is required. Possibly excluding some test groups would be needed due too many tests.

refs: https://github.com/scylladb/argus/issues/446

soyacz:
Ideas for further develpoment and usecases: https://scylladb.atlassian.net/wiki/x/PgA5B

PR #636, opened 2025-04-08
feature(graphed-stats): new view widget

soyacz:
Introducing new View widget (to replace nemesis stats one) that will include graphed stats of tests and nemesis.

Widget will present graphed test statistics: pie chart with test pass/failure rate stats, bar chart with more details.
Tests bar chart presents tests in hierarchical view and can be drilled down to the single test level where all runs are shown in table.
Nemesis charts are like in Nemesis Widget with additional filter based on test drill down level.

refs: https://github.com/scylladb/argus/issues/625

Available on staging env. Example screenshot:
image

### DIFF
--- a/argus/backend/controller/views_widgets/graphed_stats.py
+++ b/argus/backend/controller/views_widgets/graphed_stats.py
@@ -19,8 +19,11 @@ def get_graphed_stats():
         "test_runs": [],
         "nemesis_data": []
     }
+
+    filters = request.args.get("filters")
+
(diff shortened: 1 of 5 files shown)


k0machi (approved):

soyacz:
@k0machi I fixed problem of default sorting (now configurable) and added one more commit with feature request from Roy (allowing filtering out some tests).
Can you take one more look at this commit?

soyacz:
ping @k0machi

Issue #604, opened 2025-03-10
Allow linking to specific run tab


soyacz:
When sharing a link to a run, we often want to link to specific tab like: results, events or discussion or any other one.
More, when we share link to events tab we might want to link specific event.

soyacz:
Linking to events tab, should allow to link also to specific event by its severity+index (and scroll down to it)

PR #638, opened 2025-04-14
improvement(TestRun): allow linking to secific tab

soyacz:
Added possibility to link to specific tab in TestRun view.

closes: https://github.com/scylladb/argus/issues/604

### DIFF
--- a/frontend/TestRun/DriverMatrixTestRun.svelte
+++ b/frontend/TestRun/DriverMatrixTestRun.svelte
@@ -24,13 +24,23 @@
     export let runId = "";
     export let buildNumber = -1;
     export let testInfo = {};
+    export let tab = "";
     const dispatch = createEventDispatcher();
     let testRun = undefined;
     let runRefreshInterval;
-    let activityOpen = false;
-    let commentsOpen = false;
-    let issuesOpen = false;
+    let activeTab = tab.toLowerCase() || "details";
     let failedToLoad = false;
+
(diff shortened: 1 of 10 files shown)


frontend/TestRun/DriverMatrixTestRun.svelte
```diff
@@ -24,14 +24,24 @@
...
+    let activeTab = tab.toLowerCase() || "details";
     let failedToLoad = false;
 
+    const setActiveTab = (tabName) => {
+        tabName = tabName.toLowerCase();
+        if (tabName !== activeTab) {
+            activeTab = tabName;
+            if (!window.location.pathname.startsWith("/workspace")) {
+                const newUrl = `/tests/${testInfo.test.plugin_name}/${runId}/${tabName}`;
+                history.replaceState({}, "", newUrl);
```
k0machi, line 40: You should probably use `pushState` here and handle `statePop` even so that user can go back and forth between tabs using the history buttons

k0machi (changes requested):
Overall looks good, aside from a minor change to how we handle navigation between tabs

frontend/TestRun/DriverMatrixTestRun.svelte
soyacz: I'm not sure about that, usually when jumping between tabs (without reload), feels that go back shouldn't jump tabs - users rather click specific tab instead going back with browser

PR #640, opened 2025-04-17
Add exclusions to parsejUnitXml in jUnitResults

Bouncheck:
Excludes child nodes `system-out` and `system-err` when looking for nodes useful for deducing test status. Currently if `system-out` is present for passing test it will in the end be classified as "Disabled".

### DIFF
--- a/frontend/TestRun/jUnitResults.svelte
+++ b/frontend/TestRun/jUnitResults.svelte
@@ -38,7 +38,7 @@
         let cases = Array
             .from(doc.querySelectorAll("testcase"))
             .map(raw_case => {
-                let statusNode = raw_case.childNodes.length > 0 ? Array.from(raw_case.childNodes).filter(v => v.nodeName != "#text")[0] : undefined;
+                let statusNode = raw_case.childNodes.length > 0 ? Array.from(raw_case.childNodes).filter(v => v.nodeName !== "#text" && v.nodeName !== "system-out" && v.nodeName !== "system-err")[0] : undefined;
                 return {
                     name: raw_case.attributes.getNamedItem("name").textContent,
                     class: raw_case.attributes.getNamedItem("classname").textContent,


Bouncheck:
@vponomaryov I would appreciate if you could take a look, since I am unfamiliar with argus. Maybe there are other places that handle test results.
I cannot add reviewers myself.

Bouncheck:
Tested in browser console with the following example:
```
const rawXml = `
<testsuite>
    <testcase name="should_pass" classname="com.example.MyTest" time="0.001"/>
    <testcase name="should_fail" classname="com.example.MyTest" time="0.002">
        <failure message="AssertionError">Expected true but got false</failure>
    </testcase>
    <testcase name="should_log" classname="com.example.MyTest" time="0.003">
        <system-out />
    </testcase>
</testsuite>
`;

const parsejUnitXml = function(raw_content) {
        let parser = new DOMParser();

        let doc = parser.parseFromString(raw_content, "text/xml");
        let cases = Array
            .from(doc.querySelectorAll("testcase"))
            .map(raw_case => {
                let statusNode = raw_case.childNodes.length > 0 ? Array.from(raw_case.childNodes).filter(v => v.nodeName !== "#text" && v.nodeName !== "system-out" && v.nodeName !== "system-err")[0] : undefined;
                return {
                    name: raw_case.attributes.getNamedItem("name").textContent,
                    class: raw_case.attributes.getNamedItem("classname").textContent,
                    time: new Number(raw_case.attributes.getNamedItem("time").textContent),
                    status: statusNode ? statusNode.nodeName : "passed",
                    message: statusNode ? statusNode.textContent : "",
                };
            });

        return cases;
};

console.log(parsejUnitXml(rawXml));
```
"should_log" receives status passed in such setting

vponomaryov:
> @vponomaryov I would appreciate if you could take a look, since I am unfamiliar with argus. Maybe there are other places that handle test results. I cannot add reviewers myself or assign myself to the PR.

I am in the same boat, added proper reviewers.

soyacz:
Maybe I don't understand something, but why not to fix it on java driver matrix tests part? E.g. not sending results when there was no tests or just fail if all skipped?
Argus serves as a display only and we should try avoid passed/failed logic at this level.

Bouncheck:
> Maybe I don't understand something, but why not to fix it on java driver matrix tests part? E.g. not sending results when there was no tests or just fail if all skipped? Argus serves as a display only and we should try avoid passed/failed logic at this level.

I think the display is incorrect in this case. The tests passed and summary xml file was generated. The problem is that in some versions of java driver a passing test can have additional sections in xml file which tricks argus into thinking it's not passing. But it's still a valid junit xml.
The difference is the same as in this example:
```
<testcase name="passed" classname="com.example.MyTest" time="0.001"/>

<testcase name="passed_and_has_system_out_section" classname="com.example.MyTest" time="0.003">
    <system-out />
</testcase>
```
The first one will show in `PASSED` category in argus, but the second one will be in `DISABLED` with system-out tile as can be seen in 3.11.4.0 results of
https://argus.scylladb.com/tests/driver-matrix-tests/9845e40a-6f3c-4ecd-b974-b0101951b963

Technically this could be fixed on driver matrix side, for example with processing the xml there first and removing all system-out and system-err sections. This is technically not a problem until we would want to utilize them somehow. It feels a bit backwards though.

soyacz (approved):

soyacz:
@k0machi can you take a look?

PR #642, opened 2025-04-22
refactor(GithubIssues): Introduce caching and rework issue store

### LINKED ISSUE #546: Cache github issues details
We use user token to retrieve details for GH issues. This causes the need to ask for additional permissions upon logging in to Argus and complicates displaying issues.
E.g. copy issues modal won't show details if issue was not updated by client (which is the case for Summary View Widget and happens in View issues list for some issues sometimes).

Idea is to periodically update basic issue details (like assignee, status, tags, title) by querying GH (e.g. by fetching only latest changes or like we do in SCT/dtest in `SkipPerIssue` feature). Set up cron that will run ArgusCli command which does that.

### LINKED ISSUE #585: GH issues titles are not downloaded
In issues tab, instead of title, `scylladb/scylla-enterprise#5174` is shown:
https://argus.scylladb.com/tests/scylla-cluster-tests/c6906c6f-7635-4148-ba09-2869d3479133

k0machi:
This commit reworks how the issues from Github are stored on the argus
side. Previously, an issue entity was created each time user linked an
issue to the run. Now, the mechanism has been reworked to instead create
a unique issue entity and a special helper table is used to link that
issue to runs (and other filtering keys) called IssueLink.

This change also allowed to introduce proper issue caching - a script was
introduced to refresh Issue state, labels and assignees at arbitrary points.
The script queries unique repositories present in Argus database and
requests updated issue lists since a specified timestamp. The timestamp
is stored inside newly introduced RuntimeStore table, which can be used
for arbitrary configuration storage during runtime. This timestamp is
refreshed on each issue refresh, reducing amount of requests needed to
update issue states.

A migration script is included with the commit to migrate issues from
the old table and every component depending on Issue functionality was
updated to reflect new changes.

With this changeset, it is now possible to remove the 'repo' scope
request for the users, as issue state is no longer fetched using User's
Oauth token. However, this commit does not implement JWT Tokens yet and
is expected to use a personal access token for now.

Fixes #546
Fixes #585


argus/backend/service/github_service.py
```diff
@@ -0,0 +1,232 @@
...
+from argus.backend.models.github_issue import GithubIssue, IssueAssignee, IssueLink, IssueLabel
+from argus.backend.plugins.core import PluginInfoBase
+from argus.backend.plugins.loader import AVAILABLE_PLUGINS
+from argus.backend.service.event_service import EventService
+from argus.backend.util.common import chunk
+
+LOGGER = logging.getLogger(__name__)
+
+
+class GithubService:
```
soyacz, line 21: general question, how will it behave when someone moves the issue?

argus/backend/service/github_service.py
```diff
@@ -0,0 +1,232 @@
...
+                match = re.match(
+                    r"http(s)?://(www\.)?github\.com/(?P<owner>[\w\d]+)/"
+                    r"(?P<repo>[\w\d\-_]+)/(?P<type>issues|pull)/(?P<issue_number>\d+)(/)?",
+                    issue.html_url,
+                )
+
+                identifier = f"{match.group('owner').lower()}/{match.group('repo').lower()}#{match.group('issue_number')}"
+                issue_to_update = issues_by_identifier.get(identifier)
+                if not issue_to_update:
+                    LOGGER.info("[%s/%s] No issue found for %s...", issue_idx + 1, "?", identifier)
```
soyacz, line 71: shouldn't be info - there will be pretty many of these polluting logs.

argus/backend/service/github_service.py
```diff
@@ -0,0 +1,232 @@
...
+                    continue
+                LOGGER.info("[%s/%s] Refreshing %s...", issue_idx + 1, "?", identifier)
```
soyacz, line 73: I'd keep it as debug

argus/backend/service/github_service.py
```diff
@@ -0,0 +1,232 @@
...
+                issue_to_update.title = issue.title
+                issue_to_update.state = issue.state
+                issue_to_update.labels = [IssueLabel(id=label.id, name=label.name, color=label.color, description=label.description) for label in issue.labels]
+                issue_to_update.assignees = [IssueAssignee(login=assignee.login, html_url=assignee.html_url) for assignee in issue.assignees]
+                issue_to_update.save()
+                update_count += 1
```
soyacz, line 79: actually, this number might not be true - if there won't be changes in title/state/labes/assignees then it won't update anything

argus/backend/service/github_service.py
```diff
@@ -0,0 +1,232 @@
...
+
+        LOGGER.info("%s", issues_by_identifier.keys())
+        LOGGER.info("Finished. Updated %s out of %s issues", update_count, len(all_issues))
+        last_ran.value = datetime.now(tz=UTC)
```
soyacz, line 83: this one should be from beginning of the process, we might miss if someone updated an issue during it

pyproject.toml
```diff
@@ -39,6 +39,7 @@ python-magic = "^0.4.24"
 uwsgi = "^2.0.20"
 python-jenkins = "^1.7.0"
 python-slugify = "^6.1.1"
+pygithub = "^2.6.1"
```
soyacz, line 42: nice, does it support these big queries with paging?

argus/backend/models/github_issue.py
```diff
@@ -0,0 +1,60 @@
...
+            return self.name == other.name and self.color == other.color and self.description == other.description
+        return super().__eq__(other)
+
+
+class IssueAssignee(UserType):
+    login = columns.Text()
+    html_url = columns.Text()
+
+
+class GithubIssue(Model):
```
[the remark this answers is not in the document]
k0machi: Will do

argus/backend/models/runtime_store.py
```diff
@@ -0,0 +1,41 @@
+from datetime import datetime, time
+from cassandra.cqlengine.models import Model
+from cassandra.cqlengine import columns
+
+class RuntimeStore(Model):
```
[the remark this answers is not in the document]
k0machi: Will do

argus/backend/plugins/sct/service.py
```diff
@@ -489,7 +490,10 @@ def get_similar_runs_info(run_ids: list[str]):
                 test_run = SCTTestRun.get(id=run_id)
                 if not test_run:
                     continue
-                issues = ArgusGithubIssue.objects.filter(run_id=run_id).all()
+                links = IssueLink.objects.filter(run_id=run_id).all()
+                issues: list[GithubIssue] = []
+                for batch in chunk(links):
```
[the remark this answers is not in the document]
k0machi: Not really, this is here just in case that happens.

argus/backend/service/github_service.py
```diff
@@ -0,0 +1,232 @@
...
+            last_ran = RuntimeStore.get(key=self.LAST_RAN_KEY)
+        except RuntimeStore.DoesNotExist:
+            last_ran = RuntimeStore()
+            last_ran.key = self.LAST_RAN_KEY
+            last_ran.value = datetime(year=2020, month=1, day=1, hour=0, minute=0, tzinfo=UTC)
+            last_ran.save()
+
+        LOGGER.info("Starting Github Issue sync...")
+
+        all_issues: list[GithubIssue] = list(GithubIssue.all())
```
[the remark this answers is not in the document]
k0machi: Thought about it as well when writing this, yes, I think updating runtimestore might be a way to keep the repos there - do one on install and then add to it if there's a new one on issue submit. Meanwhile I think currently we have around ~3k issues and the query itself is... reasonably fast, about instant right now, but I agree, doing it that way is not very efficient.

argus/backend/service/github_service.py
k0machi: Badly. I contemplate also syncing the github internal issue id to alleviate this problem (also would bulletproof comparing local issue with remote), but I'm not sure what happens on the API side in a situation where you request a repo update and there's an issue that has been moved to an untracked repo.

argus/backend/service/github_service.py
k0machi: Yes, this was me mostly debugging some issues with the sync, I think I will change this to periodic (maybe once every 100 checks) logging at maybe INFO still.

argus/backend/service/github_service.py
k0machi: Ditto.

argus/backend/service/github_service.py
k0machi: True, I've used this one to check if I got exactly the amount of issues expected, not necessarily updated, so this number represents "Issues found"

pyproject.toml
k0machi: Exactly, you can set up page size on the entire client (100 max) and then every query that supports paging will use that size.

argus/backend/service/github_service.py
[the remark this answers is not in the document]
k0machi: Went to check - prod has 7144 entities of which 2144 are unique issues (so about 3 IssueLink per GithubIssue), the query for all 7k using the mapper takes about 800ms-1.8s

argus/backend/models/github_issue.py
```diff
@@ -0,0 +1,60 @@
+from enum import unique
+from uuid import uuid4
+from datetime import datetime
+from cassandra.cqlengine.models import Model
+from cassandra.cqlengine.usertype import UserType
+from cassandra.cqlengine import columns
+
+
+class IssueLabel(UserType):
```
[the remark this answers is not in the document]
soyacz: I see, ok

argus/backend/service/github_service.py
soyacz: badly - means breaking the process, or just not updating?
if issue disappears from the repo, we could add a label `moved`, without pointing where if it's hard to get (anyway, rare situation).

argus/backend/models/github_issue.py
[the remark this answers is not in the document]
k0machi: Done

argus/backend/models/runtime_store.py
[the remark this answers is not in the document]
k0machi: Done

soyacz (approved):
one small remark, LGTM
How do you think to deploy it? Maybe first doing it on staging, doing migration and if all looks ok then switch prod?

k0machi:
> one small remark, LGTM How do you think to deploy it? Maybe first doing it on staging, doing migration and if all looks ok then switch prod?

Yep, that sounds good, I'll pick a window to do that and then deploy on drop along with the cron script

fruch:
@k0machi how's  the deployment of this going ?

k0machi:
> @k0machi how's the deployment of this going ?

Deployment done.

soyacz:
Somehow, Argus works much faster - should be? related?

Issue #641, opened 2025-04-22
Losing discussion content when switching from tab to tab


soyacz:
After adding possibility to create link to specific tab in Run details (https://github.com/scylladb/argus/pull/638),
when user starts to write e.g. comment in discussion tab, switching tabs clears it out.
I often was switching back and forth when writing comment and now it breaks my flow.

PR #644, opened 2025-04-25
fix(testRun): fix disappearing comment content

soyacz:
After recent changes, tab content is recreated when swiching tabs. This causes error of comment content disappear and also reloads events each time events tab is visited (etc.).

This fix makes behavior like before, so tab content is not recreated after visiting it.

fixes: https://github.com/scylladb/argus/issues/641


soyacz:
@k0machi can you take a look at this one - it fixes frustrating bug

k0machi (approved):

PR #646, opened 2025-04-29
fix(highlightWidget): fix update highlight

soyacz:
Highlight update didn't work due to missing body.message update.


juliayakovlev (approved):
suppose it's OK

k0machi (approved):

PR #648, opened 2025-05-05
fix(summaryWidget): remove filtering tests based on product

soyacz:
We used to mix tests for enterprise/OSS versions in a view. When selecting one of it we wanted to exclude tests from the other (seeing only enterprise tests when enterprise version were selected). 
Now it's no longer needed and we started to mix master and enterprise tests in one view causing tests from `scylla-master` release not appearing.

fix by removing this filter.


soyacz:
deployed on staging, shows missing test: https://argus-staging.lab.dbaas.scyop.net/view/performance-regression-tests-weekly-vnodes

PR #651, opened 2025-05-12
feature(argus): Support for AWS services and S3 proxying

### LINKED ISSUE #270: siren-tests: fix the screenshot
Currently the screenshot in case of a test failure in siren-tests job is not displayed.

Cloud qa team has a new aqs account  - Account: sc-dbaas-data-qa (400694452550)
 we can probably use for Argus, just specify which permissions are needed and I can take it with @d-helios.

k0machi:
This commit adds support for interacting with AWS services such as
S3. This means that from now on, Argus will proxy requests made to S3
through itself, proxifying screenshots links to both SCT and Sirenada
and creating pre-signed URLs for logs. This allows settings those
buckets as private.

Task: scylladb/qa-tasks#1874
Fixes #270


fruch (approved):
LGTM

k0machi:
URL Examples:

Log: `http://base_url/api/v1/tests/scylla-cluster-tests/d17288c2-89af-4ee9-8bd4-a04bc30fe6dc/log/longevity-mv-si-4d-2025-1-db-node-d17288c2-4/download`

Screenshot: `http://base_url/api/v1/tests/scylla-cluster-tests/d17288c2-89af-4ee9-8bd4-a04bc30fe6dc/screenshot/grafana-screenshot-overview-20250408_193241-longevity-mv-si-4d-2025-1-monitor-node-d17288c2-1.png`

k0machi:
Okay so after some discussion we're also going to add a generic image proxy route to handle places where it might be tricky to get to the url stored in the database. This route will only allow certain buckets and only certain mime types.

k0machi:
Added `/api/v1/s3/$bucket-name/$bucket-path` route, as well as a new config property to limit which buckets it's going to pull from. It uses libmagic to tell the mime type (AWS' Content-Type header is not reliable - it returns application/octet-stream for image files, which explains why browser doesn't embed them by default).

soyacz (commented):
we store also screenshot links in generic results - need to handle those too
also probably we need to allow more connections in uwsgi config - when user downloads it will take one for longer

fruch:
> we store also screenshot links in generic results - need to handle those too also probably we need to allow more connections in uwsgi config - when user downloads it will take one for longer

everything is redirects... should be fast enough

k0machi:
> we store also screenshot links in generic results - need to handle those too also probably we need to allow more connections in uwsgi config - when user downloads it will take one for longer

I've added a route for that - `/api/v1/s3/$bucket-name/$bucket-path`, since getting to those screenshots or determening what's a screenshot and what's just a text field is a bit tricky, I think it should be handled on the frontend side - check if url is s3 conforming and put the argus s3 proxy there.

PR #655, opened 2025-05-12
improvement(client): Allow GenericResultTable with dynamic parameters

pehala:
### Problem statement

Currently, `GenericResultTable` is highly tied to the Meta class, which contains validation logic for validation rule. This creates a necessity of using Meta class even for cases where name is dynamic, which results in nested classes like this:

```python
def send_perf_simple_query_result_to_argus(argus_client: ArgusClient, result: dict):
    stats = result["stats"]
    workload = result["test_properties"]["type"]
    parameters = result["parameters"]

    class PerfSimpleQueryResult(GenericResultTable):
        class Meta:
            name = f"{workload} - Perf Simple Query"
            description = json.dumps(parameters)
            Columns = [ColumnMetadata(name="allocs_per_op", unit="", type=ResultType.FLOAT, higher_is_better=False),
                       ColumnMetadata(name="cpu_cycles_per_op", unit="", type=ResultType.FLOAT, higher_is_better=False),
                       ColumnMetadata(name="instructions_per_op", unit="",
                                      type=ResultType.FLOAT, higher_is_better=False),
                       ColumnMetadata(name="logallocs_per_op", unit="", type=ResultType.FLOAT, higher_is_better=False),
                       ColumnMetadata(name="mad tps", unit="", type=ResultType.FLOAT, higher_is_better=True),
                       ColumnMetadata(name="max tps", unit="", type=ResultType.FLOAT, higher_is_better=True),
                       ColumnMetadata(name="median tps", unit="", type=ResultType.FLOAT, higher_is_better=True),
                       ColumnMetadata(name="min tps", unit="", type=ResultType.FLOAT, higher_is_better=True),
                       ColumnMetadata(name="tasks_per_op", unit="", type=ResultType.FLOAT, higher_is_better=False),
                       ]

            ValidationRules = {
                "allocs_per_op": ValidationRule(best_pct=5),
                "instructions_per_op": ValidationRule(best_pct=5),
            }
```
which in my opinion, should not be necessary as it is impossible to reuse such class.

### Solution

* Move validation and all arguments into `GenericResultTable`
   * This means validation of validation rules is now only on instantiation of the class, not on class creation, but given how we used it, it should make no difference
* Introduce `StaticGenericResultTable`, which reads the data from `Meta` class as before.
   * Users can use either  `GenericResultTable` or `StaticGenericResultTable`, dependening on the use case.
   * All current usages in SCT will have to be changed to `StaticGenericResultTable` instead. This could be prevented by renaming but that would cause other problems (i.e. we would just switch the part we have to change)
   * Add alternative names in the Meta class (`Columns` -> `columns`, `ValidationRules` -> `validation_rules`)
        * This unifies the naming as some of the variables were using different naming scheme
* Add tests to both verify and showcase the new possibilities

### DIFF
--- a/argus/backend/service/planner_service.py
+++ b/argus/backend/service/planner_service.py
@@ -144,7 +144,7 @@ class PlanningService:
         else:
             plan.view_id = plan_request.view_id
             view = self.update_view_for_plan(plan, existing=True)
-        
+
         plan.save()
 
         return plan
--- a/argus/backend/tests/client_service/test_submit_results.py
+++ b/argus/backend/tests/client_service/test_submit_results.py
@@ -5,7 +5,8 @@ import pytest
 
 from argus.backend.error_handlers import DataValidationError
 from argus.backend.tests.conftest import get_fake_test_run
-from argus.client.generic_result import GenericResultTable, ColumnMetadata, ResultType, ValidationRule, Status
+from argus.client.generic_result import ColumnMetadata, ResultType, ValidationRule, Status, \
+    StaticGenericResultTable
 
 
 @dataclass
@@ -16,7 +17,7 @@ class SampleCell:
     status: Status = Status.UNSET
 
 
-class SampleTable(GenericResultTable):
+class SampleTable(StaticGenericResultTable):
     class Meta:
         name = "Test Table"
         description = "Test Table Description"
--- a/argus/backend/tests/results_service/test_best_results.py
+++ b/argus/backend/tests/results_service/test_best_results.py
@@ -8,12 +8,12 @@ from argus.backend.error_handlers import DataValidationError
 from argus.backend.plugins.sct.testrun import SCTTestRun
 from argus.backend.tests.conftest import get_fake_test_run, fake_test
 from argus.common.enums import TestInvestigationStatus
-from argus.client.generic_result import GenericResultTable, ColumnMetadata, ResultType, ValidationRule, Status
+from argus.client.generic_result import ColumnMetadata, ResultType, ValidationRule, Status, StaticGenericResultTable
 
 LOGGER = logging.getLogger(__name__)
 
 
-class SampleTable(GenericResultTable):
+class SampleTable(StaticGenericResultTable):
     class Meta:
         name = "Test Table Name"
         description = "Test Table Description"
@@ -102,7 +102,7 @@ def test_can_enable_best_results_tracking(fake_test, client_service, results_ser
     best_results = results_service.get_best_results(fake_test.id, results.name)
     assert 'non tracked col name:row' not in best_results  # non tracked column should not be tracked
 
-    class TrackingAllSampleTable(GenericResultTable):
+    class TrackingAllSampleTable(StaticGenericResultTable):
         class Meta:
             name = "Test Table Name"
             description = "Test Table Description"
--- a/argus/backend/tests/results_service/test_validation_rules.py
+++ b/argus/backend/tests/results_service/test_validation_rules.py
@@ -7,12 +7,12 @@ import pytest
 
 from argus.backend.error_handlers import DataValidationError
 from argus.backend.tests.conftest import get_fake_test_run, fake_test
-from argus.client.generic_result import GenericResultTable, ColumnMetadata, ResultType, ValidationRule, Status
+from argus.client.generic_result import ColumnMetadata, ResultType, ValidationRule, Status, StaticGenericResultTable
 
 LOGGER = logging.getLogger(__name__)
 
 
-class SampleTable(GenericResultTable):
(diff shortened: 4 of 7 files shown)


pehala:
Note: I ran black on the whole file, but it introduced some unwanted changes I will revert them, do we run black + pylint regularly on the codebase?

k0machi:
> Note: I ran black on the whole file, but it introduced some unwanted changes I will revert them, do we run black + pylint regularly on the codebase?

Not reguraly at the moment, I run it periodically but haven't done in a while.

pehala:
> Not reguraly at the moment, I run it periodically but haven't done in a while.

If we want these tools to be useful they should be run on each PR (e.g. as a GH action)
Also given they are not used regularly, I would suggest switching to the same toolset as in SCT (i.e. UV + ruff + same rules), but thats your call if you think this is a good idea

k0machi:
> > Not reguraly at the moment, I run it periodically but haven't done in a while.
> 
> If we want these tools to be useful they should be run on each PR (e.g. as a GH action) Also given they are not used regularly, I would suggest switching to the same toolset as in SCT (i.e. UV + ruff + same rules), but thats your call if you think this is a good idea

It's in the plans to do that, yeah. Also to run prettify and eslint for js part.

soyacz:
Things mentioned in 'meta' are for all results in given test and table name pair - if one run changes columns, it will affect all other runs (and graphs). If table name changes, it will generate new graphs for each column. Original intention was to make it deliberate. 
If you need just change the name, class instance `.name` attribute may be changed.
I'm ok with this change if this breaks some valuable use case - @pehala can you share it so I can better understand the need?

pehala:
I would like to replace usages like this
```python
def send_adaptive_timeout_results_to_argus(self, results: dict):
    argus_client = self.test_config.argus_client()
    if not argus_client:
        LOGGER.warning("Will not submit to argus - no client initialized")
        return

    class AdaptiveTimeoutResultsTable(GenericResultTable):
        class Meta:
            name = f"{results['operation']} - Timeout Statistics"
            description = "measurement of specific operation timeouts (ex. decommission, adding nodes etc.)"
            Columns = [
                ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION, higher_is_better=False),
                ColumnMetadata(name="timeout", unit="HH:MM:SS", type=ResultType.DURATION),
                ColumnMetadata(name="end_time", unit="", type=ResultType.TEXT),
            ]

    table = AdaptiveTimeoutResultsTable()
    ...
```
with this
```python
class AdaptiveTimeoutResultsTable(GenericResultTable):
    def __init__(self, operation):
        super().__init__(name=f"{operation} - Dynamic Results")

    class Meta:
        description = "measurement of specific operation timeouts (ex. decommission, adding nodes etc.)"
        Columns = [
            ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION, higher_is_better=False),
            ColumnMetadata(name="timeout", unit="HH:MM:SS", type=ResultType.DURATION),
            ColumnMetadata(name="end_time", unit="", type=ResultType.TEXT),
        ]

def send_adaptive_timeout_results_to_argus(self, results: dict):
    argus_client = self.test_config.argus_client()
    if not argus_client:
        LOGGER.warning("Will not submit to argus - no client initialized")
        return

    table = AdaptiveTimeoutResultsTable(results["operation"])
    ...
```



> Things mentioned in 'meta' are for all results in given test and table name pair - if one run changes columns, it will affect all other runs (and graphs). If table name changes, it will generate new graphs for each column. Original intention was to make it deliberate.

That makes sense and I like the reason, but as can be seen from our current usage we do need some kind of dynamicity in the result reporting and we want to reuse those classes, albeit within the one test, but our current API restricts us so we have to create class dynamically instead (i.e. in the method itself). So we have a static API via class variables but we are bypassing that restriction by creating the class dynamically.

My proposal basically legitimises the idea of some parameters being dynamic within the Result, so it does not bring new use case but makes it so existing use cases are supported properly and we dont have to do workaround to achieve them.

TLDR: If we want to use dynamic names, which we are already using and they do make sense to use, it is better to support this usecase through the API directly.

pehala:
> LGTM,
> But update version of argus-alm in pyproject.toml - so we can publish new client.

Rebased and updated. SCT will require a change after this, I will start incorporating it once this is merged and released (maybe even switch to using released version of argus in the process).

soyacz (approved):

fruch:
> > when updating in SCT, run `python utils/update_argus_client.py` to update Argus Client
> 
> Why are we not using released argus through pypi?

We are using releases, but decide to vendor them in code, and not via hydra docker image.

PR #656, opened 2025-05-12
fix(sut_timestamp): lookup all the possible package names

fruch:
since for `scylla-server` we can have multiple names availble in diffrent stages of the test we should honor those name, so if we report thing before the some of them are reported, it should still work.

regardless the default behavier of `sut_timestamp` should follow the `sut_package_name` if provided, it make much more sane usability-wise.


soyacz (approved):
LGTM

k0machi (approved):

PR #666, opened 2025-05-19
improvement(TestDashboard): Extended filters

### LINKED ISSUE #319: Add "filter by" for tests on the releases dashboard
We need to be able to filter test runs by :
1. Date/s - tests that ran only yesterday or last week or specific dates.
2. Status - passed/failed/aborted or other statuses which should be added.
3. Image id
4. Backend
5. Region

k0machi:
This commit adds new filtering options to the TestDashboard component.
To simplify adding additional filters, a filtering stack object was
introduced to define a filtering operation, its state and its comparison
operation. Each filter state is saved into local storage upon
interaction and restored when reloading the page. Previously available
filters were reworked for the new system.

The filters are:

 * Assignee
 * Status (Improved) - Can now filter out any status, not just not planned
 * Backend (New) - Filters to SCT tests using specified backend
 * ImageId (New) - Filters to SCT tests using specified image

Fixes #319

### DIFF
--- a/argus/backend/controller/api.py
+++ b/argus/backend/controller/api.py
@@ -139,6 +139,19 @@ def release_pytest_results(release_id: str):
     })
 
 
+@bp.route("/release/<string:release_id>/images")
+@api_login_required
+def release_images(release_id: str):
+    release_id = UUID(release_id)
+    service = ArgusService()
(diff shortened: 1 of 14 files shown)


k0machi:
Demo: 
image

soyacz (approved):
LGTM

Issue #668, opened 2025-05-22
Allow multi-line fields


cezarmoise:
Some fields, like `extra_environment_variables` are defined as `text` in the pipeline and can be multi-line.
If possible, Argus should detect this and also have them as multi-line.

Example:
In Argus:
![Image](https://github.com/user-attachments/assets/8b77e339-0574-4f51-a6d3-a427f08b3e36)

In Jenkins:
![Image](https://github.com/user-attachments/assets/76dfda58-60a6-40fb-ab90-53ec616b03ba)

cezarmoise:
I think the solution is simple

```diff
diff --git a/frontend/TestRun/Jenkins/ParameterEditor.svelte b/frontend/TestRun/Jenkins/ParameterEditor.svelte
index b40d6aa..0cde77b 100644
--- a/frontend/TestRun/Jenkins/ParameterEditor.svelte
+++ b/frontend/TestRun/Jenkins/ParameterEditor.svelte
@@ -71,7 +71,11 @@
                 {#if param.value || paramTab == 2}
                     <div class="mb-1">
                         <label class="form-label fw-bold" for={param.name}>{param.name}</label>
-                        <input class="form-control" id={param.name} type="text" bind:value={buildParams[param.name]}>
+                        {#if param.type === "text"}
+                            <textarea class="form-control" id={param.name} bind:value={buildParams[param.name]} rows="4"></textarea>
+                        {:else}
+                            <input class="form-control" id={param.name} type="text" bind:value={buildParams[param.name]}>
+                        {/if}
                         <div id="paramHelp{sanitizeSelector(param.name)}" class="form-text">{@html param.description}</div>
                     </div>
                 {/if}
```

but I'm not sure how to determine if the param should be multiline. ` {#if param.type === "text"}` is a placeholder

PR #670, opened 2025-05-22
feature(TestRun/Jenkins): allow some parameters to be multiline

cezarmoise:
Allow some parameters that are multiline in Jenkins to be multiline in Argus, so editing them is easier without having to add newline chars manually.

closes: #668


cezarmoise:
Change not tested (yet).

k0machi:
It would be better that instead of hardcoding them you change `jenkins_service.py` a bit so that it provides `_class` field from the Jenkins API to the frontend, then you can use the type as provided by the API:

image

k0machi:
Corrected slightly, plus added `TextParam.svelte` for future use in a wizard

PR #672, opened 2025-05-22
feature(frontend): Pytest Overview Widget

k0machi:
This commit adds a new View Widget, intended to provide overview into
specific test's pytest results, such as statistics, charts and a
searchable result list. The new widget supports both views and release
pages, supporting both always open and collapsed options.

Widget features:

* Filter tests by cut off date or by range of dates
* Filter by test status
* Filter by test marker
* Filter by test user specified field presence
* Filter by text within the test

Fixes scylladb/qa-tasks#273

### DIFF
--- a/argus/backend/controller/view_api.py
+++ b/argus/backend/controller/view_api.py
@@ -10,6 +10,7 @@ from argus.backend.controller.views_widgets.summary import bp as summary_bp
 from argus.backend.controller.views_widgets.graphs import bp as graphs_bp
 from argus.backend.controller.views_widgets.nemesis_stats import bp as nemesis_stats_bp
 from argus.backend.controller.views_widgets.graphed_stats import bp as graphed_stats_bp
+from argus.backend.controller.views_widgets.pytest import bp as pytest_bp
 from argus.backend.error_handlers import handle_api_exception
 from argus.backend.models.web import User
 from argus.backend.service.stats import ViewStatsCollector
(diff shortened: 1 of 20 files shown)


soyacz:
@k0machi can you give this PR a bit of description?

k0machi:
> @k0machi can you give this PR a bit of description?

Yes, I'll add some extra things today and give it a proper message

k0machi:
Demo:

https://github.com/user-attachments/assets/47fcbe3f-8233-4951-9caf-b3b25cfc9981

k0machi:
Updated and deployed on staging @fruch @soyacz

soyacz:
> Updated and deployed on staging @fruch @soyacz

I went to staging into dtest view and it took > 60 seconds and shown nothing (error 504).
Also we agreed to filter out passed tests (and skipped) from defaults - still turned on.

k0machi:
@soyacz Redeployed to staging

k0machi:
Made graphs use buckets instead of using each result's session timestamp, now it is actually readable (and shows empty dates):

image

k0machi:
@soyacz ready for review I think, deployed to staging with partial migration of current data, changes since last time:

* Fixed small issues with test name filtering
* Added migration script
* Added lazy fetch for user fields - now they are only fetched when user wants to build a graph while filtering by filters
* Fixed message display

soyacz:
@k0machi is it better/safe to merge and then migrate? (so we don't lose data during migration)

k0machi:
@soyacz Do you think we are ready to merge? Can you add your review?

soyacz:
> @soyacz Do you think we are ready to merge? Can you add your review?

We didn't yet upgrade cluster - I don't know the reason.
But I think we can merge - it will work better anyway with this code.

PR #676, opened 2025-05-28
replace poetry with uv as the packaging tool

fruch:
* remove poetry completly
* use setuptools_scm
* switch the release to use trusted publishers

Ref: https://docs.pypi.org/trusted-publishers


k0machi:
Missing _version.py file I think

fruch:
> Missing _version.py file I think

it's a generated file by setuptools_scm

fruch:
> > Missing _version.py file I think
> 
> it's a generated file by setuptools_scm

it needed only on build time

.github/workflows/release.yml
```diff
@@ -3,24 +3,59 @@ name: Release on Pypi and GitHub
...
-
-      - name: Build and publish to pypi
-        uses: JRubics/poetry-publish@v2.1
-        with:
-          pypi_token: ${{ secrets.PYPI_TOKEN }}
-
-      - name: Create GitHub release
-        uses: softprops/action-gh-release@v1
-        with:
-          tag_name: v${{ env.VERSION }}
```
k0machi, line 25: why was this removed?

.github/workflows/release.yml
```diff
@@ -3,24 +3,59 @@ name: Release on Pypi and GitHub
...
-      - uses: actions/checkout@v3
-
-      - name: Retrieve version from pyproject.toml
-        id: version
-        run: echo "VERSION=$(grep -m 1 version pyproject.toml | tr -s ' ' | tr -d '"' | tr -d "'" | cut -d' ' -f3)" >> $GITHUB_ENV
-
-      - name: Build and publish to pypi
-        uses: JRubics/poetry-publish@v2.1
-        with:
-          pypi_token: ${{ secrets.PYPI_TOKEN }}
```
k0machi, line 20: Token is now missing - is the new action always assuming that `PYPI_TOKEN` is the secret name?

.github/workflows/release.yml
```diff
@@ -3,24 +3,59 @@ name: Release on Pypi and GitHub
...
-
-      - name: Create GitHub release
-        uses: softprops/action-gh-release@v1
-        with:
-          tag_name: v${{ env.VERSION }}
-          generate_release_notes: true
+    - name: Download all the dists
```
k0machi, line 47: Why is this step needed?

pyproject.toml
```diff
@@ -1,63 +1,79 @@
...
-ruff = "^0.11.9"
-autopep8 = "^2.3.2"
+[project.scripts]
+argus-client-generic = 'argus.client.generic.cli:cli'
+argus-driver-matrix-client = 'argus.client.driver_matrix_tests.cli:cli'
 
-[build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+[project.optional-dependencies]
```
k0machi, line 44: Any option in UV to keep using semver? AFAIK `>=` for, example, `3.10.0` will pull `4.0.0` as a valid installation compared to `^3.10.0` which will always be limited to `3.x.y`

k0machi (changes requested):

.github/workflows/release.yml
fruch: Since we are switching to make base on tags only

No need to specify it anymore

fruch:
@k0machi @soyacz now the CI also works

once last thing would to be to test the release CI

.github/workflows/release.yml
```diff
@@ -3,24 +3,57 @@ name: Release on Pypi and GitHub
 on:
   workflow_dispatch:
 
+  push:
+    tags:
+      - 'v*'  # Trigger on version tags like v1.0.0
```
pehala, line 8: Is trhe name of the tag used anywhere?

pehala, line 8: Original workflow assumed you are gonna initiate this workflow manually and as part of it both tag and GH release is created. What is expected release workflow now?

k0machi:
> @k0machi @soyacz now the CI also works
> 
> once last thing would to be to test the release CI

CI for tests seems to be catching a module outside of `tests` directory, `TestLookup` is not a test, it's a class for searching tests in Argus.

.github/workflows/release.yml
fruch: Yes it's used by setuptools_scm

fruch: The workflow is to tag a version on master branch, and this flow would be triggered to do the release.

Version of the tag is what would be used to publish

Hence you can release with commiting any extra code related to versioning

pyproject.toml
k0machi: Should be safe to update

.github/workflows/release.yml
fruch: yes, the git tag would be the source of truth

pyproject.toml
```diff
@@ -1,62 +1,75 @@
...
-python-jenkins = "^1.7.0"
-python-slugify = "^6.1.1"
-pygithub = "^2.6.1"
-boto3 = "^1.38.9"
-
-[tool.poetry.group.docker-image.dependencies]
-supervisor = "^4.2.4"
-
-[tool.poetry.group.dev.dependencies]
-boto3-stubs = {extras = ["s3"], version = "^1.38.9"}
```
k0machi, line 49: One thing is that this is missing from dev group on the new manifest.

k0machi (changes requested):
Just one thing, otherwise LGTM

PR #691, opened 2025-06-05
test(client): add unittest to test the client cli

fruch:
this test is using request-mock package to simulate the argus backend


fruch:
> shouldn't assert at least mock was called?

It could be extended to do that
The reason I've picked it first was problem in how the command line options were handled 

Plus some issue I've seen about how we should mock the server in client test

k0machi (approved):

Issue #690, opened 2025-06-05
Have a way to hide some reports from SCT result tab


fruch:
While working on adaptive timeouts reporting, 

we seen that in some cases we might not want to show all of the data (while still saving the data),

in some cases we would want to not show the adaptive timeouts (while still saving the data), 
cause it would clutter the performance reports

soyacz:
Is it something that should be controlled by client (e.g. using a flag) or something to be hidden through Argus UI? (or both?)

soyacz:
ok, for columns hiding - I'll add  this feature to client&Argus
for not showing at all - let's not send it for specific test cases (like performance)

PR #692, opened 2025-06-06
improvement(results): allow to send hidden columns

soyacz:
Sometimes we might send data to argus but not display it in Argus UI. E.g. for adaptive timeouts feature, where we collect metrics data for further analysis out of test scope.

Hide columns by defining visibility in column metadata set by client.

fixes: https://github.com/scylladb/argus/issues/690


k0machi (approved):

fruch (approved):
LGTM

PR #693, opened 2025-06-09
improvement(TestRunInfo): Add text-break to Test method

pehala:
Test methods name can be long, overflowing into other half of the table and overshadowing text there
Example:
https://argus.scylladb.com/tests/scylla-cluster-tests/6624c84b-65a9-4086-8286-312911feeb69
obrazek


k0machi (approved):

fruch (approved):
LGTM

PR #700, opened 2025-06-12
fix(generic_result): Add missing init params to StaticGenericResultTable

pehala:
This makes it on par with GenericResultTable in terms of init parameters


fruch:
what's going on ? and why it was missing in the PR with 0.15.2 ?

pehala:
Added the test

fruch (approved):
LGTM

PR #714, opened 2025-06-25
feature(AdminPanel): Utilities menu for Administrators

k0machi:
This commit adds a new panel to the Admin Menu, allowing Administrators
to execute various utility scripts. Currently only script implemented is
"Reload Config", allowing Administrator to reload application config
without restarting the application.

### DIFF
--- a/argus/backend/controller/admin_api.py
+++ b/argus/backend/controller/admin_api.py
@@ -6,6 +6,7 @@ from flask import (
     Request,
 )
 from argus.backend.error_handlers import handle_api_exception
+from argus.backend.service.admin import AdminService
 from argus.backend.service.release_manager import ReleaseEditPayload, ReleaseManagerService
 from argus.backend.service.user import UserService, api_login_required, check_roles
 from argus.backend.models.web import User, UserRoles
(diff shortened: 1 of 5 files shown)


soyacz:
to change config, you need anyway to ssh to argus host - what having it in admin panel gives?

k0machi:
> to change config, you need anyway to ssh to argus host - what having it in admin panel gives?

This is an initial implementation for this page, I plan to add a config editor later as well. The current benefit is not having to restart the app and interrupt what it could be doing in the background, even if you still have to ssh to the host.

fruch:
> > to change config, you need anyway to ssh to argus host - what having it in admin panel gives?
> 
> This is an initial implementation for this page, I plan to add a config editor later as well. The current benefit is not having to restart the app and interrupt what it could be doing in the background, even if you still have to ssh to the host.

How many times a week you need to change configuration? How many times it not while you need to restart Argus anyhow cause of code change ?

Im not sure if we need such an admin UI, for those rare changes.

I think it's much more important to have a clear process and machineary to how and where the production configuration is saved

Issue #709, opened 2025-06-18
Differentiating Test Statuses in Performance View Dashboards


juliayakovlev:
Performance view dashboards.

Currently, the performance view dashboards use the same icon ![Image](https://github.com/user-attachments/assets/afb55c48-405d-4782-af5c-b49c308a1ec2)  to indicate two distinct test statuses:

1. when a test was not started, users must expand the relevant section (div) to understand why. This extra step hinders quick comprehension.
![Image](https://github.com/user-attachments/assets/5e7f8079-b80f-4b47-aea9-b32494ec5b3a)

2. test with a "TEST-ERROR" status (may be in other cases also?).
![Image](https://github.com/user-attachments/assets/03c3119e-befa-4f13-bd75-054768d20444)

To avoid confusion and provide clearer visual cues, I propose using a different icon for each of these statuses. This differentiation will allow users to quickly understand if a test did not complete successfully.

PR #720, opened 2025-07-07
improvement(summary_widget): distinguish test statuses

soyacz:
In summary view widget test status was either failed, passed or other so user needed to go into the test to see actual status (and sometimes it was aborted or running etc.).

Show different icons for each test status so user quickly get's idea what's the actual status without getting into test run details.

fixes: https://github.com/scylladb/argus/issues/709

### DIFF
--- a/frontend/Views/Widgets/SummaryWidget/SummaryWidget.svelte
+++ b/frontend/Views/Widgets/SummaryWidget/SummaryWidget.svelte
@@ -8,8 +8,13 @@
         faChevronDown,
         faChevronUp,
         faExclamationTriangle,
+        faExclamationCircle,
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

fruch:
@soyacz you have an example picture of how it look now ? (maybe in the example @juliayakovlev gave)

soyacz:
> @soyacz you have an example picture of how it look now ? (maybe in the example @juliayakovlev gave)

I'd need to deploy it to staging... I would test it on prod (and change if still something is not looking right)

soyacz:
example with aborted test:
image

Issue #721, opened 2025-07-08
Show all people that are not participating in the testplan in the error


pehala:
Currently, you are shown only one person, even when multiple people are not assigned.
`User @k0machi is participating but is not assigned.`

soyacz:
any errors in the console?
How can I reproduce it?

PR #722, opened 2025-07-08
fix(ReleasePlanCreator): show all missing participants

soyacz:
When several participants are not assigned to tests, Argus shows error message only about one.

Fix by collecting list of unassigned participants and showing all of them in the message.

closes: https://github.com/scylladb/argus/issues/721


pehala (approved):

fruch (approved):
LGTM

PR #726, opened 2025-07-10
feature(similar_events): hide duplicates

soyacz:
When user looks at events tab, often there are duplicates, blurring the image and it's harder to spot distinct events.

Fix by using similar events engine to detect duplicates within the same run and hide them from user (with ability to show).

How it works:
1. similar event engine detects duplicate events in the same run
2. save them along with similar events (new `duplicates_list` column)
3. Argus UI fetches duplicates along with similar list
4. Argus constructs 'distinct_events' (ones that are not duplicates - meaning being a first occurence of this event) and next events that are duplicates of it, are appended to a list of duplicates
5. When user clicks 'Show duplicates', duplicates of selected event are shown with light blue background color
6. User may then hide duplicates by clicking again the same button (it switches to 'Hide duplicates' button)

closes: https://github.com/scylladb/scylla-cluster-tests/issues/11334

### DIFF
--- a/argus/backend/models/argus_ai.py
+++ b/argus/backend/models/argus_ai.py
@@ -11,6 +11,8 @@ class ErrorEventEmbeddings(Model):
     columns.BaseCollectionColumn._freeze_db_type(embedding)
     similars_map = columns.Map(key_type=columns.UUID(), value_type=columns.Integer())
     columns.BaseCollectionColumn._freeze_db_type(similars_map)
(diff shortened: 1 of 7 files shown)


soyacz:
@k0machi I'll merge it without your approval, please do review when you have some time and I'll fix it in the followup (I wanted to recalculate all embeddings and similarities over the weekend)

PR #732, opened 2025-07-21
fix(GithubIssue): Display correct user for each page

### LINKED ISSUE #697: job issues tab show the wrong users
issues in the issue list of a job, show the wrong users:

Image

one of them was attached by me, and one was attached by someone else,
in the activity log we can see the correct user

https://argus.scylladb.com/tests/scylla-cluster-tests/498f9311-4f67-49d7-b3e1-bccd563df3db/issues

I'm guess since the refactor of the issues table, we are pulling the information of the first one who attached this issue
and not the correct one anymore.

k0machi:
This commit fixes an issue where user displayed on issue overview would
show first user who submitted the issue ever, instead of whichever user
submitted this issue to this specific view. For example, previously, an
issue added on scylla-staging by user1 would be displayed as added by
user1 on scylla-master, even though it was added by user2 to that
release. Previous issues are not migrated but the new ones will use
correct behaviour.

Fixes #697

### DIFF
--- a/argus/backend/models/github_issue.py
+++ b/argus/backend/models/github_issue.py
@@ -58,3 +58,5 @@ class IssueLink(Model):
     release_id = columns.UUID(index=True)
     group_id = columns.UUID(index=True)
     test_id = columns.UUID(index=True)
+    user_id = columns.UUID(index=True)
+    added_on = columns.DateTime(default=datetime.utcnow)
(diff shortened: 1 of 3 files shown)


k0machi:
> Frankly, I bit don't understand this approach: why on the view it cannot simply use `user_id` from issue_link?

I don't understand the question - previously we would store no information as to who added the link (not the issue), which leads to problems in places where a specific view (release, view, test) would show the person who FIRST added the issue, not who linked it to said view (release, view, test). This is what this PR tries to address.

PR #734, opened 2025-07-21
improvement(PytestRun.svelte): Pytest View improvements 

### LINKED ISSUE #728: DTest/Pytest View improvements
There are currently several issues with individual run dtest overview:

- [ ] Filters should default to just Failed and Error related statuses
- [ ] Filters should be shown as grid
- [ ] Individual test views should look similar to driver matrix view.
- [ ] distinct colors for different statuses

Tracking as part of #718

k0machi:
This commit brings some refactoring towards PytestRun items.

* Default filters have been changed to only show bad statuses.
* Added focus check to not refresh data when tab is not in focus
* Reworked the accordion stack into rows similar to overview widget with
  some fields rearranged
* Added lazy fetch for user fields
* Fixed status colors

Fixes #728

### DIFF
--- a/frontend/TestRun/Pytest/PytestItem.svelte
+++ b/frontend/TestRun/Pytest/PytestItem.svelte
@@ -1,80 +1,103 @@
 <script lang="ts">
+    import Fa from "svelte-fa";
+    import { PytestBgStyles, PytestTextStyles } from "../../Common/TestStatus";
     import {type PytestData, PytestStatus} from "./types";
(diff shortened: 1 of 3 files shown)


k0machi:
Depends on #672

argus/_version.py
soyacz, line 1: @k0machi please don't commit this file - isn't it in gitignore already?

soyacz (approved):
LGTM
Just remove _version file
Merge & migrate data

argus/_version.py
k0machi: It should be - it keeps coming back and it's actually part of parent PR, this PR is clear of it.  I'll check gitignore

Issue #731, opened 2025-07-17
Argus does not show events when similar events table is empty for given run


soyacz:
as in the title, empty similar events response makes events tab empty

PR #736, opened 2025-07-21
Fix: prevent crash when similar_events is empty on Events tab

soyacz:
- Ensure distinctEvents and shownDuplicates are initialized for all severities (ERROR, CRITICAL, WARNING, NORMAL, DEBUG)
- Add null checks and simplify display logic to avoid accessing undefined properties
- Now WARNING, NORMAL, and DEBUG events are always shown when enabled, regardless of similar_events response
- Prevents "Cannot read properties of undefined (reading 'hasOwnProperty')" error

fixes: https://github.com/scylladb/argus/issues/731

### DIFF
--- a/frontend/TestRun/EventsTab.svelte
+++ b/frontend/TestRun/EventsTab.svelte
@@ -14,8 +14,14 @@
     let nemesisPeriods = {};
     let eventEmbeddings = {};
     let eventDuplicates = {};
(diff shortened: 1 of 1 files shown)


soyacz:
@k0machi I'm merging this to fix prod issue. Will fix comments in followup

PR #741, opened 2025-07-23
chore(project): Update dependencies

### LINKED ISSUE #740: update python driver to latest
it's been a while we didn't update scylla-driver - we should do it before upgrading db instances.

k0machi:
Fixes #740

### DIFF
--- a/argusAI/uv.lock
+++ b/argusAI/uv.lock
@@ -1,6 +1,10 @@
 version = 1
 revision = 2
 requires-python = ">=3.12"
+resolution-markers = [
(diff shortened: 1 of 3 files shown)


soyacz:
can you also update packages in argusAI dir?

k0machi:
> can you also update packages in argusAI dir?

Done

k0machi:
> @k0machi will you take care of deploying? As to similar events - you need to deploy to staging (~/argusAI dir) and restart similar events service (you'll find it in bash history). Then please confirm with journalctl if all works.

Done, deployed on both

PR #755, opened 2025-07-30
fix(argus_service): Hide disabled jobs when requesting planned jobs

k0machi:


soyacz:
was there an issue for this, or you just spotted something and fixed?

PR #757, opened 2025-07-31
improvement(stats.py): Relax version filter to match from the start

### LINKED ISSUE #752: Some tests are shown in version filter as not run, even though they did run
Example https://argus.scylladb.com/view/scylla-2025-2-2025-2-2025-2-test-plan?productVersion=2025.2.1 with 2025.2.1 version filter (link itself does not select the filter due to https://github.com/scylladb/argus/issues/751)

The test `longevity/longevity-10gb-3h-gce-test` is shown as not run even though https://argus.scylladb.com/tests/scylla-cluster-tests/6264b63c-fdcf-47c7-949b-8862ec6e1733 this run exists and did use correct version (2025.2.1-20250716.7bb43d812e1c)

Image

k0machi:
This commit changes behaviour of the scylla version filter within
TestDashboard widget. Previously, scyllaVersion query param would result
in a strict a == b comparison, which could exclude runs from appearing
in the filtered dashboard. This also allows us to fold long versions
based on regex rule to match both short version strings and long version
strings.

Fixes #752

### DIFF
--- a/argus/backend/service/stats.py
+++ b/argus/backend/service/stats.py
@@ -153,6 +153,14 @@ def _get_image(row: dict):
         return cs.db_node.image_id
 
 
+def check_version(filter_string: str, version: str) -> bool:
+    if not version:
+        return False
+    if version.startswith(filter_string):
+        return True
+
+    return False
+
 def _fetch_multiple_release_queries(entity: Model, releases: list[str]):
     result_set = []
     for release_id in releases:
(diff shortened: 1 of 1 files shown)


argus/backend/service/stats.py
```diff
@@ -542,11 +551,11 @@ def collect(self, limited=False, force=False, include_no_version=False, image_id
             }
         if self.release_version:
             if include_no_version:
-                expr = lambda row: row["scylla_version"] == self.release_version or not row["scylla_version"]
+                expr = lambda row: check_version(self.release_version, row["scylla_version"]) or not row["scylla_version"]
```
soyacz, line 554: why not just `row.get("scylla_version", "").startswith(self.release_version)`?

soyacz (approved):
LGTM

PR #759, opened 2025-07-31
fix(generic/model.py): Properly handle promote-release version

k0machi:
Since promote-release job reports versions with a dash instead of a
tilde, this causes short version to become just "rc1", which is not very
helpful. This commit adds dash as a valid character for short version.


soyacz (approved):
LGTM

PR #762, opened 2025-08-01
improvement(TestDashboard.svelte): Fold long versions into short ones

### LINKED ISSUE #760: Fold Version widget inside TestDashboard
Currently, some of the views that use TestDashboard are very crowded and unpleasant to use due to being bloated with different versions:

Image

### Proposal

We should instead use the same regexp we use to create a short version string and only allow this filter to display those - folding remaining versions into this one. As #757 is merged, we will not require such precise version filtering as we have currently and this would allow us to structure widget like this:
```
[All] [No version] [1.2.3] [2025.1.1] [unrecognized]
```

If the regexp fails - those versions will be put at the end as "unrecognized ones", to deprioritize them. Otherwise, selecting `1.2.3` for example would select `1.2.3-abcde`, `1.2.3-bcdef`, etc.

k0machi:
As #757 relaxes version requirements, we no longer need to strictly
compare versions to get a result set of them, allowing less versions to
be displayed on TestDashboard widget.

Fixes #760


k0machi:
Before:
image

After:
image

soyacz:
can you make 2025 be before 2023?

k0machi:
Done:

image

Issue #618, opened 2025-03-25
Include instance type that was used in the Resources page


pehala:
It would be a great help if I could look into the Resource page and understand quickly what instances were used for this test

pehala:
I would like to know what instance types (i.e. i4i.large etc.) was used for that node/loader.

PR #763, opened 2025-08-01
feature(plugins/sct): Add instance_type to CloudInstanceDetails

k0machi:
This commit adds a new field to the CloudInstanceDetails UDT, which
allows specifying individual resource instance type (as opposed to
providing config-specified instance type) and adds necessary code to
show it in the resource table on SCT run.

Task: #618

### DIFF
--- a/frontend/TestRun/ResourcesInfo.svelte
+++ b/frontend/TestRun/ResourcesInfo.svelte
@@ -20,6 +20,7 @@
         provider: ["instance_info", "provider"],
         state: "state",
         name: "name",
+        type: ["instance_info", "instance_type"],
         shards: ["instance_info", "shards_amount"],
         publicIp: ["instance_info", "public_ip"],
         privateIp: ["instance_info", "private_ip"],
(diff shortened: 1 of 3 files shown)


frontend/TestRun/ResourcesInfo.svelte
```diff
@@ -350,6 +370,7 @@
         {#each sortResourcesByKey(resources, sortHeader, sortAscending) as resource (resource.name)}
             <tr class:d-none={filterResource(resource)}>
                 <td>{resource.instance_info.provider}</td>
+                <td>{resource.instance_info.instance_type}</td>
```
soyacz, line 373: this will be None for existing tests (sent with old client) - how will it look? shouldn't we fallback to `N/A`?

soyacz (approved):
Double checking (I think it should work) - old clients should not fail when posting to `/<string:run_id>/resource/create` - do you agree?

Otherwise, small remark about None value display in Argus.

frontend/TestRun/ResourcesInfo.svelte
k0machi: Fixed

PR #764, opened 2025-08-01
feature(EventsTab.svelte): Add quick issue attach to Similar Events

### LINKED ISSUE #708: Improving issue reporting in Argus
To simplify the process of reporting issues for tests within Argus, we propose adding a "Report in the test" icon next to the issue in two key locations:

- "All issues for this test" list: Found under the "Issue" tab within the job run.

Image

- "Similar Events" list:

Image

**The aim is to enable users to associate an issue with a test run with a single click.**

k0machi:
This commit adds an ability to quickly attach issues to runs when
viewing the "Similar Events" tab.

Fixes #708


k0machi:
Demo: 
image

PR #768, opened 2025-08-04
fix(release-page.js): Disable shorthand stats

k0machi:
This commit disables stat bars found on the "Release" overview page, as
they can be quite costly to run for not much benefit


soyacz (approved):
please add line of comment why

PR #780, opened 2025-09-03
fix(SCTParameterWizard): Add fuzzy ami re searching

### LINKED ISSUE #775: AMI name is not allowed in `Rebuild` wizard
It's not allowed to use AMI name in `Rebuild wizard`, `Scylla AMI` parameter, only AMI id.
I think we need to allow that.
The new added performance trigger will start the performance jobs with AMI name for `master:latest`. With this limitation we can't use re-build feature

https://argus.scylladb.com/tests/scylla-cluster-tests/4fc49dc3-090b-4afc-8411-786dc204ed5a

Image

k0machi:
Jenkins pipeline allows searching for AMIs using scylla build tag,
this commit adds this to the SCTParameterWizard.

Fixes #775

### DIFF
--- a/frontend/TestRun/Jenkins/SCTParameterWizard.svelte
+++ b/frontend/TestRun/Jenkins/SCTParameterWizard.svelte
@@ -266,7 +266,8 @@
                     validationHelpText: "AMI Id should be in the format of \"ami-hexid\"",
                     validate: function (params) {
                         const AMI_RE = /^ami-([abcdef\d]{4,17})$/;
-                        return !!params[this.internalName] && AMI_RE.test(params[this.internalName]);
+                        const AMI_FUZZY_RE = /^scylla-(.+)/;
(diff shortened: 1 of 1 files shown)


frontend/TestRun/Jenkins/SCTParameterWizard.svelte
```diff
@@ -266,7 +266,8 @@
                     validationHelpText: "AMI Id should be in the format of \"ami-hexid\"",
                     validate: function (params) {
                         const AMI_RE = /^ami-([abcdef\d]{4,17})$/;
-                        return !!params[this.internalName] && AMI_RE.test(params[this.internalName]);
+                        const AMI_FUZZY_RE = /^scylla-\d{4}\.\d\.\d(-dev)?(-[\w\d_]*)(-\d{4}-\d{2}-\d{2})T(\d{2}-?){3}/;
```
soyacz, line 269: I think we could be less strict and test only if it starts with scylla-\d{4}
Better not to add this maintenance burden when someone changes this pattern

PR #784, opened 2025-09-10
improvement(TestDashboardTest): Add triage information to test

### LINKED ISSUE #597: UI - display no of passed/skipped/failed Nemesis runs
This is what I see today - pass or fail per run:

Image

It'd be much more useful if I get a clearer, from the overview, picture - how many passed/failed/skipped. It'll allow me to weigh in what the importance of investigating a run.

Moreover - if I have the 1st Nemesis run, that would be even more helpful - if I see it's the same Nemesis, I may decide to de-prioritize the investigation.

k0machi:
This commit adds extra information to test blocks on a TestDashboard,
namely:

* The amount of passed/failed nemesis is now reflected for SCT runs
* The time taken is displayed under the test

Fixes #597

### DIFF
--- a/argus/backend/plugins/sct/testrun.py
+++ b/argus/backend/plugins/sct/testrun.py
@@ -117,7 +117,7 @@ class SCTTestRun(PluginModelBase):
     @classmethod
     def _stats_query(cls) -> str:
         return ("SELECT id, test_id, group_id, release_id, status, start_time, build_job_url, build_id, "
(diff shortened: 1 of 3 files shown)


soyacz:
@mykaul was it meant to be as part of 'investigate selected'` or this one is good enough?

mykaul:
> @mykaul was it meant to be as part of 'investigate selected'` or this one is good enough?

This looks good enough - it allows me to pick what to investigate based on run time.

PR #803, opened 2025-09-19
test: add sct client endpoint tests

soyacz:
Also reduced tests flakiness and some error handling to have clearer output without errors.

Also adjusted code by removing deprecated utcnow method and other small fixes to reduce number of warnings in tests.

### DIFF
--- a/argus/backend/tests/results_service/test_results_service.py
+++ b/argus/backend/tests/results_service/test_results_service.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+from time import sleep
 from uuid import uuid4
 
 import pytest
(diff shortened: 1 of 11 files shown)


argus/backend/tests/results_service/test_results_service.py
```diff
@@ -123,6 +124,7 @@ def test_get_tests_by_version_groups_runs_correctly(argus_db):
         investigation_status='',
         packages=[pkg_v4_0]
     ).save()
+    sleep(0.001)  # Ensure different start_time
```
k0machi, line 127: Instead of adding sleep, why not just increment start_time manually?

k0machi (changes requested):
Some minor things, otherwise LGTM

k0machi (approved):

Issue #769, opened 2025-08-04
Add a "download all" button to logs tab


radoslawcybulski:
When viewing logs tab in Argus i'm offered with download button for each entry (db-cluster, sct-runner-events and so on), usually 6 or 7. Since more often than not i'm not 100% sure where logs are for things i'm interested in i just download them all and unpack into some directory alltogether. I assume others do similar things, so i propose:
- add 'download all' button, that will download everything (i'm not sure page allows you to download a directory, so i guess you'd have to tar files together with no compression or something similar)
- or rest api, so i could get list of files and each file based by Argus workspace and run number

soyacz:
I think over time you'll get the idea where specific logs are - download all would be needed only temporarily.
Clicking couple log links is also not that problematic and benefit of such button is minimal.
Closing for now.

radoslawcybulski:
I disagree - i don't want to get the idea where specific logs are. And i don't think anyone else wants to either - this is just a band aid. Previously i made a script that auto downloaded all possible combinations of logs based on hash (i just tried wget all possible links and ignored errors), then unpack them and search them all. I don't know, where my logs are, but i'm sure i won't miss anything.

I understand you don't want to add additional work, but maybe at least some rest api? The server knows all files for given run, so maybe api in form <addr/argus_id/run_index/file_index>, while file_index goes from 0 and you reply with nth log file or 404 if index is too large? That would solve my need and should not take more than 10 minutes of coding.

soyacz:
I don't understand what more you need to know to download proper log file? db logs are in `db-cluster....zst` (or if node was decommissioned, then specific log is listed there), sct events (and all smaller additional logs) are in `sct-runner-events....zst`, sct logs (test logic) in `sct....zst` etc.
why do you need to download e.g. db log if you track test flow error or downloading sct log if you want to see what happened to specific db node?

Please explain the use case so I can reconsider this one.

PR #819, opened 2025-10-10
feature(artifacts): download all artifacts button

soyacz:
Added "Download all" button on artifact page.
Clicking it triggers download of all artifacts one by one with slight delay of 800ms to avoid cancelling by web browser.

closes: https://github.com/scylladb/argus/issues/769

image

### DIFF
--- a/frontend/TestRun/ArtifactTab.svelte
+++ b/frontend/TestRun/ArtifactTab.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-    import { faPlus } from "@fortawesome/free-solid-svg-icons";
+    import { faPlus, faDownload } from "@fortawesome/free-solid-svg-icons";
     import ArtifactRow from "./ArtifactRow.svelte";
     import { sendMessage } from "../Stores/AlertStore";
     import { createEventDispatcher } from "svelte";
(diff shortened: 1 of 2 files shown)


frontend/TestRun/ArtifactTab.svelte
```diff
@@ -10,6 +10,30 @@
...
+            const downloadUrl = getArtifactDownloadLink(logName);
+            const linkElement = document.createElement("a");
+            linkElement.href = downloadUrl;
+            linkElement.setAttribute("download", "");
+            linkElement.style.display = "none";
+            document.body.appendChild(linkElement);
+            linkElement.click();
+            document.body.removeChild(linkElement);
+
+            await delay(DOWNLOAD_DELAY_MS);
```
k0machi: Could dispatch `setTimeout` with increasing timeout for each link instead:

```js
let delay = 0;
for ... {
   setTimeout(click, delay);
   delay += DOWNLOAD_DELAY_MS;
}
```
In any case it would be good to block download buttton until all delays were fired, whether with "blocking" loop like this or stored promises - just add a final callback that would unblock the button and lock it at the start of the function.

soyacz:
@k0machi thanks for review. I've adjusted, can you check again?

k0machi (approved):

PR #825, opened 2025-10-20
sct events frontend rework

k0machi:
- **fix(TestRun.svelte): Migrate to svelte5**
- **feature(SCTTestRun): Real-time event component**
- **chore: update ts lang mode and minor deps**

### DIFF
--- a/frontend/TestRun/SCT/SctEvent.svelte
+++ b/frontend/TestRun/SCT/SctEvent.svelte
@@ -4,17 +4,20 @@
     import type { EventSeverityFilter, Options, SCTEvent } from "./SctEvents.svelte";
     import { faNode } from "@fortawesome/free-brands-svg-icons";
     import Fa from "svelte-fa";
-    import { faClipboard, faClock, faServer, faSpider } from "@fortawesome/free-solid-svg-icons";
+    import { faClipboard, faClock, faCopy, faServer, faSpider } from "@fortawesome/free-solid-svg-icons";
 
     interface Props {
         event: SCTEvent,
         filterState: EventSeverityFilter,
         options: Options,
+        filterString: string,
     }
-    let { event, filterState, options }: Props = $props();
+    let { event, filterState, options, filterString = $bindable() }: Props = $props();
+    const MESSAGE_CUTOFF = 600;
+    const SHORT_MESSAGE_LEN = 500;
 
     const sliceMessage = function (message: string): string {
-        return fullMessage ? message : message.slice(0, Math.max(Math.ceil(message.length * 0.15), 500));
+        return fullMessage && message.length <= MESSAGE_CUTOFF ? message : message.slice(0, SHORT_MESSAGE_LEN);
     }
 
     const messageRegex = /\((?<eventType>\w+) Severity\.(?<severity>[A-Z]+)\) (?<rawFields>.+)\n?/s;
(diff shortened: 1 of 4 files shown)


frontend/TestRun/SCT/SctEvent.svelte
```diff
@@ -0,0 +1,194 @@
...
+        parsedMessage.fields = parsedFields;
+        parsedMessage.logLine = logLine.trim();
+        parsedMessage.parsed = true;
+        parsedMessage.message = message;
+
+        return parsedMessage;
+    };
+
+    let fullMessage = $state(false);
+    //let parsedMessage = $derived(parseEventMessage(event.message));
```
soyacz: leftover code

frontend/TestRun/SCT/SctEvent.svelte
```diff
@@ -0,0 +1,194 @@
...
+
+    interface Props {
+        event: SCTEvent,
+        filterState: EventSeverityFilter,
+        options: Options,
+    }
+    let { event, filterState, options }: Props = $props();
+
+    const sliceMessage = function (message: string): string {
+        return fullMessage ? message : message.slice(0, Math.max(Math.ceil(message.length * 0.15), 500));
```
soyacz: 500 is magic number - please add it to const with description.
Btw. this approach is bad when we have message with length of 501 - we add ... to uncover one character.
I'd say we should have margin, e.g. if message is greater than 600 characters, slice by 500

frontend/TestRun/SCT/SctEvent.svelte
k0machi: Not leftover - re-enabled in latest commit

frontend/TestRun/SCT/SctEvents.svelte
```diff
@@ -0,0 +1,309 @@
...
+        let timeline = [...nemesisEvents, ...remainder].sort((l, r) => l.ts - r.ts);
+        return timeline;
+    }
+
+    let timeline: TimelineEvent[] = $derived(createTimeline(nemeses, events));
+
+    onMount(async () => {
+        events = await fetchEvents(false, 10000);
+        updateCounters();
+        refreshTimer = setInterval(async () => {
```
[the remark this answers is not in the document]
k0machi: Test Status might be set before all events have arrived - are you sure about this? While it would be a minor thing that would affect just one refresh after test end, it could still result in worse UX. Plus if we add attaching issues to events it would be nice seeing them appear in real time.

frontend/TestRun/SCT/SctNemesis.svelte
```diff
@@ -0,0 +1,84 @@
...
+            <div>
+                <Fa icon={faSpider} /> <span class="fw-bold">{event.name}</span>
+            </div>
+            <div class="ms-auto rounded bg-light text-dark">
+                {timestampToISODate(event.end_time * 1000, true)}
+            </div>
+            <div class="ms-2 px-1 rounded {NemesisStatusBg[event.status as keyof typeof NemesisStatusBg]} {NemesisStatusFg[event.status as keyof typeof NemesisStatusFg]}">
+                {event.status.toLocaleUpperCase()}
+            </div>
+            {#if innerEvents.length > 0}
```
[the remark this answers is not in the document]
k0machi: This is not related to that, this part works fine, the part where a nemesis should show itself if it has any errors without any filters was broken in another place. Fixed.

soyacz:
LGTM

Issue #826, opened 2025-10-20
Cannot show all failed build for a view


pehala:
https://argus.scylladb.com/view/scylla-2025-4-2025-4-0-rc1-2025-4-0. Clicking on Failed in non-investigated doesnt do anything and it throws JS error:
```
Uncaught TypeError: can't access property "105eef8c-f855-4915-9fc0-0d28c7863ebf", stats().groups is undefined
    handleQuickSelect https://argus.scylladb.com/s/dist/ViewDashboard-BNiYboDx.js:116
    handleQuickSelect https://argus.scylladb.com/s/dist/ViewDashboard-BNiYboDx.js:115
    bubble_event https://argus.scylladb.com/s/dist/SelectUtils-twQEqot0.js:178
    quickSelect https://argus.scylladb.com/s/dist/ViewTypes-DNEIwbuC.js:79
    bubble_event https://argus.scylladb.com/s/dist/SelectUtils-twQEqot0.js:178
    quickSelect https://argus.scylladb.com/s/dist/TestPopoutSelector-BB5J85zA.js:52
    createEventDispatcher https://argus.scylladb.com/s/dist/index-client-C4KcLXT0.js:102
    on_click https://argus.scylladb.com/s/dist/NumberStats-B_EwHxX1.js:17
    handle_event_propagation https://argus.scylladb.com/s/dist/index-ZFsBtOSq.js:3601

```
Firefox 143
<img width="1242" height="477" alt="Image" src="https://github.com/user-attachments/assets/fcc3874c-e88a-4249-8bbb-3028a58653bd" />

PR #833, opened 2025-10-30
fix(stats): include group data in quick select events

soyacz:
- Root cause: NumberStats only dispatched tests, so the view dashboard (which stores widget stats in a keyed map without a top-level groups) tried to resolve stats.groups[...] and crashed on undefined.
- Fix: include the relevant group stats in the quick-select event payload and teach all handlers to use the provided map (or fall back to their local stats) before selecting tests, keeping existing release dashboard behavior intact.

fixes: https://github.com/scylladb/argus/issues/826

### DIFF
--- a/frontend/ReleaseDashboard/ReleaseDashboard.svelte
+++ b/frontend/ReleaseDashboard/ReleaseDashboard.svelte
@@ -40,14 +40,16 @@
     };
 
     const handleQuickSelect = function (e) {
-        let tests = e.detail.tests;
+        let tests = e.detail.tests ?? [];
+        const detailGroups = e.detail.groups ?? {};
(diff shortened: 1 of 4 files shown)


k0machi:
> > Root cause: NumberStats only dispatched tests, so the view dashboard (which stores widget stats in a keyed map without a top-level groups) tried to resolve stats.groups[...] and crashed on undefined.
> 
> This is incorrect. Stats object always contains `groups` property (and `tests` as well), unless it hasn't been loaded yet. I will investigate the original bug.

This is a bug with #393, the `stats` object is different in Views and quick select needs to be aware of the key it is selecting from (both source and target) and refuse to select if that isn't possible. I'll set up a PR shortly

PR #837, opened 2025-11-06
improvement(jenkins_service): Replace user github token with global one

k0machi:
We don't want to use user tokens to resolve github repositories due to
us switching to a different authorization provider. This commit changes
the user token to the global app token with no changes in functionality.

### DIFF
--- a/argus/backend/controller/api.py
+++ b/argus/backend/controller/api.py
@@ -47,27 +47,6 @@ def app_version():
     })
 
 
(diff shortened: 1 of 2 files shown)


fruch (approved):

soyacz (approved):
I see also we have `@bp.route("/profile/github/token")` but I don't see where it is used.
can it be removed too?

PR #841, opened 2025-11-12
fix(ModalWindow.svelte): Size to 100% on small displays

### LINKED ISSUE #827: Improve sizing of Release planner popup for smaller displays
When only on half the screen (or on phone), the release planner pop up looks like this:

Image

I think it makes more sense to extend the popup to not have that big margin/padding on smaller widths. It is unnecessary small right now

k0machi:
This fixes an issue where modal windows would still be trying to size
themselves to 50 or 75% of the screen even if it was too small.

Since ModalWindow is not used everywhere this doesn't affect all modal
windows present in Argus, but it should affect most of them.

Closes #828
Fixes #827

### DIFF
--- a/frontend/Common/ModalWindow.svelte
+++ b/frontend/Common/ModalWindow.svelte
@@ -10,7 +10,8 @@
         body?: import('svelte').Snippet;
     }
 
-    let { widthClass = "h-50", title, body }: Props = $props();
+    let { widthClass = "w-50", title, body }: Props = $props();
+    let innerWidth = $state(0);
 
     function handleKeyDown(event) {
         if (event.key === "Escape") {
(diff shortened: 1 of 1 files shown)


k0machi:
Large-ish display:

image

Small-ish display:

image

frontend/Common/ModalWindow.svelte
```diff
@@ -19,10 +20,10 @@
     }
 </script>
 
-<svelte:window onkeydown={handleKeyDown} />
+<svelte:window onkeydown={handleKeyDown} bind:innerWidth={innerWidth} />
 <div class="modal-window">
     <div class="d-flex align-items-center justify-content-center p-4">
-        <div class="rounded bg-white p-4 {widthClass}">
+        <div class="rounded bg-white p-4 {innerWidth >= 768 ? widthClass : "w-100"}">
```
pehala: why not use [bootstrap](https://getbootstrap.com/docs/5.3/layout/containers/) containers?

pehala: ok, lets keep it as is

PR #850, opened 2025-11-20
improvement(ViewsManager): Add drag'n'drop for widget arrangement

### LINKED ISSUE #794: Views missing jobs overall summary
views doesn't have overwall jobs summary (like release dashboards has):
Image

I don't remember the exact reason we didn't do that, maybe cause of the quick filter that doesn't the same things.
but the numbers and summaries doesn't have a replacement in view.

@mykaul was noticing it missing, and was asking about it.

k0machi:
This commit adds the drag and drop functionality to view manager widget
list window, allowing user to quickly adjust widget settings.

Fixes #794 (Root cause since most users replace that widget with
something else)

### DIFF
--- a/frontend/AdminPanel/ViewWidget.svelte
+++ b/frontend/AdminPanel/ViewWidget.svelte
@@ -13,18 +13,25 @@
      * @typedef {Object} Props
      * @property {Widget} [widgetSettings]
      * @property {any} [items]
+     * @property {} drag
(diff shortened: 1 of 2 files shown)


soyacz:
> Nice, much better UX
> 
> Now can we reorder the views list as well ?

not yet, task added: https://github.com/scylladb/argus/issues/854

PR #859, opened 2025-12-15
fix(PytestOverview): URI encode tests to prevent resolution errors

k0machi:
This commit fixes an issue where a test with a `/` in its name would
cause the request to return 404 as Flask parses %2F as / before
dispatching the endpoint. For safety, the test names are now also URI
encoded to prevent other kinds of issues. Primary fix is to allow `/`
inside this particular endpoint parameter.


k0machi:
> can you add some tests for it?

I'll look into it.

PR #885, opened 2026-01-15
ARGUS-67 feature(plugins/sct): Configuration submission support

### LINKED ISSUE #838: Show test configuration
Often it is required to view test configuration:
- SCT test config (resulted one)
- scylla.yaml (to quickly know its content when investigating a run without downloading db logs)
- specific configs required by test e.g. for the task in https://github.com/scylladb/argus/issues/840

We should store and show such configuration files content in Argus.

Migrated to Jira: [ARGUS-67](https://scylladb.atlassian.net/browse/ARGUS-67)

[ARGUS-67]: https://scylladb.atlassian.net/browse/ARGUS-67?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

k0machi:
This commit adds support for submitting SCT Configuration into argus,
which includes SCT Config itself, derived from its yaml test case files,
scylla.yaml composed during the test and support for arbitrary configs
provided by the misc endpoint (custom name and content)

Fixes #838

### DIFF
--- a/argus/backend/service/client_service.py
+++ b/argus/backend/service/client_service.py
@@ -173,6 +173,14 @@ class ClientService:
         return {"status": "ok", "message": "Results submitted"}
 
     @staticmethod
+    def get_config_property(name: str, value: Any | str, run_id: str = None) -> list[RunConfigParam]:
+        dml = RunConfigParam.filter(name=name, value=str(value))
+        if run_id:
+            dml.filter(run_id=run_id)
+
+        return list(dml.all())
+
+    @staticmethod
     def get_config_store(run_id: str, config_name: str) -> RunConfiguration:
         try:
             config_store = RunConfiguration.get(run_id=run_id, name=config_name)
--- a/argus/backend/tests/client_service/test_client_service.py
+++ b/argus/backend/tests/client_service/test_client_service.py
@@ -1,0 +1,118 @@
+import base64
+from dataclasses import asdict, dataclass
+import json
+
+import pytest
+from flask.testing import FlaskClient
+
+from argus.backend.models.web import ArgusTest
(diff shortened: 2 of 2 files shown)


argus/backend/plugins/sct/testrun.py
```diff
@@ -391,3 +391,10 @@ class SCTJunitReports(Model):
     test_id = columns.UUID(primary_key=True, partition_key=True, required=True)
     file_name = columns.Text(primary_key=True, required=True)
     report = columns.Text(required=True)
+
+
+class SCTConfiguration(Model):
+    run_id = columns.UUID(primary_key=True, partition_key=True, required=True)
+    sct_config = columns.Text(required=True)
```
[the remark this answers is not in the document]
soyacz: @fruch do you anticipate such Argus usage scenario where we would like to find tests with the same specific param?

argus/backend/service/client_service.py
```diff
@@ -168,3 +171,61 @@ def submit_results(self, run_type: str, run_id: str, results: dict) -> dict[str,
         if result_failed:
             raise DataValidationError()
         return {"status": "ok", "message": "Results submitted"}
+
+    def get_config_store(self, run_id: str, config_name: str) -> RunConfiguration:
```
soyacz: these are still static methods

k0machi: Done

k0machi:
> one more thing - tests - would be good to have them here (maybe again will find some issue)

Added tests.

PR #926, opened 2026-03-11
feature(runs-assign): default job assignment based on priority

CodeLieutenant:
In argus, when you start a job, that job gets unassigned. This means that runner (or in special cases, one who needs to investigate the job) does not get the notification about the job. This changes to the priority list of assigning. This feature places the current job in the views when opening argus

Priority Assignment List:
1. Scheduled Assignee (one who will investigate in argus terms)
2. User who triggered run gets assigned
3. Some error happened -> unassigned as previous

### DIFF
--- a/argus/backend/plugins/core.py
+++ b/argus/backend/plugins/core.py
@@ -17,7 +17,8 @@ from argus.backend.models.web import (
     ArgusScheduleGroup,
     ArgusSchedule,
     ArgusScheduleTest,
-    ArgusScheduleAssignee
+    ArgusScheduleAssignee,
+    User,
 )
 from argus.backend.util.common import chunk
 from argus.common.enums import TestInvestigationStatus, TestStatus
@@ -87,6 +88,20 @@ class PluginModelBase(Model):
 
     def get_scheduled_assignee(self) -> UUID:
         return self.get_assignment()
+
+    def get_initial_assignee(self, started_by: str | None = None) -> UUID | None:
+        try:
+            investigation_assignee = self.get_scheduled_assignee()
+            if investigation_assignee:
+                return investigation_assignee
+        except Model.DoesNotExist:
+            pass
+
+        if not started_by:
+            return None
+
+        trigger_user = User.exists_by_name(started_by)
+        return trigger_user.id if trigger_user else None
 
     def _legacy_get_scheduled_assignee(self, associated_test: ArgusTest, associated_release: ArgusRelease) -> UUID:
         """
--- a/argus/backend/plugins/core.py
+++ b/argus/backend/plugins/core.py
@@ -20,10 +20,13 @@ from argus.backend.models.web import (
     ArgusScheduleAssignee,
     User,
 )
-from argus.backend.util.common import chunk
+from argus.backend.util.common import chunk, get_build_number
 from argus.common.enums import TestInvestigationStatus, TestStatus
+from argus.backend.service.jenkins_service import JenkinsService
 
 LOGGER = logging.getLogger(__name__)
+# Jenkins stores the requester as the email local part only (see JenkinsService.build_job)
+JENKINS_USER_EMAIL_DOMAIN = "scylladb.com"
 
 
 class PluginModelBase(Model):
@@ -86,22 +89,46 @@ class PluginModelBase(Model):
         # FIXME: Legacy fallback until we fully migrate to new plans
         return self._legacy_get_scheduled_assignee(associated_test=associated_test, associated_release=associated_release)
 
-    def get_scheduled_assignee(self) -> UUID:
-        return self.get_assignment()
-
-    def get_initial_assignee(self, started_by: str | None = None) -> UUID | None:
+    def get_assignee(self, started_by: str | None = None) -> UUID | None:
         try:
-            investigation_assignee = self.get_scheduled_assignee()
+            investigation_assignee = self.get_assignment()
             if investigation_assignee:
                 return investigation_assignee
         except Model.DoesNotExist:
             pass
 
-        if not started_by:
+        if started_by:
+            trigger_user = User.exists_by_name(started_by)
+            if trigger_user:
+                return trigger_user.id
+
+        jenkins_username = self._get_jenkins_requested_by_user()
+        if jenkins_username:
+            jenkins_user = User.exists_by_email(f"{jenkins_username}@{JENKINS_USER_EMAIL_DOMAIN}")
+            if jenkins_user:
+                return jenkins_user.id
+
+        return None
+
+    def _get_jenkins_requested_by_user(self) -> str | None:
+        if not self.build_job_url or not self.build_id:
             return None
 
-        trigger_user = User.exists_by_name(started_by)
-        return trigger_user.id if trigger_user else None
+        build_number = get_build_number(self.build_job_url)
+        if not build_number:
+            return None
+
+        try:
+            service = JenkinsService()
+            return service.get_requested_by_user(build_id=self.build_id, build_number=build_number)
+        except Exception:  # noqa: BLE001 – Jenkins may be unreachable; do not fail run submission
+            LOGGER.warning(
+                "Could not fetch REQUESTED_BY_USER from Jenkins for build %s #%s",
+                self.build_id,
+                build_number,
+                exc_info=True,
+            )
+            return None
 
     def _legacy_get_scheduled_assignee(self, associated_test: ArgusTest, associated_release: ArgusRelease) -> UUID:
         """
--- a/argus/backend/plugins/driver_matrix_tests/model.py
+++ b/argus/backend/plugins/driver_matrix_tests/model.py
@@ -168,10 +168,7 @@ class DriverTestRun(PluginModelBase):
         run.build_job_url = req.job_url
         run.start_time = datetime.utcnow()
         run.assign_categories()
-        try:
-            run.assignee = run.get_scheduled_assignee()
-        except Exception:
-            run.assignee = None
+        run.assignee = run.get_assignee()
 
         run.status = TestStatus.CREATED.value
         run.save()
@@ -347,10 +344,7 @@ class DriverTestRun(PluginModelBase):
         run.build_id = req.job_name
         run.build_job_url = req.job_url
         run.assign_categories()
-        try:
-            run.assignee = run.get_scheduled_assignee()
-        except Exception:
-            run.assignee = None
+        run.assignee = run.get_assignee()
         for key, value in req.test_environment.items():
             env_info = EnvironmentInfo()
             env_info.key = key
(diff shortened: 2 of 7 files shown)


k0machi (approved):
LGTM, minor comment

k0machi (changes requested):

argus/backend/plugins/core.py
```diff
@@ -88,6 +90,50 @@ def get_assignment(self, version: str | None = None) -> UUID | None:
...
+            pass
+
+        if started_by:
+            trigger_user = User.exists_by_name(started_by)
+            if trigger_user:
+                return trigger_user.id
+
+        jenkins_username = self._get_jenkins_requested_by_user()
+        if jenkins_username:
+            jenkins_user = User.exists_by_name(jenkins_username)
```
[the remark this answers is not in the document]
CodeLieutenant: Will then Argus figure it out if there is no `@scylladb.com`?

argus/backend/plugins/core.py
```diff
@@ -88,6 +90,49 @@ def get_assignment(self, version: str | None = None) -> UUID | None:
...
+            if jenkins_user:
+                return jenkins_user.id
+
+        return None
+
+    def _get_jenkins_requested_by_user(self) -> str | None:
+        build_job_url = getattr(self, "build_job_url", None)
+        build_id = getattr(self, "build_id", None)
```
pehala: why do we need defensive programming here? getattr is usually code smell of badly designed code

argus/backend/tests/test_assignee.py
```diff
@@ -138,3 +139,280 @@ def test_assign_testrun_to_other(flask_client, sct_run_for_assignee, test_user,
...
+        schedule_id=schedule.id,
+    )
+    schedule_test.save()
+
+    schedule_assignee = ArgusScheduleAssignee(
+        assignee=investigator.id,
+        release_id=fake_test.release_id,
+        schedule_id=schedule.id,
+    )
+    schedule_assignee.save()
```
pehala: Setup should not be part of the test, but instead a fixtures. It muddles the test and make ti harder to understand what is important to the test and what is only a setup.

argus/backend/tests/test_assignee.py
```diff
@@ -138,3 +139,280 @@ def test_assign_testrun_to_other(flask_client, sct_run_for_assignee, test_user,
...
+        content_type="application/json",
+    )
+    assert resp.status_code == 200, resp.json
+    assert resp.json["status"] == "ok"
+
+    run = SCTTestRun.get(id=run_id)
+    assert run.assignee is None, "run should remain unassigned when started_by user does not exist"
+
+
+def test_investigation_assignee_takes_priority_over_triggerer(flask_client, fake_test):
```
pehala: Please add docstrings to the tests

fruch:
@CodeLieutenant why are we dragging this one that long ? From the PR description sounds like it should be quite straight forward.

Seems like we are a bit over reviewing it ...

soyacz:
@CodeLieutenant please resolve conflicts and ping for final review

argus/backend/plugins/core.py
```diff
@@ -89,8 +92,46 @@ def get_assignment(self, version: str | None = None) -> UUID | None:
         # FIXME: Legacy fallback until we fully migrate to new plans
         return self._legacy_get_scheduled_assignee(associated_test=associated_test, associated_release=associated_release)
 
-    def get_scheduled_assignee(self) -> UUID:
-        return self.get_assignment()
+    def get_assignee(self, started_by: str | None = None) -> UUID | None:
+        try:
+            investigation_assignee = self.get_assignment()
```
soyacz: **Completed plans still win.** `get_assignment()` does not check `plan.completed` or `plan.ends_at`. For a test with no mapping, it falls back to `plan.owner`. This lookup has priority 1, so a finished plan that lists the test sends every new run to the old plan owner. The person who started the run is never assigned. Filter out completed or expired plans before you use the owner.

argus/backend/plugins/core.py
```diff
@@ -89,8 +92,46 @@ def get_assignment(self, version: str | None = None) -> UUID | None:
...
+            if investigation_assignee:
+                return investigation_assignee
```
soyacz: **The owner is not checked.** This returns the plan owner or mapped assignee UUID without a check that the user still exists. If the account is deleted, or `assignee_mapping` holds a stale UUID, the run gets a UUID with no user behind it. The `started_by` and Jenkins fallbacks then never run, even when the triggerer exists.

argus/backend/plugins/core.py
```diff
@@ -89,8 +92,46 @@ def get_assignment(self, version: str | None = None) -> UUID | None:
...
+            pass
+
+        if started_by:
+            trigger_user = User.exists_by_name(started_by)
+            if trigger_user:
+                return trigger_user.id
+
+        jenkins_username = self._get_jenkins_requested_by_user()
+        if jenkins_username:
+            jenkins_user = User.exists_by_email(f"{jenkins_username}@{JENKINS_USER_EMAIL_DOMAIN}")
```
soyacz: **Hardcoded email domain.** `JenkinsService.build_job` writes `REQUESTED_BY_USER` as `requested_by.email.split('@')[0]` for any domain. This line always rebuilds the email as `<local>@scylladb.com`. A user `alice@partner.com` gets no match. A different user `alice@scylladb.com` gets the run by mistake. Look up by username, or pass the full email in the build parameter.

argus/backend/plugins/core.py
```diff
@@ -89,8 +92,46 @@ def get_assignment(self, version: str | None = None) -> UUID | None:
...
+            if jenkins_user:
+                return jenkins_user.id
+
+        return None
+
+    def _get_jenkins_requested_by_user(self) -> str | None:
+        if not self.build_job_url or not self.build_id:
+            return None
+
+        build_number = get_build_number(self.build_job_url)
```
soyacz: **Unneeded Jenkins calls.** This path also runs for driver-matrix and sirenada, which never send `started_by`, and for generic runs from other CI systems. Each of those runs makes one extra Jenkins request on the submit path, or gets a 404. The code also parses the URL again when `self.build_number` is already set. Reuse `self.build_number` and query Jenkins only for plugins that use Jenkins.

argus/backend/plugins/core.py
```diff
@@ -89,8 +92,46 @@ def get_assignment(self, version: str | None = None) -> UUID | None:
...
+        if not build_number:
+            return None
+
+        try:
+            service = JenkinsService()
```
soyacz: **No timeout on the Jenkins call.** `JenkinsService()` builds python-jenkins with the default socket timeout, which is none. This call is on the synchronous submit path. If Jenkins is slow or hangs, `POST /testrun/{type}/submit` blocks until the client times out, and the run is never created. The broad `except` does not catch a hang. Set an explicit short timeout.

argus/backend/plugins/driver_matrix_tests/model.py
```diff
@@ -173,10 +173,7 @@ def submit_run(cls, request_data: dict) -> 'DriverTestRun':
         run.build_number = get_build_number(req.job_url)
         run.start_time = datetime.now(UTC)
         run.assign_categories()
-        try:
-            run.assignee = run.get_scheduled_assignee()
-        except Exception:
-            run.assignee = None
+        run.assignee = run.get_assignee()
```
soyacz: **Submission can now fail.** `get_assignee` catches only `DocumentNotFound` around `get_assignment()`, and it catches nothing around the `User` lookups. A database error, a read timeout, or two users with the same username or email now fails the submission with a 500. Before, the run was saved without an assignee. The same applies to line 353.

argus/backend/plugins/sct/testrun.py
```diff
@@ -300,20 +300,17 @@ def get_perf_results_for_test_name(cls, build_id: str, start_time: float, test_n
...
         run = cls.model_construct()
         run.build_id = req.job_name
+        run.build_job_url = req.job_url
+        run.build_number = get_build_number(req.job_url)
         run.assign_categories()
-        try:
-            run.assignee = run.get_scheduled_assignee()
-        except DocumentNotFound:
-            run.assignee = None
+        run.assignee = run.get_assignee(req.started_by)
```
soyacz: **No check on who can assign.** `started_by` is free text from the client payload. `REQUESTED_BY_USER` is a normal Jenkins build parameter. Any API token holder, or anyone who can start a Jenkins build, can put runs in the assigned-jobs view of any user. Decide if that is acceptable. If not, validate the triggerer against the caller.

argus/backend/tests/test_assignee.py
```diff
@@ -1,71 +1,126 @@
...
+        }
+        resp = api_client.post(SUBMIT_ENDPOINT, json=payload)
+        assert resp.status_code == 200, resp.json()
+        assert resp.json()["status"] == "ok"
+        return run_id
+    return _submit
+
+
+@pytest.fixture
+def sct_run_for_assignee(submit_sct_run, fake_test):
```
soyacz: **Real Jenkins calls in tests.** This fixture submits `started_by='test_user'` with `job_url='.../job/1'` and does not patch `JenkinsService`. The assign and unassign tests pass only because the constructor fails without `JENKINS_URL` and the blind `except` hides the error. On a machine with Jenkins credentials, these tests call the real Jenkins. Patch `JenkinsService` in the fixture.

argus/backend/tests/test_assignee.py
```diff
@@ -131,6 +179,90 @@ def test_assign_testrun_to_other(api_client, sct_run_for_assignee, test_user, sa
...
+
+    run = SCTTestRun.get(id=uuid.UUID(run_id))
+    assert run.assignee == triggerer.id, "run should be assigned to the person who triggered it"
+
+
+def test_run_unassigned_when_triggerer_unknown(submit_sct_run):
+    """A run stays unassigned when started_by matches no Argus user and Jenkins returns nothing."""
+    with jenkins_returns(None):
+        run_id = submit_sct_run(started_by="ghost_user_that_does_not_exist",
+                                job_url="http://example.com/job/unknown")
```
soyacz: **The Jenkins branch never runs.** `get_build_number('http://example.com/job/unknown')` returns `None`, so the code returns before it calls Jenkins, and the `jenkins_returns(None)` patch does nothing. A regression that assigns someone when Jenkins returns `None` would not fail this test. Use a URL with a build number.

soyacz (commented):
**Process:** this PR changes code but has no `tasks/<KEY>/` artifacts (`intent.md`, `spec.md` or `rca.md`) and no Jira key. The description also does not say that the flow was skipped. The repository `CLAUDE.md` requires the artifacts before the code. A very small fix may skip the flow, but the PR description must say so.

PR #957, opened 2026-04-02
docs: add request replay design document

CodeLieutenant:
Design document for an always-on JSONL replay log in the Python client and an `argus replay` CLI command to recover lost data when Argus is unavailable during test runs.

Closes QAINFRA-43

### DIFF
--- a/docs/plans/request_replay.md
+++ b/docs/plans/request_replay.md
@@ -1,0 +1,777 @@
+# Argus Request Replay
+
+## Problem Statement
+
+Argus can go down for various reasons — maintenance windows, infrastructure failures, bugs, database issues. Test results produced by SCT, DTest, driver-matrix-tests, and sirenada are expensive to generate (long-running tests on cloud infrastructure). When Argus is unavailable during a test run, all submitted data is permanently lost.
+
+The current retry mechanism in the Python client (`urllib3.Retry` with `max_retries=3` and `backoff_factor=1`) only handles transient network blips. It cannot recover from extended Argus outages that may last minutes or hours during a multi-hour test run.
+
+**Goal:** Build a replay system that captures every Argus API request in a machine-parseable format, always on by default, so that when Argus recovers, all failed submissions can be replayed to reconstruct the intended system state.
+
+---
+
+## 1. Current Implementation
+
+### 1.1 Client Architecture
+
+All Argus API communication flows through the `ArgusClient` base class in `argus/client/base.py`. Every HTTP request passes through exactly two methods:
+
+- `get()` (line 116) — used for status checks, data retrieval
+- `post()` (line 132) — used for all data submissions
+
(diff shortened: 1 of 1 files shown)


docs/plans/request_replay.md
```diff
@@ -0,0 +1,777 @@
...
+  "test_type": "scylla-cluster-tests",
+  "success": true
+}
+```
+
+**Field descriptions:**
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `seq` | int | Atomic, persistent sequence number. Determines replay order. |
```
soyacz: I'm not sure if we need that - just order of lines tells the sequence.
This adds complexity that I don't see reason for.

docs/plans/request_replay.md
```diff
@@ -0,0 +1,777 @@
...
+```bash
+# Find all replay logs and replay failed requests
+find /path/to/downloaded/logs -name "argus_replay_*.jsonl" \
+  | xargs argus replay --failed-only --report json > replay_report.json
+
+# Review the report
+cat replay_report.json | jq '.[] | select(.failed > 0)'
+```
+
+### 6.4 Server-Side Replay Ingest Endpoint
```
soyacz: this is the best solution.
It opens also more features like making test only saving to replay log and jenkins at the end sends this file to Argus. 
Make sure argus cli supports this one in this way.

soyacz (commented):
Generally, a very good plan.
I think we don't need seq and all these recovery methods - just one with submitting whole file to Argus should be enough.
Verify idempotency - probably all the endpoints are like that.
No need for get() replay I suppose, nor heartbeats

soyacz (commented):
I think it's over-engineered. 
I'd prefer wrapping client's post method with ctx manager, that would be thread safe.
In case we have out-of-order requests logged (which may happen) we will anyway sort it during replay by ts.

soyacz (commented):
why do we need to post heartbeat at all?

docs/plans/request_replay.md
```diff
@@ -0,0 +1,667 @@
...
+
+- Add `log_dir` parameter to `ArgusClient.__init__()` (required, no fallback — each consumer provides its test log directory)
+- Capture `init_ts_ms` at construction time; pass to `ReplayLog`
+- Initialize `ReplayLog` (always on)
+- Add `replay_only` mode: `get()` and `post()` skip the HTTP call but still write to the replay log
+- Wrap `post()` with `self._replay_log.record(...)` context manager
+
+**Remove from `argus/client/sct/client.py`:**
+
+- `submit_events()` method (legacy aggregated severity/count)
```
k0machi: Please do not remove it now. It is deprecated but not yet staged for removal.

k0machi (changes requested):

soyacz:
@k0machi can you take a look again?

k0machi (approved):

PR #960, opened 2026-04-08
refactor(plugins/sct): Move nemesis_data to a separate table

k0machi:
This commit changes the way nemesis data is stored inside argus,
migrating from a UDT inside SCTTestRun model into a new SCTNemesis table,
while introducing new nemesis_stats field in place to provide nemesis
statistics to TestDashboard component, mitigating the need to fetch data
from this table during stats queries. Finalize and Submit nemesis
endpoints were updated as well as new nemesis_get endpoint was
introduced inside the client in case it wants to look at nemesis data separately.

`get_run_response` was updated to fetch SCTNemesis data as that
particular endpoint only does a single run fetch, negating the need for
updating frontend components.

Fixes ARGUS-7

### DIFF
--- a/argus/backend/plugins/sct/testrun.py
+++ b/argus/backend/plugins/sct/testrun.py
@@ -2,6 +2,7 @@ from enum import Enum
 import logging
 from datetime import UTC, datetime, timezone
 from dataclasses import dataclass, field
+from threading import Lock
 from typing import Optional
 from uuid import UUID, uuid4
 
@@ -17,6 +18,7 @@ from argus.backend.plugins.sct.udt import (
     CloudSetupDetails,
     EventsBySeverity,
     NemesisRunInfo,
+    NodeDescription,
     PackageVersion,
     PerformanceHDRHistogram
 )
@@ -56,7 +58,6 @@ class SCTTestRunSubmissionRequest():
     runner_private_ip: Optional[str] = field(default=None)
 
 
-
 class SCTEventSeverity(str, Enum):
     CRITICAL = "CRITICAL"
     ERROR = "ERROR"
@@ -65,14 +66,13 @@ class SCTEventSeverity(str, Enum):
     DEBUG = "DEBUG"
 
 
-
 class StressCommand(Model):
     run_id = columns.UUID(primary_key=True, partition_key=True)
-    ts = columns.DateTime(primary_key=True, clustering_order="DESC", default=lambda: datetime.now(tz=UTC))
+    ts = columns.DateTime(
+        primary_key=True, clustering_order="DESC", default=lambda: datetime.now(tz=UTC))
     cmd = columns.Text()
     log_name = columns.Text()
     node_name = columns.Text()
-
 
 
 class SCTEvent(Model):
@@ -88,7 +88,7 @@ class SCTEvent(Model):
 
     # DB Log columns
     node = columns.Text()
-    received_timestamp = columns.DateTime() # SCT DB message timestamp
+    received_timestamp = columns.DateTime()  # SCT DB message timestamp
 
     # Nemesis columns
     nemesis_name = columns.Text()
@@ -98,7 +98,6 @@ class SCTEvent(Model):
 
     # Misc
     known_issue = columns.Text()
-
 
     @classmethod
     def table_name(cls):
@@ -114,6 +113,20 @@ class SCTUnprocessedEvent(Model):
     ts = columns.DateTime(primary_key=True, clustering_order="DESC")
 
 
+class SCTNemesis(Model):
+    __table_name__ = "sct_nemesis"
+
+    run_id = columns.UUID(partition_key=True)
+    start_time = columns.Integer(primary_key=True, clustering_order="ASC")
+    class_name = columns.Text()
+    name = columns.Text()
+    duration = columns.Integer()
+    target_node = columns.UserDefinedType(user_type=NodeDescription)
+    status = columns.Text()
+    end_time = columns.Integer()
+    stack_trace = columns.Text()
+
+
 class SCTTestRun(PluginModelBase):
     __table_name__ = "sct_test_run"
     _plugin_name = "scylla-cluster-tests"
@@ -126,7 +139,8 @@ class SCTTestRun(PluginModelBase):
     origin_url = columns.Text()
     started_by = columns.Text()
     config_files = columns.List(value_type=columns.Text())
-    packages = columns.List(value_type=columns.UserDefinedType(user_type=PackageVersion))
+    packages = columns.List(
+        value_type=columns.UserDefinedType(user_type=PackageVersion))
     scylla_version = columns.Text()
     version_source = columns.Text()
     yaml_test_duration = columns.Integer()
@@ -137,11 +151,16 @@ class SCTTestRun(PluginModelBase):
     cloud_setup = columns.UserDefinedType(user_type=CloudSetupDetails)
 
     # Test Runtime Resources
-    allocated_resources = columns.List(value_type=columns.UserDefinedType(user_type=CloudResource))
+    allocated_resources = columns.List(
+        value_type=columns.UserDefinedType(user_type=CloudResource))
 
     # Test Results
-    events = columns.List(value_type=columns.UserDefinedType(user_type=EventsBySeverity))
-    nemesis_data = columns.List(value_type=columns.UserDefinedType(user_type=NemesisRunInfo))
+    events = columns.List(
+        value_type=columns.UserDefinedType(user_type=EventsBySeverity))
+    nemesis_data = columns.List(
+        value_type=columns.UserDefinedType(user_type=NemesisRunInfo))
+    nemesis_stats = columns.Map(
+        key_type=columns.Text(), value_type=columns.Integer())
     screenshots = columns.List(value_type=columns.Text())
 
     # Subtest
@@ -175,8 +194,8 @@ class SCTTestRun(PluginModelBase):
 
     @classmethod
     def _stats_query(cls) -> str:
-        return ("SELECT id, test_id, group_id, release_id, status, start_time, build_job_url, build_id, "
-                f"assignee, end_time, investigation_status, heartbeat, build_number, scylla_version, cloud_setup, allocated_resources, nemesis_data FROM {cls.table_name()} WHERE build_id IN ? PER PARTITION LIMIT 15")
+        return ("SELECT id, test_id, group_id, release_id, status, start_time, build_job_url, build_id, nemesis_stats, "
+                f"assignee, end_time, investigation_status, heartbeat, build_number, scylla_version, cloud_setup, allocated_resources FROM {cls.table_name()} WHERE build_id IN ? PER PARTITION LIMIT 15")
 
     @classmethod
     def load_test_run(cls, run_id: UUID) -> 'SCTTestRun':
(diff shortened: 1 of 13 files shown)


soyacz (commented):
## PR Review: `refactor(plugins/sct): Move nemesis_data to a separate table`

The PR correctly addresses the core problem from the original PR #913 — moving nemesis data out of the `SCTTestRun` UDT list and into its own table. Most of the reviewer feedback from k0machi has been incorporated. Below is what's fixed, what's new, and what remaining issues I found.

---

### What's Been Fixed vs. PR #913

| Original concern                                               | Status                                                                                           |
| -------------------------------------------------------------- | ------------------------------------------------------------------------------------------------ |
| N queries in `stats.py` (one per run)                          | Fixed: `nemesis_stats` moved into `SCTTestRun` as a `Map` column; no extra queries in stats path |
| N queries in `nemesis_stats.py` and `graphed_stats.py`         | Fixed: both now batch-fetch via `SCTNemesis.filter(run_id__in=batch)` with `chunk()`             |
| Migration: unnecessary batch operation                         | Fixed: new migration (`migration_2026-04-08.py`) uses simple `nem.save()` per row                |
| `finalize_nemesis` race condition with `nemesis_stats__update` | Fixed: stats now tracked in `SCTTestRun.nemesis_stats` via `update_nemesis_stats()`              |
| Migration paging                                               | Fixed: old migration is replaced entirely; new one fetches via the ORM                           |
| `start_time` not cast to `int` in `finalize_nemesis`           | Fixed: `int(nem_req.start_time)` is used in `finalize_nemesis`                                   |
| `GET_NEMESIS` client route missing backend endpoint            | Client route added (`GET_NEMESIS`) but **no corresponding backend route exists** (see below)     |

---

### Bugs and Issues Found

**1. `SCTNemesis` is missing `__table_name__`** (`testrun.py:118`)

The model has no `__table_name__` attribute. cqlengine will derive the table name from the class name (`sctnemesis`), not `sct_nemesis` as assumed by both the migration script and by `SCTNemesis.table_name()` references throughout the code.

```python
class SCTNemesis(Model):
    # Missing: __table_name__ = "sct_nemesis"
    run_id = columns.UUID(partition_key=True)
```

This is a silent regression introduced in the PR — the original PR #913 diff did have `__table_name__ = "sct_nemesis"`.

**2. `nemesis_stats.py` has a copy-paste bug: populates `nemesis_data` instead of `nemesis_rows`** (`nemesis_stats.py:19–22`)

```python
nemesis_rows = []
for batch in {r.id for r in rows}:
    nemesis_data.extend(SCTNemesis.filter(run_id__in=batch).all())  # BUG: should be nemesis_rows
```

The results go into `nemesis_data` (the output list) instead of `nemesis_rows`. Then `nemesis_runs_data` (built from `nemesis_rows`) is always empty, so all per-run nemesis lookups return nothing. The function silently returns an empty list for every test.

**3. `{r.id for r in rows}` makes batching a no-op** (`nemesis_stats.py:19`, `graphed_stats.py:31`)

Both files iterate over a **set** of UUIDs one at a time instead of passing the full list to `chunk()`:

```python
for batch in {r.id for r in rows}:  # iterates over individual UUIDs, not batches
    nemesis_rows.extend(SCTNemesis.filter(run_id__in=batch).all())
```

`chunk()` is never called here. This results in one query per run — the exact N+1 problem the batch was meant to prevent. The correct pattern (as in the reviewed PR diff) is:

```python
run_ids = [r.id for r in rows]
for batch in chunk(run_ids):
    nemesis_rows.extend(SCTNemesis.filter(run_id__in=batch).all())
```

**4. `submit_nemesis` still fetches the full `SCTTestRun` row unnecessarily** (`service.py:394`)

After the refactor, `add_nemesis()` just calls `nemesis.save()` — it doesn't use the `run` object at all. The `SCTTestRun.get(id=run_id)` fetch is retained only to verify the run exists, which is a valid concern, but it's not documented and the `DoesNotExist` catch has now drifted: the `nemesis_row` is created _before_ the try block (`service.py:382`), so if `run_id` is invalid, the `SCTNemesis` row is already saved before the error is raised.

**5. `GET_NEMESIS` client route has no backend endpoint** (`client.py:29`)

The commit message says "new nemesis_get endpoint was introduced inside the client" but there is no corresponding `@bp.route("/<run_id>/nemesis/get")` in `controller.py`. The client will get a 404.

**6. `add_nemesis()` method is now misleading** (`testrun.py:398`)

The method signature accepts `SCTNemesis` and just calls `.save()` on it — it's a thin wrapper with a name that implies list mutation but does a DB write. Either remove it and call `nemesis.save()` directly (as the reviewed PR #913 diff did), or rename it to clarify the intent.

**7. `NemesisRunInfo` UDT is still imported in `testrun.py`** (`testrun.py:19`)

```python
from argus.backend.plugins.sct.udt import (
    ...
    NemesisRunInfo,
    NodeDescription,
```

`NemesisRunInfo` is no longer used directly in `testrun.py` (it still lives in the UDT module for use via the `nemesis_data` column on `SCTTestRun`, and in the migration), but the import was left in. This will trigger a linting warning.

**8. Migration does a full table scan without paging** (`migration_2026-04-08.py:16`)

```python
runs = SCTTestRun.filter().limit(None).only(["id", "nemesis_data", "build_id", "start_time"]).all()
```

The ORM `.all()` without `fetch_size` on the CQL session will buffer the entire result set in memory. The earlier PR #913 migration added explicit paging via `SimpleStatement(fetch_size=PAGE_SIZE)` precisely to avoid this for large tables. This regression should be addressed.

**9. `run_count` query may be slow on large tables** (`migration_2026-04-08.py:15`)

```python
run_count = DB.session.execute("SELECT count(*) FROM sct_test_run").one()["count"]
```

`COUNT(*)` on a Cassandra/ScyllaDB table is a full-table aggregation with no index — it will time out on production-scale tables. Use a best-effort counter or just omit the progress denominator.

---

### Minor Observations

- `graphed_stats.py:34`: the `run_id__in=batch` filter on a single UUID (bug #3) will still work functionally per-call, but defeats batching.
- `testrun.py:456`: `list(SCTNemesis.filter(run_id=run.id))` is missing `.all()` — should be `list(SCTNemesis.filter(run_id=run.id).all())` for consistency with other call-sites.
- `frontend/ReleaseDashboard/TestDashboardTest.svelte`: the nemesis stats display change looks correct and handles the new `nemesis_stats` map shape cleanly.

soyacz (commented):
# PR #960 Review: Move nemesis_data to a separate table

## Overview

This PR migrates nemesis data from an embedded UDT list (`nemesis_data` on `SCTTestRun`) to a dedicated `SCTNemesis` table, partitioned by `run_id` with `start_time` as clustering key. A new `nemesis_stats` map on `SCTTestRun` provides aggregated counts (total/failed/succeeded) so dashboard queries don't need to join the nemesis table. Includes a migration script and updates to all consumers.

This is a solid architectural improvement — moving unbounded list data out of the main model is the right call for ScyllaDB.

---

## Bug: Broken batching in `nemesis_stats.py`

**`argus/backend/service/views_widgets/nemesis_stats.py:19`** — The batching loop iterates over a bare set, not over chunks:

```python
for batch in {r.id for r in rows}:
    nemesis_rows.extend(SCTNemesis.filter(run_id__in=batch).all())
```

This iterates over **individual UUIDs**, passing each one to `run_id__in=<single_uuid>`. Compare with the correct version in `graphed_stats.py:31`:

```python
for batch in chunk({r.id for r in rows}):
    nemesis_rows.extend(SCTNemesis.filter(run_id__in=batch).all())
```

The `chunk()` call is missing. This won't crash but issues N individual queries instead of batched queries.

---

## Bug: Dead exception handler in `finalize_nemesis`

**`argus/backend/plugins/sct/service.py:421`** — `except StopIteration` is now unreachable:

```python
nemesis = SCTNemesis.get(run_id=run.id, start_time=int(nem_req.start_time))
# ...
except StopIteration as exception:  # <-- dead code
```

The old code used `next(nem for nem in ...)` which raised `StopIteration`. `SCTNemesis.get()` raises `SCTNemesis.DoesNotExist` (or the cassandra-driver equivalent). A missing nemesis will now bubble up unhandled as a 500 instead of the intended `SCTServiceException`.

Should be `except SCTNemesis.DoesNotExist`.

---

## Bug: `get_nemeses` client method returns nothing

**`argus/client/sct/client.py:283-291`** — The new `get_nemeses()` method calls `self.check_response(response)` but never returns the result:

```python
def get_nemeses(self):
    response = self.get(...)
    self.check_response(response)
    # missing: return response
```

Callers will always get `None`.

---

## Issue: Duplicate submit inflates `nemesis_stats`

**`argus/backend/plugins/sct/service.py:397-401`** — The old code had a deduplication guard:

```python
if not any(nem.name == nem_req.name and nem.start_time == int(nem_req.start_time) ...):
```

The new code unconditionally saves and increments `nemesis_stats["total"]`. Since `(run_id, start_time)` is the primary key, the Cassandra upsert means the nemesis row itself is idempotent, but `update_nemesis_stats("total")` will increment on every duplicate submit. If clients retry, the total count drifts.

---

## Issue: `nemesis_stats` read-modify-write race

**`argus/backend/plugins/sct/testrun.py:398-403`** — `update_nemesis_stats` does:

```python
stats = self.nemesis_stats if self.nemesis_stats else {}
val = stats.get(key, 0)
val += 1
```

This is a non-atomic read-modify-write. Two concurrent `submit_nemesis` calls for the same `run_id` will race, and one increment can be lost. For the current workload this may be acceptable (typically one SCT runner per run), but it's worth calling out.

---

## Minor: Missing `__table_name__` on `SCTNemesis`

**`argus/backend/plugins/sct/testrun.py:118`** — All other models (`SCTEvent`, `SCTUnprocessedEvent`, `SCTTestRun`) have explicit `__table_name__`. `SCTNemesis` relies on the driver's auto-generated table name (likely `s_c_t_nemesis` or `sctnemesis` depending on driver version). Should have an explicit name for consistency and predictability.

---

## Minor: Migration performance

**`scripts/migration/migration_2026-04-08.py:33`** — Each nemesis is saved individually (`nem.save()`) in a tight loop. For large datasets, batching inserts (using cassandra-driver's `BatchStatement` or at least async saves) would improve migration throughput substantially.

---

## Minor: Inconsistent optional chaining in frontend

**`frontend/ReleaseDashboard/TestDashboardTest.svelte:59`**:

```svelte
{lastRun?.nemesis_stats?.failed ?? 0} / {lastRun?.nemesis_stats.total}
```

The second access (`nemesis_stats.total`) lacks `?.` while the first uses it. It's guarded by the conditional, so it won't crash, but it's inconsistent.

---

## Summary

| Severity | Issue |
|----------|-------|
| **Bug** | `nemesis_stats.py` — missing `chunk()` in batching loop |
| **Bug** | `finalize_nemesis` — `except StopIteration` is dead, nemesis-not-found gives 500 |
| **Bug** | `client.py` — `get_nemeses()` doesn't return the response |
| **Medium** | Duplicate submit inflates `nemesis_stats` totals |
| **Low** | Read-modify-write race on `nemesis_stats` |
| **Low** | Missing `__table_name__` on `SCTNemesis` |
| **Low** | Migration script does single-row inserts |
| **Nit** | Inconsistent optional chaining in Svelte template |

The 3 bugs should be fixed before merge. The rest are worth discussing.

argus/backend/plugins/sct/testrun.py
```diff
@@ -114,9 +113,22 @@ class SCTUnprocessedEvent(Model):
...
+    target_node = columns.UserDefinedType(user_type=NodeDescription)
+    status = columns.Text()
+    end_time = columns.Integer()
+    stack_trace = columns.Text()
+
+
 class SCTTestRun(PluginModelBase):
     __table_name__ = "sct_test_run"
     _plugin_name = "scylla-cluster-tests"
+    _nemesis_update_lock = Lock()
```
soyacz: not used

soyacz (approved):
worth to add table name `Add __table_name__ = "sct_nemesis"`to SCTNemesis
Besides LGTM (fix the small nit about _nemesis_update_lock)

argus/backend/plugins/sct/testrun.py
k0machi: Forgot to remove.

soyacz: ok,then remove andadjust this table name as we discussed.
Then merge

PR #961, opened 2026-04-08
feature(cli): add GoReleaser config, release workflow, and changelog script

CodeLieutenant:
Add a complete GitHub Actions-based release pipeline for the Argus CLI:

- cli/.goreleaser.yml: GoReleaser v2 config that cross-compiles the 'argus' binary for linux/amd64, linux/arm64, darwin/amd64, darwin/arm64, and windows/amd64. Injects version/commit/date via ldflags. Packages Linux and macOS as .tar.gz and Windows as .zip. Produces a SHA256 checksums file. GoReleaser's built-in changelog is disabled in favour of the custom script.

- cli/scripts/generate-changelog.sh: pure-bash changelog generator that filters git log to only commits touching cli/, groups entries by conventional-commit type (Features, Bug Fixes, Improvements, Refactoring, etc.), strips the redundant '(cli)' scope prefix, and appends a collapsible full-commit table. Handles the FIRST sentinel for the initial release and produces a graceful fallback when no CLI commits exist in the given range.

- .github/workflows/cli-release.yml: three-job workflow triggered by 'cli/v*' tags. The 'changelog' job resolves the previous cli/* tag, runs the script, and uploads the notes as an artifact. The 'test' job runs the full Go test suite with -race, blocking the release on failure. The 'release' job downloads the changelog artifact, strips the 'cli/' tag prefix via GORELEASER_CURRENT_TAG, and runs GoReleaser to build and publish the GitHub Release with all binaries and checksums.


cli/scripts/generate-changelog.sh
```diff
@@ -0,0 +1,199 @@
...
+    GIT_RANGE="${FROM_REF}..${TO_REF}"
+fi
+
+# ---------------------------------------------------------------------------
+# Collect commits that touched the cli/ subtree.
+# We use a record separator (RS=0x1E, ASCII "record separator") between
+# fields so subjects containing pipes or special chars are handled safely.
+# Format per commit: SHA RS subject RS author NL
+# ---------------------------------------------------------------------------
+FIELD_SEP=$'\x1e'   # ASCII record separator — safe to use as field delimiter
```
k0machi: Can use a null byte as well

k0machi (approved):
Overall looks okay - I see it is currently failing with linting parts of code that have unchecked returns. You can address this in a separate commit in this PR so other PRs based on new master won't error out unnecessarily. 

Left two minor comments.

CodeLieutenant:
> > * **`BREAKING CHANGE` / `feat!:` not handled** (`cli/scripts/generate-changelog.sh:120`): The regex `^([a-zA-Z]+)(\([^)]*\))?[[:space:]]*:[[:space:]]*(.+)$` won't match `feat!: something` (the `!` breaks it). Breaking changes silently fall into `### Other Changes`. Add `!?` after the optional scope group:
> >   ```shell
> >   CONV_COMMIT_RE='^([a-zA-Z]+)(\([^)]*\))?(!)?[[:space:]]*:[[:space:]]*(.+)$'
> >   ```
> 
> We don't do commits like that.

this is just a placeholder, i really dont know what regex to use, looks like we have to go deep into files, and then check if the commit changes the file and then do changelog

cli/scripts/generate-changelog.sh
CodeLieutenant: Is it safe to use null byte?

k0machi (approved):

soyacz (commented):
@CodeLieutenant there are lint issues.
Also one from Claude - can you see if valid?

# PR 961 Review — CLI release pipeline

## Summary
Add GoReleaser config, tag-triggered release workflow, custom CLI-scoped changelog generator. Cross-compile argus binary for linux/darwin/windows × amd64/arm64.

## Critical

**`main.go` lacks `version`/`commit`/`date` package vars.** `cli/.goreleaser.yml:24-26` sets `-X main.version=...` but `cli/main.go` never declares these. Go's `-X` silently no-ops on missing symbols. Result: binary ships with empty version metadata. Also no `argus version` command exists to surface it.

Fix: add to `cli/main.go`:
```go
var (
    version = "dev"
    commit  = "none"
    date    = "unknown"
)
```
Plus a `version` subcommand.

## Notable

- **`cli/.goreleaser.yml:57-62` globs won't match.** Repo has no `cli/LICENSE*` or `cli/README*` (root-only). Archives ship without license/docs. Either point to `LICENSE*`/`README*` at repo root, or copy to cli/ before packaging.
- **Lint CI fails.** Pre-existing errors in `cli/cmd/cache.go`, `cli/cmd/logs.go`, `cli/internal/cache/cache.go` etc. Not introduced by this PR, but release workflow runs `test` not `lint` — broken lint won't block release. Consider adding lint to gating jobs.
- **`sort -V` wrong for pre-releases.** `cli/v1.0.0-rc1` sorts after `cli/v1.0.0`. Incremental tagging past a pre-release picks the RC as "previous." Minor — switch to `git tag --sort=-v:refname` or add filter.
- **Tag pattern `cli/v*` too loose.** Matches `cli/vtest`, `cli/vfoo`. Tighten to `cli/v[0-9]*` or `cli/v*.*.*`.
- **Markdown table injection.** Changelog table cells (`author`, `subject`) not escaped. Commit subject containing `|` breaks rendering. Replace `|` → `\|` before emit.
- **`draft: false` + `replace_existing_draft: true` contradictory but harmless.** Flag only fires when draft exists; config tolerates both paths.
- **`ignore: windows/arm64`** intentional per comment. Fine.
- **No master-ancestry check on tag.** Anyone pushing `cli/v*` tag on any branch triggers release. Consider guarding with `if: contains(github.event.base_ref, 'master')` or verifying `git merge-base --is-ancestor`.

## Nitpicks

- `cli/scripts/generate-changelog.sh:146-155` pattern `fix *` uses literal space — won't match `fix:something`, though `fix:*` branch catches it. Ordering OK.
- `cli/scripts/generate-changelog.sh:135` `scope_clean#cli` strips prefix only; `cli-foo(scope)` → `-foo`. Edge case.
- `cli/scripts/generate-changelog.sh:63-69` exits 0 with empty-changelog notice — workflow happily releases with no notes. Intentional per fallback comment.
- Duplicate test job between `cli-release.yml:78` and `cli-test.yml`. Defensible (tag push bypasses path-filtered PR workflow).
- Actions pinned to floating majors (`@v4`, `@v5`, `@v6`). SHA-pinning preferred for supply-chain hardening. Low priority.
- `actions/setup-go@v5` uses Node.js 20 (deprecation warning). Will be forced to Node 24 June 2026. Track, no action needed now.

## Correctness wins

- `fetch-depth: 0` on checkout — needed for `git tag --merged` and full log. Good.
- `--diff-filter=ACDMRT` on log — correct filter for commits actually touching `cli/`.
- Field separator `\x1e` in pretty-format — robust against pipes/quotes in subjects.
- `CGO_ENABLED=0` — correct for static cross-compile.
- `go mod tidy` removed from before.hooks with explicit rationale — correct call.
- `GORELEASER_CURRENT_TAG` strip trick — standard pattern for monorepo tags.
- Least-privilege default `contents: read`, escalation only in release job. Good.

## Verdict

Ship-blocker: missing `version`/`commit`/`date` vars. Binary version metadata is the primary artifact of a release pipeline — without the vars, GoReleaser ldflags are dead code. Fix that + add `version` subcommand, then LGTM. Other items are follow-ups.

.github/workflows/cli-release.yml
```diff
@@ -0,0 +1,157 @@
...
+
+                  if [[ -z "$PREV" ]]; then
+                    echo "No previous CLI tag found – treating this as the first release."
+                    PREV="FIRST"
+                  fi
+
+                  echo "Previous tag: ${PREV}"
+                  echo "tag=${PREV}" >> "$GITHUB_OUTPUT"
+
+            - name: Generate changelog
```
pehala, line 52: Just out of curiosity, cannot you generate changelog automatically via github releases?

soyacz: it's just for cli - so it would take all Argus changes which are not relevant for cli

k0machi:
@CodeLieutenant What's the status on this one? Are we good to merge it?

CodeLieutenant:
> @CodeLieutenant What's the status on this one? Are we good to merge it?

I need to refactor it, it is not what i wanted. Tomorrow we can merge it

CodeLieutenant:
@k0machi This one is ready for merging and official release of Argus CLI

k0machi (approved):

PR #962, opened 2026-04-09
feat(cli): deduplicate events with repeated_at timestamps

soyacz:
## Summary

- Deduplicate events in CLI output by aggregating duplicates (events with `duplicate_id` set by the backend AI similarity processor) into their original events
- Original events gain a `repeated_at` list containing the timestamps of all their duplicates; duplicate events are removed from output
- Table output includes a new "Repeats" column showing the duplicate count per event

Tested with run `348eae9d-2a41-40ff-976c-0262e8d64ce0`: 37 total events collapsed to 8 unique, with one `DatabaseLogEvent` showing 27 repeats.

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `go test ./...` — all 217 tests pass
- [x] JSON output shows `repeated_at` array with duplicate timestamps
- [x] Table output shows "Repeats" count column (empty when no duplicates)
- [x] Events without duplicates are unaffected

### DIFF
--- a/cli/cmd/run_nemeses_events.go
+++ b/cli/cmd/run_nemeses_events.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"sort"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cli/internal/models/runs.go
```diff
@@ -604,12 +606,17 @@ func (r SCTEventsResponse) Rows() [][]string {
 		if len(msg) > 200 {
 			msg = msg[:200] + "..."
 		}
+		repeats := ""
+		if n := len(e.RepeatedAt); n > 0 {
+			repeats = strconv.Itoa(n)
```
k0machi, line 611: This seems to me that this would output a timestamp value (milliseconds), instead of a date.

cli/internal/models/runs.go
```diff
@@ -207,6 +208,7 @@ type SCTEvent struct {
 	TargetNode        string           `json:"target_node"`
 	NemesisStatus     string           `json:"nemesis_status"`
 	KnownIssue        string           `json:"known_issue"`
+	RepeatedAt        []string         `json:"repeated_at,omitempty"`
```
k0machi, line 211: Here it is specified as an array of string

cli/internal/models/runs.go
k0machi, line 611: And here it would try to convert each value to string again.

cli/cmd/run_nemeses_events.go
```diff
@@ -121,6 +121,32 @@ func tsFromFlag(s string) int64 {
...
+			originals = append(originals, e)
+		}
+	}
+
+	for _, e := range events {
+		if e.DuplicateID == "" {
+			continue
+		}
+		if idx, ok := idxByEventID[e.DuplicateID]; ok {
+			originals[idx].RepeatedAt = append(originals[idx].RepeatedAt, e.Ts)
```
k0machi, line 150: Also I think this might result in a very tall column when outputting events in some cases, say, you have 1 event with 1000 duplicates. Maybe just a counter would be enough?

cli/cmd/run_nemeses_events.go
```diff
@@ -121,6 +121,32 @@ func tsFromFlag(s string) int64 {
 	return v
 }
 
+// deduplicateEvents aggregates duplicate events into their originals.
+// Duplicate events (DuplicateID != "") are removed from the output; their
+// timestamps are collected into the RepeatedAt field of the original event.
+func deduplicateEvents(events []models.SCTEvent) []models.SCTEvent {
+	idxByEventID := make(map[string]int, len(events))
```
k0machi, line 133: Supplying length to a hashmap allocation shouldn't be necessary.

cli/internal/models/runs.go
soyacz: no it gives:
```
"repeated_at": [
      "2026-04-07T01:56:38.736Z",
      "2026-04-07T01:56:41.385Z"
    ]
```

cli/cmd/run_nemeses_events.go
soyacz: future idea, currently it shows 1000 whole events, not lines

cli/internal/models/runs.go
k0machi: See my comment below - this operation isn't needed as you have those timestamps as a string and you are passing it to `itoa` (convert number to string).

k0machi: Nvm, it's converting a counter. I see

k0machi (approved):

CodeLieutenant (approved):

PR #963, opened 2026-04-10
fix(cli): trigger cloudflared login flow when cached CF token is missing or expired

CodeLieutenant:
## Summary

The CLI's Cloudflare Access authentication was broken: when the cached CF Access JWT was expired or missing, `cloudflared` never triggered the browser-based login flow, leaving the user stuck with opaque unauthorized errors. Additionally, Go's default HTTP redirect handling stripped auth credentials on 302 hops issued by CF Access, causing requests to land on an HTML login page instead of receiving JSON.

### Auth flow: cache-then-login (`cmd/root.go`, `internal/auth/argus.go`)

- Replace the read-only `cachedCFToken()` helper with `ensureCFToken()`, which first attempts the fast local cache read (`cloudflared access token`) and, if that fails and the session is interactive, falls back to the full browser-based login (`cloudflared access login`).
- In non-interactive mode (`--non-interactive`), return a clear error instead of silently proceeding without a token.
- Export `GetOrFetchCFToken` from the auth package so the root command can invoke the full token-fetch flow directly.

### Redirect auth preservation (`internal/api/api.go`)

- Install a custom `CheckRedirect` policy (`reattachAuthOnRedirect`) on the HTTP client that re-attaches the `Authorization` header and `CF_Authorization` cookie on every redirect hop.
- This fixes the scenario where CF Access responds with a 302 to its interstitial page and Go's default client strips the credentials, resulting in an HTML login page instead of JSON.

### Content-Type detection fix (`internal/api/api.go`)

- Switch from exact string comparison to `mime.ParseMediaType` when checking the response `Content-Type` in `DoJSON`.
- `application/json; charset=utf-8` is now correctly accepted instead of being misclassified as an unauthorized HTML response.

### Tests (`internal/api/api_test.go`)

- Content-Type with charset parameter is accepted as valid JSON.
- HTML Content-Type (with and without charset) is detected as unauthorized.
- HTTP 401 and 403 status codes return `ErrUnauthorized`.
- Auth credentials (Authorization header + CF_Authorization cookie) survive a same-origin 302 redirect.
- Redirect to an HTML login page still surfaces as `ErrUnauthorized`.

### DIFF
--- a/cli/internal/api/api.go
+++ b/cli/internal/api/api.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
 	"net/url"
 	"time"
@@ -136,6 +137,19 @@ func New(rawBaseURL string, opts ...ClientOption) (*Client, error) {
 		o(c)
 	}
 
+	// Install a redirect policy that re-attaches auth credentials on every
+	// hop.  Go's default http.Client strips cookies set via Request.AddCookie
+	// and drops the Authorization header on cross-origin redirects, which
+	// breaks Cloudflare Access: CF responds with a 302 to its interstitial
+	// page, the redirect loses the CF_Authorization cookie / API token, and
+	// the CLI ends up with an HTML login page instead of JSON.
+	//
+	// This must be set after applying opts so that WithHTTPClient (used by
+	// tests) can override the entire http.Client including CheckRedirect.
+	if c.httpClient.CheckRedirect == nil {
+		c.httpClient.CheckRedirect = c.reattachAuthOnRedirect
+	}
+
 	return c, nil
 }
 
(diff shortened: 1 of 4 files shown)


cli/internal/api/api.go
```diff
@@ -136,6 +137,19 @@ func New(rawBaseURL string, opts ...ClientOption) (*Client, error) {
...
+	// hop.  Go's default http.Client strips cookies set via Request.AddCookie
+	// and drops the Authorization header on cross-origin redirects, which
+	// breaks Cloudflare Access: CF responds with a 302 to its interstitial
+	// page, the redirect loses the CF_Authorization cookie / API token, and
+	// the CLI ends up with an HTML login page instead of JSON.
+	//
+	// This must be set after applying opts so that WithHTTPClient (used by
+	// tests) can override the entire http.Client including CheckRedirect.
+	if c.httpClient.CheckRedirect == nil {
+		c.httpClient.CheckRedirect = c.reattachAuthOnRedirect
```
k0machi, line 150: This doesn't sound very secure because you can potentially send cookies to a 3rd party website - exposing Authorization token to them, for example.

k0machi: Regardless, I don't think the assumption there is correct - you won't get json response with "unauthorized" if you get stuck on cloud flare, it'll just redirect you to an okta capture page.

cli/internal/api/api.go
```diff
@@ -199,6 +213,23 @@ func (c *Client) attachAuth(req *http.Request) {
...
 }
 
+// reattachAuthOnRedirect is a [http.Client.CheckRedirect] callback that
+// re-applies the client's authentication credentials to every redirect
+// request.
+//
+// Go's default redirect handling strips cookies set via [http.Request.AddCookie]
+// and drops the Authorization header when the redirect target is on a different
+// origin.  This breaks Cloudflare Access, which sits in front of the Argus
+// backend and may issue same-origin redirects (e.g. to /cdn-cgi/access/…)
```
k0machi, line 223: is that true that the cookie is dropped on same-origin?

k0machi: > /cdn-cgi/access/

That would not be same origin, argus sits at `argus.scylladb.com` while that endpoint is at `scylladb.cloudflareaccess.com`

k0machi:
In general receiving 302 while issuing API call means your CF_Authorization is no longer valid, you can't get 302 from argus itself if using api endpoints only (they'll return json response with error message in this case)

CodeLieutenant:
> In general receiving 302 while issuing API call means your CF_Authorization is no longer valid, you can't get 302 from argus itself if using api endpoints only (they'll return json response with error message in this case)

That is really good idea actually, didnt think about that

soyacz:
Merging as we need it. 
Refactor in a followup

PR #964, opened 2026-04-14
refactor(plugins/sct): Move allocated_resources to a separate table

k0machi:
This commit moves out the allocated_resources resources field to a
separate table, following up the same rework happening for Events and Nemeses.

All interfaces are preserved while removing unnecessary access in stats.py

Fixes ARGUS-4

### DIFF
--- a/argus/backend/plugins/sct/controller.py
+++ b/argus/backend/plugins/sct/controller.py
@@ -52,6 +52,27 @@ def sct_set_runner(run_id: str):
     }
 
 
+
+@bp.route("/<string:run_id>/resource/all", methods=["GET"])
+@api_login_required
+def sct_resource_all(run_id: str):
+    result = SCTService.get_resources(run_id=run_id)
+    return {
+        "status": "ok",
+        "response": result
+    }
+
+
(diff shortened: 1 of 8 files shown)


k0machi:
NB: I reformatted one of the files (service.py)

soyacz:
Findings
1. scripts/migration/migration_2026-04-14.py:29
   High: this migration is not idempotent and can overwrite newer resource state with stale legacy data. SCTResource.create(...) is an upsert on (run_id, name), so if the app has already started writing to sct_resource for a long-running pre-migration run, rerunning this migration will replace newer state, shards_amount, or termination fields with the older snapshot from allocated_resources. This needs an existence check or a rollout order that guarantees the migration finishes before writers switch over.
2. argus/backend/plugins/sct/testrun.py:323 and argus/backend/plugins/sct/testrun.py:461
   Medium: reads now fully ignore legacy allocated_resources. If this code is deployed before every existing SCT run is backfilled, older runs will start returning an empty resource list in the run API and in email rendering, even though the data is still present in sct_test_run. Keep a fallback to allocated_resources until migration completion is guaranteed, or make the migration a strict prerequisite for rollout.

soyacz (approved):

PR #965, opened 2026-04-14
Improve local development setup

pehala:
## Summary

- **Add seed data script**: Standalone `dev-db/seed_data.py` that populates ScyllaDB with realistic test data (admin user, releases, groups, tests, runs, events, results, issues) — idempotent by default, `--force` for clean recreation
- **Rewrite dev setup docs**: Restructured `docs/dev-setup.md` with a "Daily Workflow" TL;DR for returning developers, detailed first-time setup steps, and separate reference sections for configuration, troubleshooting, and frontend development

Fix https://scylladb.atlassian.net/browse/ARGUS-58

### DIFF
--- a/dev-db/seed_data.py
+++ b/dev-db/seed_data.py
@@ -1,0 +1,872 @@
+#!/usr/bin/env python3
+"""Seed a local Argus dev database with synthetic data.
+
+Usage:
+    python dev-db/seed_data.py                    # seed with admin/admin
+    python dev-db/seed_data.py --username dev --password dev123
+    python dev-db/seed_data.py --create-keyspace  # also create the keyspace
+    python dev-db/seed_data.py --force             # wipe and recreate seed data
+
+Run from the repository root so that argus_web.yaml is found automatically.
+"""
+
+import argparse
+import json
+import logging
+import os
+import random
+from datetime import UTC, datetime, timedelta
+from time import time
+from uuid import uuid4
+
+# Must be set before any cqlengine import
+os.environ["CQLENG_ALLOW_SCHEMA_MANAGEMENT"] = "1"
+
+from cassandra.auth import PlainTextAuthProvider
+from cassandra.cluster import Cluster
+from argus.backend.cli import sync_models
+from werkzeug.security import generate_password_hash
+
+from argus.backend.db import ScyllaCluster
+from argus.backend.models.github_issue import GithubIssue, IssueAssignee, IssueLabel, IssueLink
+from argus.backend.models.jira import JiraIssue
+from argus.backend.models.result import (
+    ArgusBestResultData,
+    ArgusGenericResultData,
+    ArgusGenericResultMetadata,
+    ArgusGraphView,
+    ColumnMetadata,
+)
+from argus.backend.models.web import (
+    ArgusEvent,
+    ArgusEventTypes,
+    ArgusGroup,
+    ArgusRelease,
+    ArgusTest,
+    User,
+    UserRoles,
+)
+from argus.backend.plugins.sct.testrun import SCTEvent, SCTEventSeverity, SCTTestRun
+from argus.backend.util.config import Config
+from argus.common.enums import TestInvestigationStatus, TestStatus
+
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+LOGGER = logging.getLogger("seed")
+
+# ---------------------------------------------------------------------------
+# Default seed content — edit this section to change what gets seeded.
+#
+# Everything the script creates is driven from this dict.  To add more
+# releases, groups, tests, event templates, result tables, or sample
+# issues just extend the corresponding list.
+# ---------------------------------------------------------------------------
+
+DEFAULT_CONTENT: dict = {
+    # ── Release & test hierarchy ────────────────────────────────────────
+    "release_name": "seed-release",
+    "release_pretty_name": "Seed Release",
+    "release_description": "Synthetic data for local development",
+    "groups": [
+        {
+            "name": "longevity-tests",
+            "pretty_name": "Longevity Tests",
+            "tests": [
+                ("longevity-10gb-3h-test", "Longevity 10GB 3h"),
+                ("longevity-100gb-12h-test", "Longevity 100GB 12h"),
+                ("longevity-large-partition-test", "Longevity Large Partition"),
+            ],
+        },
+        {
+            "name": "performance-tests",
+            "pretty_name": "Performance Tests",
+            "tests": [
+                ("perf-regression-latency-test", "Performance Regression Latency"),
+                ("perf-throughput-test", "Performance Throughput"),
+                ("perf-large-partition-test", "Performance Large Partition"),
+            ],
+        },
+    ],
+    # ── Test runs ───────────────────────────────────────────────────────
+    # Each profile produces one run per test.
+    "run_profiles": [
+        # (status, investigation_status)
+        (TestStatus.PASSED, TestInvestigationStatus.NOT_INVESTIGATED),
+        (TestStatus.PASSED, TestInvestigationStatus.NOT_INVESTIGATED),
+        (TestStatus.FAILED, TestInvestigationStatus.IN_PROGRESS),
+        (TestStatus.ERROR, TestInvestigationStatus.INVESTIGATED),
+        (TestStatus.ABORTED, TestInvestigationStatus.NOT_INVESTIGATED),
+    ],
+    "branches": ["master", "branch-2025.1", "next", "enterprise", "branch-6.2"],
+    "committers": ["jenkins-bot", "dev-user", "ci-runner", "test-automation"],
+    "events_per_run": (2, 4),  # (min, max) SCT events per run
+    # ── SCT event templates ─────────────────────────────────────────────
+    # Each tuple: (severity, event_type, message_template).
+    # Templates may use {node}, {step}, {dur}, {size} placeholders.
+    "event_templates": [
+        (SCTEventSeverity.NORMAL, "CassandraStressEvent", "stress command completed successfully on node {node}"),
+        (SCTEventSeverity.NORMAL, "TestFrameworkEvent", "Test step '{step}' passed — duration {dur}s"),
+        (
+            SCTEventSeverity.WARNING,
+            "DatabaseLogEvent",
+            "WARN  [shard 0] compaction - Compaction is slow, took {dur}ms for {size}MB",
+        ),
+        (
+            SCTEventSeverity.WARNING,
+            "YcsbStressEvent",
+            "YCSB stress operation timed out after {dur}ms on loader node {node}",
+        ),
+        (
+            SCTEventSeverity.ERROR,
+            "DatabaseLogEvent",
+            "ERROR [shard 2] storage_service - Decommission failed: node {node} unreachable",
+        ),
+        (SCTEventSeverity.ERROR, "CassandraStressEvent", "stress command failed with exit code 1 on node {node}"),
+        (
+            SCTEventSeverity.CRITICAL,
+            "DatabaseLogEvent",
+            "CRITICAL [shard 0] database - Commitlog disk error: /var/lib/scylla/commitlog",
+        ),
+        (SCTEventSeverity.CRITICAL, "InstanceStatusEvent", "Node {node} is down and not responding to health checks"),
+    ],
+    # ── Performance result tables ───────────────────────────────────────
+    # Each entry creates a result table attached to tests whose name
+    # starts with the given prefix (empty string = all tests).
+    "result_tables": [
+        {
+            "test_prefix": "perf-",
+            "name": "throughput",
+            "description_template": "Throughput and latency results for {test_pretty_name}",
+            "columns": [
+                {"name": "op_rate", "unit": "ops/s", "type": "float", "higher_is_better": True},
+                {"name": "latency_99th", "unit": "ms", "type": "float", "higher_is_better": False},
+                {"name": "latency_mean", "unit": "ms", "type": "float", "higher_is_better": False},
+            ],
+            "rows": ["write", "read", "mixed"],
+            # Value ranges for random data generation, keyed by column name.
+            # Columns not listed here get a uniform (1.0, 100.0) range.
+            "value_ranges": {
+                "op_rate": (50_000, 200_000),
+                "latency_99th": (2.0, 50.0),
+                "latency_mean": (0.5, 15.0),
+            },
+            # Which columns to include in the default graph view.
+            "graph_columns": ["op_rate"],
+        },
+    ],
+    # ── Sample issues ───────────────────────────────────────────────────
+    # Linked to failed/errored runs (round-robin).
+    # Uses nonexistent orgs/repos/projects so links never resolve to real issues.
+    "github_issues": [
+        {
+            "owner": "fake-org",
+            "repo": "fake-db",
+            "number": 19001,
+            "state": "open",
+            "title": "Decommission fails intermittently under high load",
+            "labels": [{"id": 1, "name": "bug", "color": "d73a4a", "description": "Something isn't working"}],
+            "assignees": [{"login": "fake-dev", "html_url": "https://github.com/fake-dev"}],
+        },
+        {
+            "owner": "fake-org",
+            "repo": "fake-cluster-tests",
+            "number": 5432,
+            "state": "open",
+            "title": "Stress command timeout on large partitions",
+            "labels": [
+                {"id": 2, "name": "test-failure", "color": "e4e669", "description": "Test infrastructure issue"},
+                {"id": 3, "name": "longevity", "color": "0075ca", "description": "Longevity test related"},
+            ],
+            "assignees": [],
+        },
+    ],
+    "jira_issues": [
+        {
+            "summary": "Investigate flaky longevity test failures in CI",
+            "key": "SEED-99999",
+            "state": "in progress",
+            "project": "SEED",
+            "permalink": "https://fake-org.atlassian.net/browse/SEED-99999",
+            "labels": [{"id": 10, "name": "flaky-test", "color": "fbca04", "description": "Flaky test"}],
+            "assignees": ["fake-dev"],
+        },
+        {
+            "summary": "Performance regression in write path after compaction refactor",
+            "key": "SEED-88888",
+            "state": "open",
+            "project": "SEED",
+            "permalink": "https://fake-org.atlassian.net/browse/SEED-88888",
+            "labels": [
+                {"id": 11, "name": "perf-regression", "color": "d93f0b", "description": "Performance regression"}
+            ],
+            "assignees": [],
+        },
+        {
+            "summary": "Node decommission timeout under mixed workloads",
+            "key": "SEED-77777",
+            "state": "in review",
+            "project": "SEED",
+            "permalink": "https://fake-org.atlassian.net/browse/SEED-77777",
+            "labels": [
+                {"id": 12, "name": "topology", "color": "0e8a16", "description": "Topology operations"},
+                {"id": 1, "name": "bug", "color": "d73a4a", "description": "Something isn't working"},
+            ],
+            "assignees": ["fake-dev"],
+        },
+    ],
+}
+
+
+# ---------------------------------------------------------------------------
+# Counters for summary
+# ---------------------------------------------------------------------------
+COUNTS = {
+    "users": 0,
+    "releases": 0,
+    "groups": 0,
+    "tests": 0,
+    "runs": 0,
+    "sct_events": 0,
+    "argus_events": 0,
+    "result_tables": 0,
+    "result_cells": 0,
+    "best_results": 0,
+    "graph_views": 0,
+    "github_issues": 0,
+    "jira_issues": 0,
+    "issue_links": 0,
+}
+
+
+# ---------------------------------------------------------------------------
+# 1. Database setup
+# ---------------------------------------------------------------------------
+
+
+def maybe_create_keyspace(config: dict):
+    """Create the keyspaces using a raw Cassandra connection (before cqlengine)."""
+    contact_points = config["SCYLLA_CONTACT_POINTS"]
+    username = config["SCYLLA_USERNAME"]
+    password = config["SCYLLA_PASSWORD"]
+    keyspace = config["SCYLLA_KEYSPACE_NAME"]
+    rf = config.get("SCYLLA_REPLICATION_FACTOR", 1)
+
+    auth = PlainTextAuthProvider(username=username, password=password)
+    cluster = Cluster(contact_points=contact_points, auth_provider=auth, protocol_version=4)
+    session = cluster.connect()
+
+    for ks in [keyspace, "argus_tablets"]:
+        LOGGER.info("Creating keyspace '%s' (RF=%s) if it does not exist...", ks, rf)
+        session.execute(
+            f"CREATE KEYSPACE IF NOT EXISTS {ks} "
+            f"WITH replication = {{'class': 'NetworkTopologyStrategy', 'replication_factor': {rf}}}"
+            f" AND tablets = {{'enabled': true}}"
+        )
+        LOGGER.info("Keyspace '%s' is ready.", ks)
+
+    cluster.shutdown()
+
+
+def setup_db():
+    """Connect to ScyllaDB and synchronize all table schemas."""
+    LOGGER.info("Connecting to ScyllaDB...")
+    cluster = ScyllaCluster.get()
+    LOGGER.info("Syncing core tables and types...")
+    sync_models(cluster.config["SCYLLA_KEYSPACE_NAME"])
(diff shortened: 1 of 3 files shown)


pehala:
> * **Fix frontend icon imports**: Pin FontAwesome packages to v6 — v7 ships only `.d.ts` stubs with no JavaScript, breaking all 99+ icon imports across the frontend

This was causing my login page to not show while trying this out, otherwise the doc should work. Let me know what you think

k0machi:
> > * **Fix frontend icon imports**: Pin FontAwesome packages to v6 — v7 ships only `.d.ts` stubs with no JavaScript, breaking all 99+ icon imports across the frontend
> 
> This was causing my login page to not show while trying this out, otherwise the doc should work. Let me know what you think

That doesn't sound right. All setups I have are on 7.0+, we have icons used from there iirc. Can you provide the stack trace on python/browser side?

dev-db/docker-compose.yaml
```diff
@@ -6,7 +6,7 @@ services:
       - HOME=/tmp
     networks:
       scylla_bridge:
-        ipv4_address: 172.18.0.2
+        ipv4_address: 172.22.0.2
```
k0machi: Actually it should be 18, the config is outdated.

k0machi:
> > > * **Fix frontend icon imports**: Pin FontAwesome packages to v6 — v7 ships only `.d.ts` stubs with no JavaScript, breaking all 99+ icon imports across the frontend
> > 
> > 
> > This was causing my login page to not show while trying this out, otherwise the doc should work. Let me know what you think
> 
> That doesn't sound right. All setups I have are on 7.0+, we have icons used from there iirc. Can you provide the stack trace on python/browser side?

image

pehala:
> > > > * **Fix frontend icon imports**: Pin FontAwesome packages to v6 — v7 ships only `.d.ts` stubs with no JavaScript, breaking all 99+ icon imports across the frontend
> > > 
> > > 
> > > This was causing my login page to not show while trying this out, otherwise the doc should work. Let me know what you think
> > 
> > 
> > That doesn't sound right. All setups I have are on 7.0+, we have icons used from there iirc. Can you provide the stack trace on python/browser side?
> 
> image

When I removed commit that downgraded the version it still worked, until I removed node_modules and did a completely clean install, then it started failing again. This is the error:

`Uncaught TypeError: The specifier “@fortawesome/free-solid-svg-icons” was a bare specifier, but was not remapped to anything. Relative module specifiers must start with “./”, “../” or “/”.`

k0machi:
could be something recent that had happened on the packaging side, I'll look into it. Generally we shouldn't downgrade font awesome.

pehala:
> could be something recent that had happened on the packaging side, I'll look into it. Generally we shouldn't downgrade font awesome.

I was, almost certainly, issue with corrupted yarn cache, which si why it reappared for me. Cleaning cache fixed it, sorry for wrongful report. The PR is now much cleaner and smaller

v2:
  * Remove .22 changes
  * Remove fontawesome donwgrade

k0machi (approved):

docs/dev-setup.md
```diff
@@ -1,146 +1,334 @@
...
+uv sync --extra web-backend --extra dev
+
+# Frontend
 yarn install
 ```
 
-## Configuring argus
-Create a `argus.local.yaml` configuration file (used to configure database connection) and a `argus_web.yaml` (used for webapp secrets) in your application install directory.
+> **Note:** `uv sync --all-extras` may fail on Python 3.14 due to
+> `onnxruntime`. Use the explicit extras above instead.
```
pehala, line 62: @soyacz There is a newer version which should have python 3.14 stubs, is it safe to upgrade?

soyacz (commented):
# Code Review — scylladb/argus#965: Improve local development setup

## Overview

Adds `dev-db/seed_data.py` (852 lines) to populate a local ScyllaDB with realistic test data in one command, and rewrites `docs/dev-setup.md` with a TL;DR daily workflow plus a more structured first-time guide. Replaces the previous fragmented "create keyspace → sync-models → download sample zst → run import" flow with a single `uv run python dev-db/seed_data.py --create-keyspace` invocation.

## Issues found

### Doc bugs

1. **Stale `argus.local.yaml` reference** — `docs/dev-setup.md:252-259` still instructs the user to edit `argus.local.yaml` with `contact_points` / `username` / `password` / `keyspace_name`. That file is never read by the codebase: `argus/backend/util/config.py` and `dev-db/import_data.py` both look only for `argus_web.yaml` with `SCYLLA_*` keys. The rewrite missed this section in the "Importing Production Data → From the production database" flow. Either delete the `argus.local.yaml` section or rewrite it to point at `argus_web.yaml` with `SCYLLA_CONTACT_POINTS`.

2. **`BASE_URL: "https://argus.scylladb.com"` in local config** (`docs/dev-setup.md:76`) — even for local dev, this sets the canonical URL to prod, which can leak into emails/share links. Consider `http://localhost:5000` for the dev snippet.

### Seed script — behavioral gaps

3. **`SCTUnprocessedEvent` queue not seeded** — in `argus/backend/plugins/sct/service.py:458-467`, the real event-submission path enqueues every ERROR/CRITICAL event into `SCTUnprocessedEvent` for embedding generation. Seed data skips that step (`create_sct_events` at `dev-db/seed_data.py:430-456` only writes to `SCTEvent`). Devs working on the similar-events / embedding features won't find seeded events in the unprocessed queue. Low priority — worth an inline comment or a TODO — but anyone relying on `find similar events` (c54ee7d) will be surprised.

4. **`create_sct_events` idempotency is fragile** — `dev-db/seed_data.py:425` samples a single `SCTEvent` with `severity=NORMAL` to decide whether to skip. Since event templates are randomly sampled (`random.sample(event_templates, k=num_events)`), a run may legitimately contain *no* NORMAL-severity event, and on re-run you'd get duplicates. Either (a) probe all severities via `run_id` alone isn't possible without an index, or (b) probe with `severity__in=[all severities]`. Safest: check if *any* event exists by iterating severities.

5. **`cleanup_seed_data` is chatty** (`dev-db/seed_data.py:723-726`) — for every run it loops over 5 severities and `.delete()`s each event individually. That's `runs × severities × events_per_severity` round trips (~450 for default content). Since `SCTEvent` is partitioned by `(run_id, severity)`, you can drop the whole partition with a single CQL `DELETE FROM sct_event WHERE run_id = ? AND severity = ?`. Same pattern applies to `ArgusGenericResultData` cleanup at lines 752-754.

### Minor

6. **Container name `dev-db-alpha-1` assumes Docker Compose v2 + project-name = dir-name** (`docs/dev-setup.md:21, 117`). Docker Compose v1 uses underscores (`dev-db_alpha_1`), and a custom `-p <name>` flag breaks it. Given the prerequisites already call out "Docker Compose" (v2 implied), this is acceptable, but `docker compose -f dev-db/docker-compose.yaml logs -f alpha` works regardless and is more robust in the wait-for-init snippet.

7. **`chmod 777` permissions fix** (`docs/dev-setup.md:133-136`) — running `alpine chmod -R 777` on config dirs is a big hammer. The compose file already uses `user: "${UID:-1000}:${GID:-1000}"`, so directories created by ScyllaDB inside the container will already be owned by the host user on Linux. The fix is mostly useful only for legacy state. Consider noting "if the dirs were created earlier under a different user" or recommending `chown` to `$UID:$GID` instead of a world-writable chmod.

8. **Hashing `dev-placeholder`-like strings in the admin user creation is fine**, but exposing the default `admin`/`admin` in logs at INFO (`dev-db/seed_data.py:285`) is loud. Not a security issue for dev data, just noise; OK as-is.

### Nice-to-haves (non-blocking)

- The `COUNTS` summary reflects only what the current invocation *created*. On a re-run where everything is skipped, it prints nothing meaningful. A "seed data already present" terminal line would reassure the user.
- `ArgusEvent.filter(run_id=runs[0].id)` at `dev-db/seed_data.py:467` uses a secondary index. Fine at this scale; worth a comment if more data is ever seeded in bulk.
- `graphs=graphs` at line 597 stores each value as `json.dumps(...)`. Worth verifying against how `ArgusGraphView` is consumed by the frontend (some codepaths may re-serialize, leading to double-encoded JSON).

## Test coverage

No tests added. Acceptable for a dev-only tooling script, but given the script's value to onboarding, a minimal smoke test (`main()` runs end-to-end against an ephemeral keyspace) would protect against regressions when models evolve. The existing test harness in `argus/backend/tests/sct_events/` provides a good template.

## Security

- Default `admin/admin` credentials are documented for local use. `SECRET_KEY: MUSTBEUNIQUE` in the example is unchanged from `argus_web.example.yaml`. Both are fine for a local-dev doc and clearly scoped as such.
- `CQLENG_ALLOW_SCHEMA_MANAGEMENT=1` is required and correctly scoped to the seed script / dev setup.

## Recommendation

**Approve with two required fixes:**

1. Drop/rewrite the stale `argus.local.yaml` section in the import-data flow (`docs/dev-setup.md:252-259`).
2. Tighten the `create_sct_events` idempotency probe (`dev-db/seed_data.py:425`) to not depend on NORMAL events existing.

The rest are nice-to-haves. The core concern — that the seed script uses current tables — checks out: `SCTEvent` (new table) for events, `GithubIssue` (new) for issues, and all other models are the current ones. No deprecated-table writes.

pehala:
Addressed comments
* Fix stale references
* Remove production url
* Add Jira issues
* Use fake Urls for Jira/Github
* Cleanup the script a bit

dev-db/seed_data.py
```diff
@@ -0,0 +1,888 @@
...
+# ---------------------------------------------------------------------------
+
+
+def maybe_create_keyspace(config: dict):
+    """Create the keyspaces using a raw Cassandra connection (before cqlengine)."""
+    contact_points = config["SCYLLA_CONTACT_POINTS"]
+    username = config["SCYLLA_USERNAME"]
+    password = config["SCYLLA_PASSWORD"]
+    keyspace = config["SCYLLA_KEYSPACE_NAME"]
+    rf = config.get("SCYLLA_REPLICATION_FACTOR", 1)
```
k0machi, line 251: Ideally this should be configurable as I for example use a 3 node cluster on dev machine.

dev-db/seed_data.py
```diff
@@ -0,0 +1,888 @@
...
+
+    auth = PlainTextAuthProvider(username=username, password=password)
+    cluster = Cluster(contact_points=contact_points, auth_provider=auth, protocol_version=4)
+    session = cluster.connect()
+
+    for ks in [keyspace, "argus_tablets"]:
+        LOGGER.info("Creating keyspace '%s' (RF=%s) if it does not exist...", ks, rf)
+        tablets_clause = " AND tablets = {'enabled': true}" if ks == "argus_tablets" else ""
```
k0machi: tablets should be enabled by default on all new deployments.

dev-db/seed_data.py
```diff
@@ -0,0 +1,888 @@
...
+            f"CREATE KEYSPACE IF NOT EXISTS {ks} "
+            f"WITH replication = {{'class': 'NetworkTopologyStrategy', 'replication_factor': {rf}}}"
+            f"{tablets_clause}"
+        )
+        LOGGER.info("Keyspace '%s' is ready.", ks)
+
+    cluster.shutdown()
+
+
+def setup_db():
```
k0machi, line 269: This can call cli module instead.

dev-db/seed_data.py
```diff
@@ -0,0 +1,888 @@
...
+def create_admin_user(username: str, password: str) -> User:
+    """Create an admin user with all roles."""
+    existing = list(User.filter(username=username).limit(1))
+    if existing:
+        LOGGER.info("  User '%s' already exists, skipping.", username)
+        return existing[0]
+
+    user = User.create(
+        id=uuid4(),
+        username=username,
```
k0machi, line 293: This should be normalized to github regex, should be safe in most cases as is, but I would like to make sure.

dev-db/seed_data.py
```diff
@@ -0,0 +1,888 @@
...
+
+def _events_exist_for_run(run_id) -> bool:
+    """Check whether any SCT events already exist for a run (across all severities)."""
+    for severity in SCTEventSeverity:
+        if list(SCTEvent.filter(run_id=run_id, severity=severity.value).limit(1)):
+            return True
+    return False
+
+
+def create_sct_events(runs: list[SCTTestRun], content: dict):
```
k0machi, line 443: As mentioned by AI review, a separate function to add SCT events to a dedup queue (in case someone is working on duplicate events) would be nice. Doesn't have to be in the same seed function, maybe a separate dev cli command to force all events into the unprocessed queue.

dev-db/seed_data.py
```diff
@@ -0,0 +1,888 @@
...
+
+    release = releases[0]
+    tests = list(ArgusTest.filter(release_id=release.id).all())
+    groups = list(ArgusGroup.filter(release_id=release.id).all())
+
+    # Collect all result table names from content for cleanup
+    result_table_names = [t["name"] for t in content.get("result_tables", [])]
+
+    # Raw CQL session for efficient partition-level deletes
+    session = ScyllaCluster.get().session
```
k0machi: Not sure what is being meant here, CQLEngine DML supports filter().delete().

docs/dev-setup.md
```diff
@@ -1,146 +1,331 @@
...
 ## Prerequisites
-1. NodeJS >=16 (with npm)
-2. Yarn (can be installed globally with `npm -g install yarn`)
-3. uv (https://docs.astral.sh/uv/getting-started/installation/)
-4. docker-compose
 
-## Installing argus dependencies
+- Python 3.12+
+- [uv](https://docs.astral.sh/uv/getting-started/installation/) (Python package manager)
+- Node.js >= 16 with npm
```
k0machi: Probably should say 22 at this point.

docs/dev-setup.md
k0machi: Should be safe as onnx is only used for chromadb which will be deprecated soon anyway. If you're not sure then don't update for now

docs/dev-setup.md
```diff
@@ -1,146 +1,331 @@
...
 SCYLLA_KEYSPACE_NAME: argus
 
 APP_LOG_LEVEL: INFO
 SECRET_KEY: MUSTBEUNIQUE
 
-GITHUB_CLIENT_ID: <>
-GITHUB_CLIENT_SECRET: <>
-GITHUB_ACCESS_TOKEN: unknown
-# List of required organization names (Comment out to disable organization requirement)
-#GITHUB_REQUIRED_ORGANIZATIONS:
```
k0machi, line 48: Why are documented fields being removed?

docs/dev-setup.md
```diff
@@ -1,146 +1,331 @@
...
+LOGIN_METHODS:
+    - password
 
-JENKINS_URL: https://jenkins.scylladb.com
-JENKINS_USER: <your username>
-JENKINS_API_TOKEN_NAME: argus-dev
-JENKINS_API_TOKEN: <jenkins api token>
+# Placeholders -- the app requires these keys at startup even if you
+# aren't using GitHub or email features. Without them every API call
```
k0machi, line 90: I don't think **every** API call would fail without Github credentials, only gh related ones should.

docs/dev-setup.md
```diff
@@ -1,146 +1,331 @@
...
-Compile frontend files from `/frontend` into `/public/dist`. Add --watch to recompile files on change.
+### From the production database
+
+1. Set up an SSH tunnel to the production cluster:
+
+    ```bash
+    ssh argus-web \
+      -L 127.0.0.10:9042:10.10.2.62:9042 \
+      -L 127.0.0.11:9042:10.10.2.157:9042 \
+      -L 127.0.0.12:9042:10.10.2.74:9042
```
k0machi: why are production IP addresses being exposed in a public Readme? They're RFC1918, sure, but still..

docs/dev-setup.md
```diff
@@ -1,146 +1,331 @@
...
 
-## Try it
-Open browser at `http://localhost:5000` (use localhost as Github OAuth setup instead of 127.0.0.1 )
+### `KeyError: 'EMAIL_SENDER'` or `KeyError: 'GITHUB_ACCESS_TOKEN'`
+
+Your `argus_web.yaml` is missing required keys. The app reads these
+eagerly at startup. Add the placeholder values shown in the config
+section above.
+
+### ScyllaDB won't start -- permission denied
```
k0machi, line 265: Actually number 1 reason might be not having enough async handles allowed - refer to ScyllaDB adminstration doc.

docs/dev-setup.md
```diff
@@ -1,146 +1,331 @@
...
+
+### `Connection error: AuthenticationFailed`
+
+ScyllaDB has authentication enabled. Always use `-u cassandra -p cassandra`
+when connecting via `cqlsh`.
+
+### `uv sync --all-extras` fails
+
+Use `uv sync --extra web-backend --extra dev` instead. The `onnxruntime`
+dependency may not build on your Python version.
```
k0machi: You can add that `uv` can install required python version for you via
```uv sync --python 3.12``` etc.

File: docs/dev-setup.md [hunk 15]
```diff
@@ -1,146 +1,331 @@
...
+
+### Port 5000 already in use
```
k0machi: SCT or test.py might use it for testing LDAP.

k0machi (changes requested):

dev-db/seed_data.py
pehala: it fetches it from the config no? so it is configurable

dev-db/seed_data.py
pehala: Do we have some kind of verification in the Argus that username satisfy the regex?

docs/dev-setup.md
pehala: They do in my experience. The GH (and Jira) are intiailized on module level I think

Edit: vefiying to make sure it wasnt a fluke and that I wasnt missing something

docs/dev-setup.md
pehala: This is preexisting, we always had direct production IPs in dev setup, see https://github.com/scylladb/argus/blob/master/docs/dev-setup.md

docs/dev-setup.md
pehala: I did not encounter it, can you share the exact wording we should use here? Alternatively, it can be added in a followup

docs/dev-setup.md
pehala: Yes, but I would not link other projects, it can happen with anything. If anything, maybe we can remove this recomendation, it is pretty clear. Let me know what you prefer

dev-db/seed_data.py
k0machi: Oh, sorry, didn't noticce, I thought it was hardcoded to 1 here.

dev-db/seed_data.py
k0machi: It was implicit from the github registration days - anything that Github allows would be what Argus allows. You can check frontend code for detecting Github username in the Comment.svelte for a regex that can verify that. At the very least I would like to avoid spaces in usernames.

docs/dev-setup.md
k0machi: Interesting, must be a recent regression.

dev-db/seed_data.py
pehala: You mean the GO module we have? I couldnt find any documentation how to use it or built it or what it can actually do.

Sidenote: having 3 languages in one repository is not a good user experience, it is annoying to setup one toolchain let alone three

dev-db/seed_data.py
pehala: Could we do this in a followup instead? I envision this as a base, and subsequent PRs can add more features, when they are needed. For example, I have a PR in pipeline which adds more version into the DB, because I need the data to test the fix I am working on.

docs/dev-setup.md
k0machi: Oh, in that case they are very old and incorrect - we should probably refer people to the confluence page about Argus tunnels.

We don't have one right now with up-to-date DNS names for the tunnel, but we should probably put it there.

docs/dev-setup.md
k0machi: https://github.com/scylladb/scylladb/issues/15010

Here's an example

docs/dev-setup.md
k0machi: At some point we might decide to switch the default port to something in the 8xxx range, let's remove this advice since it's fairly self-explanatory on what to do.

docs/dev-setup.md
pehala: I will leave as is, the note can be removed when we do removal of the dependency

dev-db/seed_data.py
k0machi: No, no, I mean the flask cli module in `argus/backend/cli.py`

k0machi: > Sidenote: having 3 languages in one repository is not a good user experience, it is annoying to setup one toolchain let alone three

Agreed, we might consider moving from monorepo approach at some point. I already have issues with go toolchain not wanting to detect go module if my CWD isn't in ./cli.

dev-db/seed_data.py
k0machi: Sure

dev-db/seed_data.py
pehala: Justifications is this:
`The two DELETE FROM statements use raw CQL to delete entire partitions in one shot, rather than the ORM's read-all-then-delete-one-by-one approach.`
But after confronting Ai about it, it admitted that it might not be the case. So I will change it to the ORM only

Good catch

docs/dev-setup.md
pehala: Github is required in Argus service init with no fallback:
https://github.com/scylladb/argus/blob/443c3676fac7b6177a8ffeb5be77690fe9cc8dbc/argus/backend/service/argus_service.py#L54

Emails is as well, through NotificationManager until it gets to Email class, which again has it in the init:
https://github.com/scylladb/argus/blob/master/argus/backend/util/send_email.py#L18

docs/dev-setup.md
pehala: I will gladyl remove "How to copy from production" guide, it shouldn't exist due to variety of reasons: 
1) Production IP leakage
2) Risks mishandling the production
    - If you connect you will have full access

Let me know if you want to keep it or remove it

docs/dev-setup.md
pehala: I can return them, but dev_setup is bad place for documenting configuring options. The new dev_setup documents minimal fields you need to have to have argus running. I would consider documenting the configuration elsewhere, prohably in README directly, or production setup doc

docs/dev-setup.md
pehala: Sidenote: I feel the same about leaking confluecen page as I do about IP leakage :D My personal recommendation would not be to link anything from the repo itself

docs/dev-setup.md
k0machi: To be honest I don't know what should be done here - referring developer to the example file with up-to-date doc is probably the best (and then updating the example file with changes in this doc).

docs/dev-setup.md
k0machi: `self.github_headers` is quite old actually and isn't related much to the issue code nowadays. Whole `argus_service.py` is actually half dead-code I think or at least features that are no longer in use. Tho I see it does access this directly, which is problematic.

I don't see github stuff in send_email however.

dev-db/seed_data.py
pehala: Given that username is parameter inputed by user and defaults to admin, I would keep as is.
It would require misuse of the script to write invalid username, which I think will be extremely rare (I suspect noone is going to use the override option anyway).

If we decide to add verificaiton into Argus itself, it might make sense to introduce verification here as well

docs/dev-setup.md
k0machi: Agreed, maybe just leave a note to ask in the relevant chat room or something.

docs/dev-setup.md
pehala: in send email, the email vars are fetched eagerily, not github, which is why we need both email and github vars with placeholders

pehala: Lets create an issue to remove it and keep the note as it is here

dev-db/seed_data.py
pehala: Done, it reuses sync_models

docs/dev-setup.md
pehala: We can discuss it more and submit as a followup.

docs/dev-setup.md
pehala: Added

docs/dev-setup.md
pehala: Removed

k0machi (approved):

pehala:
v3:
* Fix Jira states to be lower case

soyacz (approved):
I didn't test it, but I assume it works

PR #967, opened 2026-04-15
feature(ssh-tunnel): add secure multi-host tunnel backend and admin APIs

CodeLieutenant:
Implements backend SSH tunnel support from `docs/plans/ssh-tunnel-design.md` with secure key registration, multi-host proxy routing, admin management APIs, and full backend test coverage.

### DB models (Steps 1 + 2)
- `argus/backend/models/ssh_key.py`
  - `SSHTunnelKey`: stores client SSH public keys with TTL expiration.
  - `ProxyTunnelConfig`: stores proxy host tunnel config (`host`, `port`, `proxy_user`, `target_host`, `target_port`, `host_key_fingerprint`, `service_user_id`, `is_active`).
  - `is_active` default is now `true` for new configs.
- `argus/backend/models/web.py`
  - SSH models registered in `USED_MODELS`.
  - Added dedicated role: `ROLE_SSH_TUNNEL_SERVER`.

### Backend service (Step 3)
- `argus/backend/service/tunnel_service.py`
  - Type-safe DTO response models used for tunnel/key/config flows.
  - `register_tunnel(...)`: validates key and TTL, stores key with TTL, returns tunnel connection details.
  - `get_tunnel_connection(...)`: returns active proxy details with round-robin selection across active hosts.
  - `get_authorized_keys()`: returns all non-expired public keys for proxy authorized_keys use.
  - Admin helpers: `list_keys(...)`, `delete_key(...)`, `get_proxy_tunnel_config(...)`, `save_proxy_tunnel_config(...)`, `list_proxy_tunnel_configs(...)`, `set_proxy_tunnel_config_active(...)`.
- Review follow-ups implemented:
  - Minimum key TTL lowered to **1 hour** (`MIN_TTL_SECONDS=3600`, default remains 24h).
  - Replaced model `_as_dict()` style returns with explicit DTO mapping.
  - `delete_key(...)` is idempotent (missing/TTL-expired key does not raise).

### Host fingerprint onboarding improvement
- `save_proxy_tunnel_config(...)` no longer requires manual `host_key_fingerprint` input.
- Backend now resolves host key fingerprint automatically via `ssh-keyscan` (preferred key order: ed25519 -> ecdsa -> rsa), computes OpenSSH-style `SHA256:...`, and stores it.
- This removes manual fingerprint copy/paste from admin config creation.

### Client API routes (Step 4a)
- `argus/backend/controller/ssh_api.py` under `/api/v1/client/ssh`:
  - `POST /tunnel`
  - `GET /tunnel`
  - `GET /tunnel/keys`
  - `GET /keys` (authorized_keys plain-text endpoint)
- `argus/backend/controller/client_api.py`
  - SSH blueprint registration.

### Admin API routes (Step 4b)
- `argus/backend/controller/admin_api.py` under `/admin/api/v1` (Admin-only):
  - `GET /proxy-tunnel/config` (optional `tunnel_id`; returns only active config)
  - `GET /proxy-tunnel/configs` (optional `active_only=true|false`)
  - `POST /proxy-tunnel/config` (creates config; `is_active` optional)
  - `POST /proxy-tunnel/config/<tunnel_id>/active` (enable/disable host)
  - `GET /ssh/keys`
  - `DELETE /ssh/keys/<key_id>`
- All endpoints follow existing admin semantics:
  - `@check_roles(UserRoles.Admin)`
  - `@api_login_required`

### Security hardening
- `argus/backend/controller/ssh_api.py`
  - `GET /client/ssh/keys` is protected with `@check_roles(UserRoles.SSHTunnelServer)`.
- `argus/backend/service/user.py` + `argus/backend/controller/auth.py`
  - `ROLE_SSH_TUNNEL_SERVER` is hard-scoped to one allowed operation: `GET /api/v1/client/ssh/keys`.
- `argus/backend/service/tunnel_service.py`
  - Proxy service users are created/reused with only `ROLE_SSH_TUNNEL_SERVER`.

## Tests
Added/updated:
- `argus/backend/tests/tunnel/test_tunnel_service.py`
- `argus/backend/tests/tunnel/test_ssh_api.py`
- `argus/backend/tests/tunnel/test_admin_proxy_tunnel_api.py`

Validation run:
- `uv run pytest`
- Result: **256 passed**

## Design doc sync
Updated `docs/plans/ssh-tunnel-design.md` to match implementation:
- Multi-active host model + round-robin behavior
- Admin config list/toggle APIs
- Active-only semantics for specific config lookup
- Backend auto-discovery of host key fingerprint
- Step status updates including admin API tests

## Not included (follow-up PRs)
- Step 4c: proxy host provisioning template
- Step 4d: Admin panel UI
- Step 5/6: client tunnel module and `base.py` integration
- Step 7/7b: CLI tunnel integration (`argus-client` and Go `argus-cli ssh-keys`)

### DIFF
--- a/argus/backend/controller/ssh_api.py
+++ b/argus/backend/controller/ssh_api.py
@@ -1,0 +1,89 @@
+import logging
+from uuid import UUID
+
+from flask import Blueprint, Response, g, request
+
+from argus.backend.error_handlers import handle_api_exception
+from argus.backend.service.tunnel_service import TunnelService
+from argus.backend.service.user import api_login_required
+
+bp = Blueprint("ssh_api", __name__, url_prefix="/ssh")
+bp.register_error_handler(Exception, handle_api_exception)
+
+LOGGER = logging.getLogger(__name__)
+
+
+@bp.route("/tunnel", methods=["POST"])
+@api_login_required
+def register_tunnel():
+    """
+    Register a client SSH public key and obtain proxy tunnel connection details.
+
+    Request JSON body
+    -----------------
+    public_key  : str  — OpenSSH-format ed25519 (or other) public key (required)
+    ttl_seconds : int  — optional key lifetime in seconds (default 86400 / 24 h)
+    tunnel_id   : str  — optional UUID of a specific ProxyTunnelConfig to use;
+                         defaults to the currently active config
+
+    Response
+    --------
+    {
+        "status": "ok",
+        "response": {
+            "key_id":               "<uuid>",
+            "tunnel_id":            "<uuid>",
+            "proxy_host":           "proxy.example.com",
+            "proxy_port":           22,
+            "proxy_user":           "argus-proxy",
+            "target_host":          "10.0.0.5",
+            "target_port":          8080,
+            "host_key_fingerprint": "SHA256:...",
+            "expires_at":           "2026-04-16T12:00:00Z"
+        }
+    }
+    """
+    payload = request.get_json() or {}
+    public_key = payload.get("public_key")
+    ttl_seconds = payload.get("ttl_seconds")
+    tunnel_id = payload.get("tunnel_id")
+
+    if tunnel_id:
+        try:
+            tunnel_id = UUID(str(tunnel_id))
+        except ValueError as exc:
+            return {"status": "error", "response": {"message": f"Invalid tunnel_id: {exc}"}}, 400
+
+    result = TunnelService().register_tunnel(
+        user=g.user,
+        public_key=public_key,
+        tunnel_id=tunnel_id,
+        ttl_seconds=ttl_seconds,
+    )
+    return {"status": "ok", "response": result}
+
+
+@bp.route("/keys", methods=["GET"])
+@api_login_required
+def get_authorized_keys():
+    """
+    Return all non-expired SSH public keys in OpenSSH ``authorized_keys``
+    format (plain text, one key per line).
+
+    This endpoint is called by the proxy host's ``AuthorizedKeysCommand``
+    (via ``argus-cli ssh-keys``) on every SSH connection attempt.
+
+    Query parameters
+    ----------------
+    tunnel_id : str — optional UUID; restrict to keys for a specific tunnel
+    """
+    tunnel_id_str = request.args.get("tunnel_id")
+    tunnel_id = None
+    if tunnel_id_str:
+        try:
+            tunnel_id = UUID(tunnel_id_str)
+        except ValueError as exc:
+            return {"status": "error", "response": {"message": f"Invalid tunnel_id: {exc}"}}, 400
+
+    keys_text = TunnelService().get_authorized_keys(tunnel_id=tunnel_id)
+    return Response(keys_text, mimetype="text/plain")
--- a/argus/backend/controller/ssh_api.py
+++ b/argus/backend/controller/ssh_api.py
@@ -1,11 +1,18 @@
 import logging
+from dataclasses import asdict
+from typing import TypedDict
 
 from flask import Blueprint, Response, g, request
 
-from argus.backend.error_handlers import APIException
 from argus.backend.error_handlers import handle_api_exception
+from argus.backend.models.web import UserRoles
 from argus.backend.service.tunnel_service import TunnelService
-from argus.backend.service.user import api_login_required
+from argus.backend.service.user import api_login_required, check_roles
+
+
+class RegisterTunnelPayload(TypedDict, total=False):
+    public_key: str
+    ttl_seconds: int
 
 bp = Blueprint("ssh_api", __name__, url_prefix="/ssh")
 bp.register_error_handler(Exception, handle_api_exception)
@@ -23,7 +30,7 @@ def register_tunnel():
     -----------------
     public_key  : str  — OpenSSH-format ed25519 (or other) public key (required)
     ttl_seconds : int  — optional key lifetime in seconds.
-                         Must be within [86400, 2592000] (24h..30d).
+                         Must be within [3600, 2592000] (1h..30d).
                          Default is 86400 (24h).
 
     Response
@@ -42,7 +49,7 @@ def register_tunnel():
         }
     }
     """
-    payload = request.get_json() or {}
+    payload: RegisterTunnelPayload = request.get_json() or {}
     public_key = payload.get("public_key")
     ttl_seconds = payload.get("ttl_seconds")
 
@@ -51,14 +58,14 @@ def register_tunnel():
         public_key=public_key,
         ttl_seconds=ttl_seconds,
     )
-    return {"status": "ok", "response": result}
+    return {"status": "ok", "response": asdict(result)}
 
 
 @bp.route("/tunnel", methods=["GET"])
 @api_login_required
 def get_tunnel_connection():
     result = TunnelService().get_tunnel_connection(proxy_host=request.args.get("proxy_host"))
-    return {"status": "ok", "response": result}
+    return {"status": "ok", "response": asdict(result)}
 
 
 @bp.route("/tunnel/keys", methods=["GET"])
@@ -72,11 +79,12 @@ def get_user_keys():
     """
     tunnel_id = request.args.get("tunnel_id")
     result = TunnelService().list_keys(tunnel_id=tunnel_id, user_id=g.user.id)
-    return {"status": "ok", "response": result}
+    return {"status": "ok", "response": [asdict(row) for row in result]}
 
 
 @bp.route("/keys", methods=["GET"])
 @api_login_required
+@check_roles(UserRoles.SSHTunnelServer)
 def get_authorized_keys():
     """
     Return all non-expired SSH public keys in OpenSSH ``authorized_keys``
@@ -85,8 +93,5 @@ def get_authorized_keys():
     This endpoint is called by the proxy host's ``AuthorizedKeysCommand``
     (via ``argus-cli ssh-keys``) on every SSH connection attempt.
     """
-    if not g.user or not g.user.is_service_user():
-        raise APIException("Only service users can fetch SSH authorized keys")
-
     keys_text = TunnelService().get_authorized_keys()
     return Response(keys_text, mimetype="text/plain")
--- a/argus/backend/controller/ssh_api.py
+++ b/argus/backend/controller/ssh_api.py
@@ -22,7 +22,9 @@ def register_tunnel():
     Request JSON body
     -----------------
     public_key  : str  — OpenSSH-format ed25519 (or other) public key (required)
-    ttl_seconds : int  — optional key lifetime in seconds (default 86400 / 24 h)
+    ttl_seconds : int  — optional key lifetime in seconds.
+                         Must be within [86400, 2592000] (24h..30d).
+                         Default is 86400 (24h).
 
     Response
     --------
--- a/argus/backend/controller/ssh_api.py
+++ b/argus/backend/controller/ssh_api.py
@@ -1,8 +1,8 @@
 import logging
-from uuid import UUID
 
 from flask import Blueprint, Response, g, request
 
+from argus.backend.error_handlers import APIException
 from argus.backend.error_handlers import handle_api_exception
 from argus.backend.service.tunnel_service import TunnelService
 from argus.backend.service.user import api_login_required
@@ -23,8 +23,6 @@ def register_tunnel():
     -----------------
     public_key  : str  — OpenSSH-format ed25519 (or other) public key (required)
     ttl_seconds : int  — optional key lifetime in seconds (default 86400 / 24 h)
-    tunnel_id   : str  — optional UUID of a specific ProxyTunnelConfig to use;
-                         defaults to the currently active config
 
     Response
     --------
@@ -32,7 +30,6 @@ def register_tunnel():
         "status": "ok",
         "response": {
             "key_id":               "<uuid>",
-            "tunnel_id":            "<uuid>",
             "proxy_host":           "proxy.example.com",
             "proxy_port":           22,
             "proxy_user":           "argus-proxy",
@@ -46,20 +43,19 @@ def register_tunnel():
     payload = request.get_json() or {}
     public_key = payload.get("public_key")
     ttl_seconds = payload.get("ttl_seconds")
-    tunnel_id = payload.get("tunnel_id")
-
-    if tunnel_id:
-        try:
-            tunnel_id = UUID(str(tunnel_id))
-        except ValueError as exc:
-            return {"status": "error", "response": {"message": f"Invalid tunnel_id: {exc}"}}, 400
 
     result = TunnelService().register_tunnel(
         user=g.user,
         public_key=public_key,
-        tunnel_id=tunnel_id,
         ttl_seconds=ttl_seconds,
     )
+    return {"status": "ok", "response": result}
+
+
+@bp.route("/tunnel", methods=["GET"])
+@api_login_required
+def get_tunnel_connection():
+    result = TunnelService().get_tunnel_connection()
     return {"status": "ok", "response": result}
 
 
@@ -72,18 +68,9 @@ def get_authorized_keys():
 
     This endpoint is called by the proxy host's ``AuthorizedKeysCommand``
     (via ``argus-cli ssh-keys``) on every SSH connection attempt.
-
-    Query parameters
-    ----------------
-    tunnel_id : str — optional UUID; restrict to keys for a specific tunnel
     """
-    tunnel_id_str = request.args.get("tunnel_id")
-    tunnel_id = None
-    if tunnel_id_str:
-        try:
-            tunnel_id = UUID(tunnel_id_str)
-        except ValueError as exc:
-            return {"status": "error", "response": {"message": f"Invalid tunnel_id: {exc}"}}, 400
+    if not g.user or not g.user.is_service_user():
+        raise APIException("Only service users can fetch SSH authorized keys")
 
-    keys_text = TunnelService().get_authorized_keys(tunnel_id=tunnel_id)
+    keys_text = TunnelService().get_authorized_keys(service_user_id=g.user.id)
     return Response(keys_text, mimetype="text/plain")
--- a/argus/backend/controller/ssh_api.py
+++ b/argus/backend/controller/ssh_api.py
@@ -7,7 +7,7 @@ from flask import Blueprint, Response, g, request
 from argus.backend.error_handlers import handle_api_exception
 from argus.backend.models.web import UserRoles
 from argus.backend.service.tunnel_service import TunnelService
-from argus.backend.service.user import api_login_required, check_roles
+from argus.backend.service.user import allow_ssh_tunnel_server_scope, api_login_required, check_roles
 
 
 class RegisterTunnelPayload(TypedDict, total=False):
@@ -83,6 +83,7 @@ def get_user_keys():
 
 
 @bp.route("/keys", methods=["GET"])
+@allow_ssh_tunnel_server_scope
 @api_login_required
 @check_roles(UserRoles.SSHTunnelServer)
 def get_authorized_keys():
--- a/argus/backend/controller/ssh_api.py
+++ b/argus/backend/controller/ssh_api.py
@@ -57,7 +57,21 @@ def register_tunnel():
 @bp.route("/tunnel", methods=["GET"])
 @api_login_required
 def get_tunnel_connection():
-    result = TunnelService().get_tunnel_connection()
+    result = TunnelService().get_tunnel_connection(proxy_host=request.args.get("proxy_host"))
+    return {"status": "ok", "response": result}
+
+
+@bp.route("/tunnel/keys", methods=["GET"])
+@api_login_required
+def get_user_keys():
+    """
+    Return SSH keys owned by the authenticated user.
+
+    Optional query params:
+    - tunnel_id: UUID of a specific tunnel to scope keys
+    """
+    tunnel_id = request.args.get("tunnel_id")
+    result = TunnelService().list_keys(tunnel_id=tunnel_id, user_id=g.user.id)
     return {"status": "ok", "response": result}
 
 
@@ -74,5 +88,5 @@ def get_authorized_keys():
     if not g.user or not g.user.is_service_user():
         raise APIException("Only service users can fetch SSH authorized keys")
 
-    keys_text = TunnelService().get_authorized_keys(service_user_id=g.user.id)
+    keys_text = TunnelService().get_authorized_keys()
     return Response(keys_text, mimetype="text/plain")
--- a/argus/backend/models/ssh_key.py
+++ b/argus/backend/models/ssh_key.py
@@ -1,0 +1,51 @@
+from datetime import UTC, datetime
+from uuid import uuid4
+
+from cassandra.cqlengine import columns
+from cassandra.cqlengine.models import Model
+
+
+def _utcnow_naive() -> datetime:
+    return datetime.now(tz=UTC).replace(tzinfo=None)
+
+
+class SSHTunnelKey(Model):
+    """
+    Stores a client-registered SSH public key for a specific (user, tunnel) pair.
+
+    Rows are inserted with a ScyllaDB TTL (default 24 h). The ``expires_at``
+    field is informational — it mirrors the TTL so the client knows when to
+    re-register. Actual expiry is handled automatically by ScyllaDB.
+    """
+
+    id = columns.UUID(primary_key=True, default=uuid4)
+    user_id = columns.UUID(required=True, index=True)
+    tunnel_id = columns.UUID(required=True, index=True)
+    public_key = columns.Text(required=True)
+    fingerprint = columns.Text(required=True, index=True)
+    created_at = columns.DateTime(required=True, default=_utcnow_naive)
+    expires_at = columns.DateTime(required=True)
+
+
+class ProxyTunnelConfig(Model):
+    """
+    Stores the configuration of an SSH proxy tunnel server.
+
+    Only one config should have ``is_active=True`` at any time. When
+    ``TunnelService.save_proxy_tunnel_config`` creates a new config it
+    deactivates the previously active one.
+
+    A dedicated Argus service user (``service_user_id``) is created per
+    proxy host so the proxy host can call the authorised-keys API with its
+    own isolated credentials.
+    """
+
+    id = columns.UUID(primary_key=True, default=uuid4)
+    host = columns.Text(required=True)
+    port = columns.Integer(required=True)
+    proxy_user = columns.Text(required=True)
+    target_host = columns.Text(required=True)
+    target_port = columns.Integer(required=True)
+    host_key_fingerprint = columns.Text(required=True)
+    service_user_id = columns.UUID()
+    is_active = columns.Boolean(default=lambda: False, index=True)
--- a/argus/backend/models/ssh_key.py
+++ b/argus/backend/models/ssh_key.py
@@ -22,7 +22,7 @@ class SSHTunnelKey(Model):
     user_id = columns.UUID(required=True, index=True)
     tunnel_id = columns.UUID(required=True, index=True)
     public_key = columns.Text(required=True)
-    fingerprint = columns.Text(required=True, index=True)
+    fingerprint = columns.Text(required=True)
     created_at = columns.DateTime(required=True, default=_utcnow_naive)
     expires_at = columns.DateTime(required=True)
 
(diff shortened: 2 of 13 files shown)


soyacz (commented):
Review from Claude. Add followup issue about this security concern - we already talked about. Could be done in separate PR (new role for getting these keys):
# PR #967 Review — SSH Tunnel Backend (Part 3)

## Overview

Implements server-side foundation for SSH tunnel feature:
- Two models (`SSHTunnelKey`, `ProxyTunnelConfig`) with ScyllaDB TTL.
- `TunnelService` — key register, fingerprint, authorized_keys, config CRUD, service-user provisioning.
- Two client routes: `POST /api/v1/client/ssh/tunnel`, `GET /api/v1/client/ssh/keys`.
- 24 docker_required tests.

Checks pass. Scope matches design doc. Follow-ups (4b/4c/4d, admin API, UI, CLI) deferred.

---

## Issues

### 🔴 Bug — fingerprint not match `ssh-keygen -lf`

`tunnel_service.py:32-38` compute `SHA256(raw_pubkey_bytes)`. OpenSSH fingerprint = `SHA256(OpenSSH wire blob)` (base64-decoded second field of `authorized_keys` line — includes key-type prefix + length frames). For ed25519 raw = 32 B, wire = 51 B. Hashes differ.

Docstring claims: *"matching the output of `ssh-keygen -lf`"* — it does not. Also `PublicFormat.Raw` works only for Ed25519/X25519/Ed448/X448; RSA/ECDSA keys raise `ValueError` even though they're valid SSH keys. So "(or other)" in the docs also wrong.

Fix — parse base64 blob directly:

```python
def _derive_fingerprint(public_key_str: str) -> str:
    try:
        parts = public_key_str.strip().split()
        if len(parts) < 2:
            raise TunnelServiceException("Malformed OpenSSH public key")
        blob = base64.b64decode(parts[1])
        # validate parseability (raises on garbage)
        load_ssh_public_key(public_key_str.strip().encode("utf-8"))
        b64 = base64.b64encode(hashlib.sha256(blob).digest()).rstrip(b"=").decode("ascii")
        return f"SHA256:{b64}"
    except (ValueError, UnsupportedAlgorithm, TypeError, binascii.Error) as exc:
        raise TunnelServiceException(f"Invalid SSH public key: {exc}") from exc
```

Low blast radius now (fingerprint informational), but admin UI in 4d will likely compare against `ssh-keygen` output → mismatch later.

---

### 🔴 Security — `/ssh/keys` exposes all keys to any authenticated user

`ssh_api.py:70` guarded by `@api_login_required` only. Any regular Argus user token can dump every registered public key across all tunnels and all users (fingerprints + raw keys).

Public keys aren't secret, but:
- discloses who has access to which proxy tunnel (user tracking across tunnels via fingerprint).
- discloses active tunnel IDs (enumeration).
- design intent per PR body: called "by the proxy host's `AuthorizedKeysCommand`" — i.e. the dedicated service user, not arbitrary clients.

Add role / service-user check, or bind to `service_user_id` of the target `ProxyTunnelConfig`:

```python
if tunnel_id:
    cfg = ProxyTunnelConfig.get(id=tunnel_id)
    if cfg.service_user_id != g.user.id:
        raise Forbidden("Not authorized for this tunnel")
```

---

### 🟡 DB anti-pattern — secondary index on low-cardinality boolean

`ssh_key.py:51` — `is_active = columns.Boolean(..., index=True)`. Scylla secondary index on 2-value column = every query hits both partitions. Since only one row is expected active, better: a tiny lookup table or a singleton partition key, or just skip the index (full table scan of a handful of rows is cheaper).

Also `_get_active_config` does `filter(is_active=True).all()` then `[0]` — add `.limit(1)` defensively.

---

### 🟡 Query — `SSHTunnelKey.objects.all()` unbounded

`tunnel_service.py:160` — when `tunnel_id` omitted, full-table scan with no limit. Proxy host path should always pass `tunnel_id`; either enforce it server-side or add explicit pagination. Also `list_keys` same issue.

Consider making `(tunnel_id, id)` the primary key since all access paths filter by `tunnel_id` — cheaper than secondary index.

---

### 🟡 `isoformat() + "Z"` fragile

`tunnel_service.py:134, 191-192`. Works only because datetime is naive-UTC. If `expires_at` ever becomes tz-aware (`.replace(tzinfo=None)` dropped), output becomes `...+00:00Z` — invalid ISO-8601. Use explicit format:

```python
expires_at.strftime("%Y-%m-%dT%H:%M:%SZ")
```

Or store aware + call `.isoformat().replace("+00:00", "Z")`.

---

### 🟡 Test isolation — `active_config` fixture leaks state

`test_ssh_api.py:60-69` deactivates *all* existing active configs at setup but never restores them at teardown. Running this test suite in shared DB leaves prior active configs deactivated. Same pattern in several service tests. Either scope to session, or restore `is_active=True` on teardown.

---

### 🟡 Test — `g.user = None` hack

`test_ssh_api.py:166-183`, `242-254` — manipulate `flask.g` from outside a request ctx to force 403. Fragile and depends on fixture-layer pre-authentication. Simpler: just `argus_app.test_client()` without auth fixture, no `g` dance.

---

## Minor

- `tunnel_service.py:316` — `import secrets` and `from argus.backend.service.user import UserService` inside function. If it's for circular-import avoidance, leave a comment; otherwise hoist to top.
- `tunnel_service.py:111` — `int(ttl_seconds) if ttl_seconds else DEFAULT_TTL_SECONDS`. `ttl_seconds=0` silently becomes default. Use `if ttl_seconds is not None`.
- `tunnel_service.py:335` — email `f"{username}@argus.internal"` where username may contain dots from hostname (`proxy-tunnel-proxy.example.com@argus.internal`). Fine for SMTP format but verify `User` model / validation doesn't choke.
- `ssh_key.py:51` — `default=lambda: False` — use `default=False`.
- `models/web.py:17` — `UserRoles` imported but only `.User.value` used — OK, flagging only because it's the only role granted; service user gets full user privileges. Consider a dedicated `ProxyService` role.
- `ssh_api.py:10` — blueprint registers its own `handle_api_exception` error handler; parent `client_api` already does. Redundant but harmless.
- `ssh_api.py:39-59` — UUID validation uses `except ValueError as exc` and echoes `exc` back to client. Small info leak; message `"Invalid tunnel_id"` enough.
- `tunnel_service.py:258` — reactivation race: two concurrent `save_proxy_tunnel_config` calls can both end with `is_active=True`. Not critical for admin-only op, but worth a lock or LWT if scaled.
- `_create_proxy_service_user` reuses existing user by username with no host-collision guard. Two different config entries for the *same host* will share one service user — maybe intended, maybe not; document.

---

## Coverage

14 service + 10 API tests. Good surface. Missing:
- Fingerprint bug caught by no test (no assertion against `ssh-keygen -lf` output).
- RSA/ECDSA key rejection (current code will raise on these — arguably a feature but untested).
- TTL auto-expiry behavior (hard to test without waiting, skip OK).
- Concurrent `save_proxy_tunnel_config` race.

---

## Summary

Solid foundation, tests thorough. Two concrete problems to fix before merging:

1. Fingerprint algorithm wrong → SHA-256 mismatch with `ssh-keygen -lf` (will bite admin UI).
2. `/ssh/keys` under-authorized → any authenticated user can dump all keys/tunnels.

Rest is polish: DB index choice, fragile ISO format, test fixture leaks.

soyacz:
@k0machi please review

argus/backend/controller/ssh_api.py
```diff
@@ -0,0 +1,89 @@
...
+    payload = request.get_json() or {}
+    public_key = payload.get("public_key")
+    ttl_seconds = payload.get("ttl_seconds")
+    tunnel_id = payload.get("tunnel_id")
+
+    if tunnel_id:
+        try:
+            tunnel_id = UUID(str(tunnel_id))
+        except ValueError as exc:
+            return {"status": "error", "response": {"message": f"Invalid tunnel_id: {exc}"}}, 400
```
k0machi: Raise an exception instead, error format is wrong here

argus/backend/controller/ssh_api.py
```diff
@@ -0,0 +1,89 @@
...
+    ----------------
+    tunnel_id : str — optional UUID; restrict to keys for a specific tunnel
+    """
+    tunnel_id_str = request.args.get("tunnel_id")
+    tunnel_id = None
+    if tunnel_id_str:
+        try:
+            tunnel_id = UUID(tunnel_id_str)
+        except ValueError as exc:
+            return {"status": "error", "response": {"message": f"Invalid tunnel_id: {exc}"}}, 400
```
k0machi: Ditto

argus/backend/models/ssh_key.py
```diff
@@ -0,0 +1,51 @@
...
+    Rows are inserted with a ScyllaDB TTL (default 24 h). The ``expires_at``
+    field is informational — it mirrors the TTL so the client knows when to
+    re-register. Actual expiry is handled automatically by ScyllaDB.
+    """
+
+    id = columns.UUID(primary_key=True, default=uuid4)
+    user_id = columns.UUID(required=True, index=True)
+    tunnel_id = columns.UUID(required=True, index=True)
+    public_key = columns.Text(required=True)
+    fingerprint = columns.Text(required=True, index=True)
```
k0machi: Why index a fingerprint?

argus/backend/models/ssh_key.py
```diff
@@ -0,0 +1,51 @@
...
+
+    id = columns.UUID(primary_key=True, default=uuid4)
+    host = columns.Text(required=True)
+    port = columns.Integer(required=True)
+    proxy_user = columns.Text(required=True)
+    target_host = columns.Text(required=True)
+    target_port = columns.Integer(required=True)
+    host_key_fingerprint = columns.Text(required=True)
+    service_user_id = columns.UUID()
+    is_active = columns.Boolean(default=lambda: False, index=True)
```
k0machi: Index on a boolean seems
excessive

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,341 @@
...
+        dict with keys: ``proxy_host``, ``proxy_port``, ``proxy_user``,
+        ``target_host``, ``target_port``, ``host_key_fingerprint``,
+        ``expires_at`` (UTC ISO-8601 string), ``tunnel_id``, ``key_id``.
+        """
+        if not public_key or not public_key.strip():
+            raise TunnelServiceException("public_key is required")
+
+        fingerprint = _derive_fingerprint(public_key)
+
+        if tunnel_id is not None:
```
k0machi: Conceptually, how would a client know about a tunnel id before receiving a key?

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,341 @@
...
+            When supplied, restrict to keys for that tunnel.
+        """
+        if tunnel_id is not None:
+            if not isinstance(tunnel_id, UUID):
+                tunnel_id = UUID(str(tunnel_id))
+            rows = SSHTunnelKey.objects.filter(tunnel_id=tunnel_id).all()
+        else:
+            rows = SSHTunnelKey.objects.all()
+
+        return [
```
k0machi: Unnecessary manual formatting

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,341 @@
...
+            id=uuid4(),
+            user_id=user.id,
+            tunnel_id=config.id,
+            public_key=public_key.strip(),
+            fingerprint=fingerprint,
+            created_at=now_utc,
+            expires_at=expires_at,
+        )
+
+        return {
```
k0machi, line 137: Ditto

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,341 @@
...
+                tunnel_id = UUID(str(tunnel_id))
+            try:
+                config = ProxyTunnelConfig.get(id=tunnel_id)
+                return self._config_to_dict(config)
+            except ProxyTunnelConfig.DoesNotExist:
+                return None
+
+        try:
+            config = self._get_active_config()
+            return self._config_to_dict(config)
```
k0machi: Again, manual dict methods are not needed, unless you want to hide a secret

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,341 @@
...
+    @staticmethod
+    def _create_proxy_service_user(host: str) -> tuple[User, str]:
+        """
+        Create (or re-use) a service ``User`` for the given proxy host and
+        return ``(user, api_token)``.
+
+        The username follows the pattern ``proxy-tunnel-<host>`` to make it
+        easy to identify in the user list.
+        """
+        import secrets
```
k0machi: This import can go to the top

k0machi (changes requested):
In general looks alright, however I found some issues here and there.

argus/backend/service/tunnel_service.py
```diff
@@ -17,6 +17,8 @@
 LOGGER = logging.getLogger(__name__)
 
 DEFAULT_TTL_SECONDS = 86400  # 24 hours
+MIN_TTL_SECONDS = 86400  # 24 hours
```
k0machi, line 20: Why is minimum 24 hours? A lot of our tests are <6 hours long. A majority of ones that run are 10 minute long artifact tests.

argus/backend/controller/ssh_api.py
```diff
@@ -0,0 +1,78 @@
...
+def get_authorized_keys():
+    """
+    Return all non-expired SSH public keys in OpenSSH ``authorized_keys``
+    format (plain text, one key per line).
+
+    This endpoint is called by the proxy host's ``AuthorizedKeysCommand``
+    (via ``argus-cli ssh-keys``) on every SSH connection attempt.
+    """
+    if not g.user or not g.user.is_service_user():
+        raise APIException("Only service users can fetch SSH authorized keys")
```
k0machi, line 89: CheckRoles decorator should be used here. Add a new role that would represent a service user.

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,337 @@
...
+            When supplied, restrict to keys for that tunnel.
+        """
+        if tunnel_id is not None:
+            if not isinstance(tunnel_id, UUID):
+                tunnel_id = UUID(str(tunnel_id))
+            rows = SSHTunnelKey.objects.filter(tunnel_id=tunnel_id).limit(DEFAULT_KEYS_LIST_LIMIT).all()
+        else:
+            rows = SSHTunnelKey.objects.limit(DEFAULT_KEYS_LIST_LIMIT).all()
+
+        return [row._as_dict() for row in rows]
```
k0machi, line 216: Doesn't need `row._as_dict()` either, can just be `row`.

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,337 @@
...
+                tunnel_id = UUID(str(tunnel_id))
+            try:
+                config = ProxyTunnelConfig.get(id=tunnel_id)
+                return config._as_dict()
+            except ProxyTunnelConfig.DoesNotExist:
+                return None
+
+        try:
+            config = self._get_active_config()
+            return config._as_dict()
```
k0machi, line 252: Ditto.

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,337 @@
...
+            port=int(payload["port"]),
+            proxy_user=payload["proxy_user"],
+            target_host=payload["target_host"],
+            target_port=int(payload["target_port"]),
+            host_key_fingerprint=payload["host_key_fingerprint"],
+            service_user_id=service_user.id,
+            is_active=True,
+        )
+
+        result = config._as_dict()
```
k0machi, line 292: Ditto.

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,337 @@
...
+            id=uuid4(),
+            user_id=user.id,
+            tunnel_id=config.id,
+            public_key=public_key.strip(),
+            fingerprint=fingerprint,
+            created_at=now_utc,
+            expires_at=expires_at,
+        )
+
+        return {
```
k0machi, line 137: Still seems excessive.

```python
return {
   **dict(config.items()),
   "key_id": key.id,
}
```

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,337 @@
...
+            "host_key_fingerprint": config.host_key_fingerprint,
+            "expires_at": expires_at,
+        }
+
+    def get_tunnel_connection(self) -> dict:
+        """
+        Return active proxy tunnel connection details for tunnel clients.
+        """
+        config = self._get_active_config()
+        return {
```
k0machi, line 160: can just `return config` here.

k0machi (changes requested):

argus/backend/service/tunnel_service.py
CodeLieutenant: Seemed reasonable to me, but you are right, lets do it like hour, even 10minute runs should create an hour long ssh key

argus/backend/service/tunnel_service.py
CodeLieutenant: Here I'm more concerned if we add something more to the proxy table, I really dont want to leak information, this is why i used a dictionary, just to return exactly what I wanted without creating special DTO

argus/backend/controller/ssh_api.py
CodeLieutenant: I've discussed with @soyacz about this one, and for now we wanted just a regular user, and later in a follow up to design a special role just for SSH Tunnel Server.

Basically to allow it only to run ONE ROUTE - Fetch SSH Public keys. In case of an leak or something going not according to plan.

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,337 @@
...
+        next ``AuthorizedKeysCommand`` call on the proxy host will not include
+        this key.
+        """
+        if not isinstance(key_id, UUID):
+            key_id = UUID(str(key_id))
+        try:
+            key = SSHTunnelKey.get(id=key_id)
+            key.delete()
+        except SSHTunnelKey.DoesNotExist as exc:
+            raise TunnelServiceException(f"SSH key {key_id} not found") from exc
```
soyacz, line 230: I don't think we need to raise - key might be ttl'ed already and this will throw

soyacz (commented):
I'd prefer stricter return types than just `dict` or `list`
With LLM it's easy to generate dataclasses for these and return them.
This approach is more readable and also ease use of these methods (IDE helps) and allows static checks.

k0machi:
Conflicts in the PR - please fix.

k0machi:
> I'd prefer stricter return types than just `dict` or `list` With LLM it's easy to generate dataclasses for these and return them. This approach is more readable and also ease use of these methods (IDE helps) and allows static checks.

Ideally we should just be returning models unless you want to select a couple of fields from there - in that case a TypedDict would work (since they'll be converted to a dict by a serializer anyway)

CodeLieutenant:
> > I'd prefer stricter return types than just `dict` or `list` With LLM it's easy to generate dataclasses for these and return them. This approach is more readable and also ease use of these methods (IDE helps) and allows static checks.
> 
> Ideally we should just be returning models unless you want to select a couple of fields from there - in that case a TypedDict would work (since they'll be converted to a dict by a serializer anyway)

Im updating PR with proper DTOs,only for this thing since it's really security sensitive and adding a special role for this as you suggested

CodeLieutenant:
Updated PR description to reflect the current implemented scope in this branch:
- added Step 4b Admin API endpoints
- added role hardening (`ROLE_SSH_TUNNEL_SERVER`) and strict endpoint scope
- lowered min SSH key TTL to 1 hour
- switched tunnel flow responses to typed DTO-backed payloads
- added admin API tunnel tests
- updated design plan status checkboxes

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,557 @@
...
+    proxy_port: int
+    proxy_user: str
+    target_host: str
+    target_port: int
+    host_key_fingerprint: str
+    expires_at: datetime
+
+
+@dataclass(frozen=True, slots=True)
+class TunnelConnectionResponseDTO:
```
soyacz: this could be a base class for other Tunnel* classes, so we would not repeat common fields

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,557 @@
...
+    tunnel_id: UUID
+    public_key: str
+    fingerprint: str
+    created_at: datetime
+    expires_at: datetime
+
+
+@dataclass(frozen=True, slots=True)
+class ProxyTunnelConfigDTO:
+    id: UUID
```
soyacz, line 69: is `id` the same as `key_id` in other Tunnel field?

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,557 @@
...
+                continue
+            parts = stripped.split()
+            if len(parts) < 3:
+                continue
+            key_type = parts[1]
+            key_data = parts[2]
+            if key_type.startswith("ssh-") or key_type.startswith("ecdsa-"):
+                keys_by_type.setdefault(key_type, f"{key_type} {key_data}")
+
+        preferred_types = (
```
soyacz, line 435: do we need all these? shouldn't we use one, always the same?

soyacz (commented):
Findings
1. Medium: save_proxy_tunnel_config() now treats ssh-keyscan output as the trusted source for host_key_fingerprint (argus/backend/service/tunnel_service.py:349-351, argus/backend/service/tunnel_service.py:400-445). ssh-keyscan is unauthenticated, so if the backend reaches the proxy host over any untrusted segment, a MITM during config creation can inject its own host key and Argus will pin that attacker fingerprint for every future client connection. That removes the explicit trust step the design relied on.
2. Medium: the admin read path mutates round-robin selection state. GET /admin/api/v1/proxy-tunnel/config calls get_proxy_tunnel_config() without a tunnel_id (argus/backend/controller/admin_api.py:379-388), and that delegates to _get_active_config_round_robin() (argus/backend/service/tunnel_service.py:319-321, argus/backend/service/tunnel_service.py:463-486), which persists the next index in RuntimeStore. Any admin refresh, polling, or health check against this endpoint changes which host the next client registration/connection will receive.
3. Medium: tunnel-scoped key queries can silently drop valid rows because list_keys() applies limit(DEFAULT_KEYS_LIST_LIMIT) before the optional tunnel_id filter (argus/backend/service/tunnel_service.py:263-277). Once there are more than 5000 candidate rows in the pre-filter result set, /api/v1/client/ssh/tunnel/keys?tunnel_id=... and any future admin tunnel-scoped listing can return incomplete data depending on row order.
4. Low: _create_proxy_service_user() will reuse any pre-existing account named proxy-tunnel-<host> and rewrite it into the proxy service account (argus/backend/service/tunnel_service.py:499-515). In the collision case, this strips the old roles, marks the user as service_user, and reuses/generates its API token for proxy provisioning. That can silently hijack an unrelated account if the username already exists.

argus/backend/service/tunnel_service.py
CodeLieutenant: Im not actually sure it's a good approach, since it's DTO, duplicating code is better then abstraction since changing it in one place does not break the other

argus/backend/service/tunnel_service.py
CodeLieutenant: Im not sure if we should support only one key type, better to have them multiple, since they are in order of importance, it will in most cases choose first one, and fall back to other if it's not active.

All of these keys are CRYPTO SAFE, and we don't have to worry about security

argus/backend/service/tunnel_service.py
CodeLieutenant: No, here no, this is the tunnel id
Key id only exist in the SSHTunnelDTO

argus/backend/service/user.py
```diff
@@ -23,6 +23,12 @@
 
 LOGGER = logging.getLogger(__name__)
 
+SSH_TUNNEL_SERVER_ALLOWED_PATH = "/api/v1/client/ssh/keys"
+SSH_TUNNEL_SERVER_ALLOWED_METHODS = {"GET", "HEAD", "OPTIONS"}
```
k0machi: This is enforced by endpoint itself, I don't think this is needed

argus/backend/service/user.py
```diff
@@ -23,6 +23,12 @@
...
+SSH_TUNNEL_SERVER_SCOPE_ERROR = (
```
k0machi: I think just returning unauthorized is better than to give hints on how it works.

argus/backend/service/user.py
```diff
@@ -23,6 +23,12 @@
 
 LOGGER = logging.getLogger(__name__)
 
+SSH_TUNNEL_SERVER_ALLOWED_PATH = "/api/v1/client/ssh/keys"
```
k0machi: How about a decorator on the route that stores allowed endpoints in `app` object and then this permission function checks those instead of having to duplicate allowed routes for this server user.

argus/backend/service/user.py
```diff
@@ -459,6 +465,36 @@ def load_logged_in_user():
...
+
+def is_ssh_tunnel_server_request_allowed() -> bool:
+    return request.path == SSH_TUNNEL_SERVER_ALLOWED_PATH and request.method in SSH_TUNNEL_SERVER_ALLOWED_METHODS
+
+
+def is_scoped_ssh_tunnel_server_blocked(user: User | None) -> bool:
+    return is_ssh_tunnel_server_user(user) and not is_ssh_tunnel_server_request_allowed()
+
+
+def enforce_ssh_tunnel_server_scope():
```
k0machi: This could be deferred to login_required I think.

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,567 @@
...
+
+        rows = rows[:DEFAULT_KEYS_LIST_LIMIT]
+
+        return [self._to_ssh_tunnel_key_dto(row) for row in rows]
+
+    def delete_key(self, key_id: UUID | str) -> None:
+        """
+        Delete a single ``SSHTunnelKey`` row.  Takes immediate effect — the
+        next ``AuthorizedKeysCommand`` call on the proxy host will not include
+        this key.
```
k0machi: I feel like this comment is somewhat redundant. Just "Deletes the key" is enough unless you want to have a docstring.

argus/backend/service/tunnel_service.py
k0machi: I agree with @CodeLieutenant, this is fine to support.

argus/backend/service/tunnel_service.py
```diff
@@ -0,0 +1,567 @@
...
+
+        The username follows the pattern ``proxy-tunnel-<host>`` to make it
+        easy to identify in the user list.
+        """
+        username = f"proxy-tunnel-{host}"
+
+        # Re-use existing tunnel service user if it already exists.
+        existing = User.exists_by_name(username)
+        if existing:
+            if not existing.service_user or UserRoles.SSHTunnelServer.value not in (existing.roles or []):
```
k0machi: this can call `check_roles` and `is_service_user`

k0machi (commented):
Some minor comments here in there, primarily in service function and pre-request handler.

argus/backend/service/tunnel_service.py
soyacz: but some params mean the same.
Anyway, it's minor

argus/backend/service/user.py
CodeLieutenant: Didnt think of that, good one

argus/backend/service/user.py
```diff
@@ -459,6 +476,34 @@ def load_logged_in_user():
...
+
+    allowed_endpoint_methods = current_app.extensions.get(SSH_TUNNEL_SERVER_ALLOWED_ENDPOINTS_KEY)
+    if allowed_endpoint_methods is None:
+        allowed_endpoint_methods = set()
+        for rule in current_app.url_map.iter_rules():
+            view = current_app.view_functions.get(rule.endpoint)
+            if getattr(view, "allow_ssh_tunnel_server_scope", False):
+                for method in rule.methods:
+                    allowed_endpoint_methods.add((rule.endpoint, method))
+        current_app.extensions[SSH_TUNNEL_SERVER_ALLOWED_ENDPOINTS_KEY] = allowed_endpoint_methods
```
k0machi: While this function will short circuit, I think it would be better to iterate over this once on app startup, since this cannot change during runtime.

k0machi (approved):
LGTM, minor comment.

argus/backend/service/user.py
CodeLieutenant: Done

CodeLieutenant:
@k0machi @soyacz Pr is fully ready

k0machi (approved):

PR #968, opened 2026-04-15
feature(issues): add product version filtering to issues endpoint and dashboard

pehala:
## Why
I would like to get answer to the question: “What issue were found during investigation of version 2025.1.12” easily, it will help me go over the issues in a reasonable manner, without having to click on each run and gather issues from there.

## Summary
- Add `productVersion` and `includeNoVersion` query params to the `/api/v1/issues/get` endpoint, enabling filtering of issues by the Scylla version of their linked test runs
- Backend centralizes link fetching in `IssueService`, builds a version map by resolving `test_id → plugin_name → plugin model`, then batch-fetches `scylla_version` using `execute_concurrent_with_args` with bounded concurrency
- Frontend passes `productVersion` and `includeNoVersion` through from the Release Dashboard and Views Dashboard to the `<Issues>` component, which re-fetches automatically when either value changes

### Backend
- **`argus/backend/plugins/core.py`** — Added `get_versions_by_run_ids()` classmethod to `PluginModelBase` using `execute_concurrent_with_args(concurrency=50)`
- **`argus/backend/service/issue_utils.py`** — New module with `build_version_map()` (resolves run versions via correct plugin table) and `filter_links_by_version()` helper
- **`argus/backend/service/issue_service.py`** — Refactored `IssueService.get()` to fetch all links once via `_get_links()`, build version map once, filter once, then delegate to `resolve_issues()` on gh/jira services
- **`argus/backend/service/github_service.py`** — Reverted `get_issues()` to original signature, added `resolve_issues()` method that takes pre-filtered links
- **`argus/backend/service/jira_service.py`** — Same pattern as github_service
- **`argus/backend/controller/testrun_api.py`** — Updated `/issues/get` endpoint to accept `productVersion` and `includeNoVersion` query params
- **`argus/backend/util/common.py`** — Moved `check_version()` from `stats.py` to avoid circular imports
- **`argus/backend/service/stats.py`** — Now imports `check_version` from `util.common`

### Frontend
- **`frontend/Github/Issues.svelte`** — Added `productVersion`/`includeNoVersion` props, single `$effect` for reactive re-fetch, version-aware empty state message, removed unused `onMount` import and duplicate inner heading
- **`frontend/ReleaseDashboard/ReleaseDashboard.svelte`** — Passes version props to `<Issues>`, captures `includeNoVersion` from `versionChange` event
- **`frontend/ReleaseDashboard/TestDashboard.svelte`** — Includes `includeNoVersion` in `versionChange` event dispatch
- **`frontend/Views/Widgets/ViewGithubIssues.svelte`** — Passes `productVersion` to `<Issues>`

### Tests
- **`argus/backend/tests/issues/test_issue_version_filter.py`** — 17 parametrized service-level tests covering exact match, prefix match, non-matching version, no filter, `!noVersion` special value, `includeNoVersion` toggle, release-level filtering, for both GitHub and Jira issue types

Fix https://scylladb.atlassian.net/browse/ARGUS-128

### DIFF
--- a/argus/backend/plugins/core.py
+++ b/argus/backend/plugins/core.py
@@ -1,4 +1,5 @@
 import logging
+from collections.abc import Iterable
 from datetime import datetime, UTC
 from math import ceil
 from uuid import UUID
@@ -7,6 +8,7 @@ from time import time
 from cassandra.cqlengine import columns
 from cassandra.cqlengine.models import Model
 from cassandra.cqlengine.usertype import UserType
+from cassandra.concurrent import execute_concurrent_with_args
 from flask import Blueprint
 from argus.backend.db import ScyllaCluster
 from argus.backend.models.plan import ArgusReleasePlan
@@ -192,6 +194,19 @@ class PluginModelBase(Model):
         rows = cluster.session.execute(query=query, parameters=(run_id,))
 
         return list(rows)
+
+    @classmethod
+    def get_versions_by_run_ids(cls, run_ids: Iterable[UUID]) -> dict[UUID, str | None]:
+        """Parallel per-run_id lookups of scylla_version from the plugin table."""
+        cluster = ScyllaCluster.get()
+        query = cluster.prepare(f"SELECT scylla_version FROM {cls.table_name()} WHERE id = ?")
+        params = [(rid,) for rid in run_ids]
+        results = execute_concurrent_with_args(cluster.session, query, params, concurrency=50)
+        return {
+            p[0]: rows.one().get("scylla_version")
+            for p, (success, rows) in zip(params, results)
+            if success and rows
+        }
 
     @classmethod
     def get_run_response(cls, run_id: UUID) -> dict | None:
--- a/argus/backend/service/issue_utils.py
+++ b/argus/backend/service/issue_utils.py
@@ -1,0 +1,69 @@
+import logging
+from collections import defaultdict
+from uuid import UUID
+
+from argus.backend.models.github_issue import IssueLink
+from argus.backend.models.web import ArgusTest
+from argus.backend.plugins.loader import AVAILABLE_PLUGINS
+from argus.backend.util.common import check_version, chunk
+
+LOGGER = logging.getLogger(__name__)
+
+
+def build_version_map(links: list[IssueLink]) -> dict[UUID, str | None]:
+    """Resolve scylla_version for run_ids by looking up only the correct plugin table.
+
+    1. Batch-fetch ArgusTest records to learn each test's plugin_name.
+    2. Group run_ids by plugin_name via link.test_id.
+    3. Query only the primary model for each plugin.
+    """
+    # test_id → plugin_name
+    unique_test_ids = {link.test_id for link in links}
+    test_plugin_map: dict[UUID, str] = {}
+    for batch in chunk(unique_test_ids):
+        for test in ArgusTest.filter(id__in=batch).only(["id", "plugin_name"]).all():
+            test_plugin_map[test.id] = test.plugin_name
+
+    # run_id → plugin model, grouped (deduplicated)
+    runs_by_plugin: dict[str, set[UUID]] = defaultdict(set)
+    for link in links:
+        plugin_name = test_plugin_map.get(link.test_id)
+        if plugin_name:
+            runs_by_plugin[plugin_name].add(link.run_id)
+
+    # Query each plugin's primary model only for its own run_ids
+    version_map: dict[UUID, str | None] = {}
+    for plugin_name, run_ids in runs_by_plugin.items():
+        plugin = AVAILABLE_PLUGINS.get(plugin_name)
+        if not plugin:
+            continue
+        resolved = plugin.model.get_versions_by_run_ids(run_ids)
+        version_map.update(resolved)
+    return version_map
+
+
+def filter_links_by_version(
+    links: list[IssueLink],
+    product_version: str,
+    include_no_version: bool = False,
+    version_map: dict[UUID, str | None] | None = None,
+) -> list[IssueLink]:
+    """Filter IssueLink records to only those whose associated run matches the given version.
+
+    If version_map is not provided, it is built by resolving test_id → plugin_name
+    and querying only the correct plugin table for each run.
+    """
+    if version_map is None:
+        version_map = build_version_map(links)
+
+    def matches(link: IssueLink) -> bool:
+        run_version = version_map.get(link.run_id)
+        if product_version == "!noVersion":
+            return not run_version
+        if check_version(product_version, run_version):
+            return True
+        if include_no_version and not run_version:
+            return True
+        return False
+
+    return [link for link in links if matches(link)]
--- a/frontend/Github/Issues.svelte
+++ b/frontend/Github/Issues.svelte
@@ -140,7 +140,6 @@
 </script>
 
 <script lang="ts">
-    import { onMount } from "svelte";
     import { PLUGIN_NAMES } from "../Common/PluginNames";
     import { newIssueDestinations } from "../Common/IssueDestinations";
     import GithubIssue from "./GithubIssue.svelte";
@@ -152,6 +151,7 @@
     import queryString from 'query-string';
     import JiraIssue from "../Jira/JiraIssue.svelte";
     import { sha1 } from "js-sha1";
+    import { untrack } from "svelte";
     interface Props {
         id?: string;
         runId: any;
@@ -160,6 +160,8 @@
         filter_key?: string;
         submitDisabled?: boolean;
         aggregateByIssue?: boolean;
+        productVersion?: string;
+        includeNoVersion?: boolean;
     }
 
     let {
@@ -169,22 +171,29 @@
         pluginName,
         filter_key = "run_id",
         submitDisabled = false,
-        aggregateByIssue = false
+        aggregateByIssue = false,
+        productVersion = undefined,
+        includeNoVersion = false,
     }: Props = $props();
     let newIssueUrl = $state("");
     let issues: Issue[] = $state([]);
     let fetching = $state(false);
     let showAllLabels = $state(false);
     export const fetchIssues = async function () {
-        issues = [];
         fetching = true;
         try {
-
-            let params = queryString.stringify({
+            let queryParams: Record<string, any> = {
                 filterKey: filter_key,
                 id: id,
                 aggregateByIssue: new Number(aggregateByIssue),
-            }).toString();
+            };
+            if (productVersion) {
+                queryParams.productVersion = productVersion;
+                if (includeNoVersion) {
+                    queryParams.includeNoVersion = 1;
+                }
+            }
+            let params = queryString.stringify(queryParams).toString();
             let apiResponse = await fetch("/api/v1/issues/get?" + params);
             let apiJson = await apiResponse.json();
             if (apiJson.status === "ok") {
@@ -332,8 +341,11 @@
 
     let sortedIssues = $derived(paginateIssues(issues, sortCriteria, reverseSort, filterString));
 
-    onMount(() => {
-        fetchIssues();
+    $effect(() => {
+        // Read both props together as a single dependency so the effect fires
+        // exactly once even when both change in the same Svelte flush.
+        void [productVersion, includeNoVersion];
+        untrack(() => fetchIssues());
     });
 </script>
 
@@ -395,7 +407,6 @@
     <div class="container-fluid mb-2">
         {#if issues.length > 0}
             <h6 class="d-flex">
-                <div>Issues</div>
                 <div class="ms-auto">
                     <IssuesCopyModal sortedIssues={sortedIssues} currentPage={currentPage} selectedLabels={selectedLabels}>
                         <Fa icon={faCopy}/>
@@ -480,11 +491,13 @@
             </div>
         {:else}
             <div class="row">
-                <div class="col text-center text-muted">
+                <div class="col text-center text-muted p-3">
                     {#if fetching}
                         <span class="spinner-border spinner-border-sm"></span> Fetching...
+                    {:else if productVersion}
+                        No issues linked for version {productVersion === "!noVersion" ? "runs without a version" : productVersion}.
                     {:else}
-                        No Issues.
+                        No issues found.
                     {/if}
                 </div>
             </div>
--- a/frontend/Github/Issues.svelte
+++ b/frontend/Github/Issues.svelte
@@ -140,7 +140,6 @@
 </script>
 
 <script lang="ts">
-    import { onMount } from "svelte";
     import { PLUGIN_NAMES } from "../Common/PluginNames";
     import { newIssueDestinations } from "../Common/IssueDestinations";
     import GithubIssue from "./GithubIssue.svelte";
@@ -160,6 +159,8 @@
         filter_key?: string;
         submitDisabled?: boolean;
         aggregateByIssue?: boolean;
+        productVersion?: string;
+        includeNoVersion?: boolean;
     }
 
     let {
@@ -169,7 +170,9 @@
         pluginName,
         filter_key = "run_id",
         submitDisabled = false,
-        aggregateByIssue = false
+        aggregateByIssue = false,
+        productVersion = undefined,
+        includeNoVersion = false,
     }: Props = $props();
     let newIssueUrl = $state("");
     let issues: Issue[] = $state([]);
@@ -179,12 +182,18 @@
         issues = [];
         fetching = true;
         try {
-
-            let params = queryString.stringify({
+            let queryParams: Record<string, any> = {
                 filterKey: filter_key,
                 id: id,
                 aggregateByIssue: new Number(aggregateByIssue),
-            }).toString();
+            };
+            if (productVersion) {
+                queryParams.productVersion = productVersion;
+                if (includeNoVersion) {
+                    queryParams.includeNoVersion = 1;
+                }
+            }
+            let params = queryString.stringify(queryParams).toString();
             let apiResponse = await fetch("/api/v1/issues/get?" + params);
             let apiJson = await apiResponse.json();
             if (apiJson.status === "ok") {
@@ -332,7 +341,10 @@
 
     let sortedIssues = $derived(paginateIssues(issues, sortCriteria, reverseSort, filterString));
 
-    onMount(() => {
+    $effect(() => {
+        // Track reactive props — re-fetch when either changes (and on mount).
+        void productVersion;
+        void includeNoVersion;
         fetchIssues();
     });
 </script>
@@ -395,7 +407,6 @@
     <div class="container-fluid mb-2">
         {#if issues.length > 0}
             <h6 class="d-flex">
-                <div>Issues</div>
                 <div class="ms-auto">
                     <IssuesCopyModal sortedIssues={sortedIssues} currentPage={currentPage} selectedLabels={selectedLabels}>
                         <Fa icon={faCopy}/>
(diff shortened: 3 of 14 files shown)


pehala:
This is a way how to do this without editing schema. It has some perf hit, as filtering on via many queries in batches is not effective, but it is easier to implement and hopefully should not introduce too much latency.

It currently works that the issues accordion is automatically filtered by the version selected (it might make sense to move the accordion under the version filter then), no action required.

I tested manually with dev setup from https://github.com/scylladb/argus/pull/965

soyacz (commented):
Now I have a complete picture. Here's my PR review:

---

# PR #968 Review — `feature(issues): add product version filtering to issues endpoint and dashboard`

## TL;DR

**Recommendation: Request changes.** The feature works and is well-tested for correctness, but the server-side version filter has a **serious N+1 (actually N×M×2) query amplification problem** that will get worse as releases accumulate runs. Combined with a handful of smaller robustness and consistency issues, this should not land as-is.

---

## 🔴 Critical — Query amplification (the N+1-or-worse problem)

### 1. `get_versions_by_run_ids` issues one async query per run_id

`argus/backend/plugins/core.py:196-207`
```python
query = cluster.prepare(f"SELECT id, scylla_version FROM {cls.table_name()} WHERE id = ?")
futures = [(rid, cluster.session.execute_async(query=query, parameters=(rid,))) for rid in run_ids]
```
These queries are parallel (not sequential N+1), which is acceptable on Scylla since a multi-partition `IN` on a partition key is an anti-pattern. **But the compound effect below is the real problem.**

### 2. `filter_links_by_version` queries all plugin models for all run_ids

`argus/backend/service/issue_utils.py:26-32`
```python
for model in all_plugin_models():          # 4 models today: sct, generic, driver_matrix_tests, sirenada
    remaining = [rid for rid in unique_run_ids if rid not in version_map]
    ...
    resolved = model.get_versions_by_run_ids(remaining)
```
Each run_id exists in exactly **one** plugin table, but the code fans out the query to all plugin tables until it happens to resolve them. In the worst case (e.g., all runs in the last-enumerated plugin), this is **4 × N** async queries for nothing to show for the first 3 rounds.

There is no cheap way to know the plugin for a `run_id` without a lookup, **but** we do have `IssueLink.test_id`, and `ArgusTest` carries `plugin_name`. If you resolve `test_id → plugin_name` first (one batched `test_id__in=` query already used in nearby code), you know exactly which model to hit and eliminate 75% of the traffic.

### 3. The same work is done **twice per HTTP request**

`argus/backend/service/issue_service.py:24-37` calls `gh.get_issues(...)` then `jira.get_issues(...)`. Each helper calls `filter_links_by_version` independently on its own `links` list. For a typical release, the underlying run_id set is largely the **same** in both, so you pay the whole query cost twice.

**Fix:** Move `filter_links_by_version` out of the per-service call and apply it **once** in `IssueService.get()` after merging GH+Jira links, or have `IssueService` compute a `version_map` once and pass it down as a predicate.

### Realistic worst-case back-of-envelope

- Release with 500 runs × 4 plugin tables × 2 (gh+jira) ≈ **4,000 parallel CQL round-trips per request**
- Plus the 500 `IssueLink` scan and the existing `chunk()`'d `GithubIssue/JiraIssue.filter(id__in=...)` lookups
- All triggered every time the accordion is opened or the user toggles a version

This will work fine on a lab DB and fall over on production traffic with many views/releases. Scylla will survive it, but:
- Driver concurrency limits can be hit (`MAX_SCHEMA_AGREEMENT_WAIT`/in-flight request limits)
- Python-side memory and event-loop pressure on the Flask worker
- Unnecessary latency for the user

### Suggested remediations (pick one or combine)

1. **(Cheapest, biggest win)** Dedupe run_ids → resolve `test_id → plugin_name` once via `ArgusTest.filter(id__in=batch)`, then only hit the correct plugin table. Keeps the parallel-async pattern.
2. **(Cleanest API)** Move filtering to `IssueService.get()`; compute `version_map` once across combined GH+Jira links.
3. **(Strategic)** Denormalize `scylla_version` onto `IssueLink` at creation time. Version is immutable per run; a tiny migration covers existing rows. This makes filtering a pure in-memory operation with zero extra queries.
4. Use `cassandra.concurrent.execute_concurrent_with_args(...)` to bound concurrency instead of firing N futures at once. Even without the plugin narrowing, this prevents pool exhaustion.

---

## 🟡 Important

### 4. Query-param parsing will 500 on non-numeric input
`argus/backend/controller/testrun_api.py:244`
```python
include_no_version = bool(int(request.args.get("includeNoVersion", 0)))
```
`?includeNoVersion=true` → `int("true")` → `ValueError` → 500. `?includeNoVersion=` → same. The existing `aggregate_by_issue` line above has a `None`-guard but not a non-numeric guard; at least match its shape, and ideally accept `"0"/"1"/"true"/"false"` (this is the public-ish dashboard URL surface).

### 5. One failing plugin table takes down the whole endpoint
`get_versions_by_run_ids`: if any `future.result()` raises (e.g., transient timeout on one plugin table), `filter_links_by_version` bubbles the error and the entire `/issues/get` request fails. Wrap per-model so unresolved run_ids degrade to "no version" rather than failing the endpoint — or at least log which plugin failed.

### 6. `ViewGithubIssues.svelte` doesn't pass `includeNoVersion`
`frontend/Views/Widgets/ViewGithubIssues.svelte:40` — only `productVersion` is forwarded. Views dashboards therefore can't combine "this version" with "also show no-version runs," which is the whole point of `includeNoVersion`. If intentional, add a comment; otherwise plumb it through like `ReleaseDashboard.svelte`.

### 7. PR description disagrees with the code
PR body says *"switched to `IssueService(dry_run=True)` since `get()` only queries the local DB."* The diff keeps plain `IssueService()`. Either revert the description or add the change — `dry_run=True` would skip JIRA/GitHub network clients and is a legitimate perf win here.

### 8. No test for the N+1 behaviour
The test suite (`test_issue_version_filter.py`) is great on correctness (22 parametrized cases + release-level test) but has no assertion about how many queries fire. A test that patches `cluster.session.execute_async` or `get_versions_by_run_ids` and counts calls would catch any regression of the above fix. Worth adding once #1–#3 are resolved.

---

## 🟢 Minor / nits

### 9. `get_versions_by_run_ids` selects `id` redundantly
You already know `rid` from the future tuple — the `SELECT id, scylla_version` and the `row["id"]` round-trip are unnecessary:
```python
query = cluster.prepare(f"SELECT scylla_version FROM {cls.table_name()} WHERE id = ?")
for rid, future in futures:
    rows = future.result()
    if rows:
        result[rid] = rows.one().get("scylla_version")
```
Saves a column fetch per row and simplifies the loop.

### 10. `check_version` import location
`issue_utils.py:6` imports `check_version` from `service.stats`. It's a 6-line pure string util and this is the second place using it. Moving it into `argus/backend/util/` (or a shared `version_utils`) would avoid dragging `stats` as a dependency of the issue pipeline.

### 11. Prefix-match false positives (pre-existing, not introduced here)
`check_version("2023.1", "2023.10.0")` returns `True`. Not your bug, but worth a note in the ticket since the feature now exposes it via URL param.

### 12. Svelte `$effect` with manual previous-value tracking is a smell
`frontend/Github/Issues.svelte:345-353`. The `previousVersion` / `previousIncludeNoVersion` pattern works, but in Svelte 5 the idiomatic way is to either (a) consolidate `fetchIssues` logic into a `$derived` that depends on the reactive inputs, or (b) use `$effect` + `untrack()` for the first run. Low priority; what's there is correct.

### 13. Removed `<div>Issues</div>` heading (`Issues.svelte:415`)
Intentional given the accordion now says "Issues" (changed from "All Issues" in the two dashboards). Confirm this isn't accidental.

### 14. Docstring wording
`get_versions_by_run_ids` says *"Batch-fetch … using async parallel queries"* — the call is parallel but not a single batch; saying "parallel per-run_id lookups" would be more honest and set reviewer expectations correctly.

---

## ✅ What's good

- **Test coverage is excellent** — 22 parametrized cases covering exact match, prefix match, no-match, no filter, `!noVersion`, `includeNoVersion=true/false`, both github and jira subtypes, plus a release-level scenario.
- Helpers `linked_github_issue` / `linked_jira_issue` / `expected_*_issue` / `strip_dynamic` are clean and reusable — good separation of static-shape vs dynamic fields.
- `issue_utils.py` as a dedicated module avoids the circular import that would come from putting this in `issue_service`.
- Frontend prop plumbing through `ReleaseDashboard → TestDashboard → Issues` is consistent with existing patterns.
- The empty-state message now surfaces the active version filter — a nice UX touch.

---

## Recommended action

1. **Must-fix before merge:** Items `#1`–`#3` (the query amplification). Even #1 alone (resolving plugin via `test_id`) would be a big win.
2. **Should-fix:** `#4` (param parsing robustness), `#5` (partial-failure handling), #6 (`includeNoVersion` in ViewGithubIssues), `#7` (align code/description on `dry_run`).
3. **Nice-to-have:** `#8` (query-count test once the fix lands), `#9`–`#14`.

Happy to review a follow-up once `#1`–`#3` are addressed — the test suite you've built already gives you a safety net to refactor the query pattern confidently.

argus/backend/plugins/core.py
```diff
@@ -193,6 +193,19 @@ def get_run_meta_by_run_id(cls, run_id: UUID | str):
 
         return list(rows)
 
+    @classmethod
+    def get_versions_by_run_ids(cls, run_ids: list[UUID]) -> dict[UUID, str | None]:
+        """Batch-fetch scylla_version for a list of run_ids using async parallel queries."""
+        cluster = ScyllaCluster.get()
+        query = cluster.prepare(f"SELECT id, scylla_version FROM {cls.table_name()} WHERE id = ?")
```
pehala: do all plugins have to have scylla_version?

pehala:
> ### 1. `get_versions_by_run_ids` issues one async query per run_id
> 
> `argus/backend/plugins/core.py:196-207`
> 
> ```python
> query = cluster.prepare(f"SELECT id, scylla_version FROM {cls.table_name()} WHERE id = ?")
> futures = [(rid, cluster.session.execute_async(query=query, parameters=(rid,))) for rid in run_ids]
> ```
> 
> These queries are parallel (not sequential N+1), which is acceptable on Scylla since a multi-partition `IN` on a partition key is an anti-pattern. **But the compound effect below is the real problem.**

I think this one is fine and expected, but I am open to suggestions how to resolve in a better way

> ### 2. `filter_links_by_version` queries all plugin models for all run_ids
> 
> `argus/backend/service/issue_utils.py:26-32`
> 
> ```python
> for model in all_plugin_models():          # 4 models today: sct, generic, driver_matrix_tests, sirenada
>     remaining = [rid for rid in unique_run_ids if rid not in version_map]
>     ...
>     resolved = model.get_versions_by_run_ids(remaining)
> ```
> 
> Each run_id exists in exactly **one** plugin table, but the code fans out the query to all plugin tables until it happens to resolve them. In the worst case (e.g., all runs in the last-enumerated plugin), this is **4 × N** async queries for nothing to show for the first 3 rounds.

Real issue, but dependening on ordering of `all_plugin_models()` it can also be a false alarm, there is a break, so it will very rarely achieve the worst case. In tries the remaining ids only, so real world impact shouldnt be that large

> ### 3. The same work is done **twice per HTTP request**
> 
> `argus/backend/service/issue_service.py:24-37` calls `gh.get_issues(...)` then `jira.get_issues(...)`. Each helper calls `filter_links_by_version` independently on its own `links` list. For a typical release, the underlying run_id set is largely the **same** in both, so you pay the whole query cost twice.

Yes, this needs to be fixed

> ## 🟡 Important
> ### 4. Query-param parsing will 500 on non-numeric input
> 
> `argus/backend/controller/testrun_api.py:244`
> 
> ```python
> include_no_version = bool(int(request.args.get("includeNoVersion", 0)))
> ```
> 
> `?includeNoVersion=true` → `int("true")` → `ValueError` → 500. `?includeNoVersion=` → same. The existing `aggregate_by_issue` line above has a `None`-guard but not a non-numeric guard; at least match its shape, and ideally accept `"0"/"1"/"true"/"false"` (this is the public-ish dashboard URL surface).

Will fix

> ### 5. One failing plugin table takes down the whole endpoint
> 
> `get_versions_by_run_ids`: if any `future.result()` raises (e.g., transient timeout on one plugin table), `filter_links_by_version` bubbles the error and the entire `/issues/get` request fails. Wrap per-model so unresolved run_ids degrade to "no version" rather than failing the endpoint — or at least log which plugin failed.

I would log The whole endpoint is for every plugin, so I would rather fail than silently didnt include issues from failing plugin.

> ### 6. `ViewGithubIssues.svelte` doesn't pass `includeNoVersion`
> 
> `frontend/Views/Widgets/ViewGithubIssues.svelte:40` — only `productVersion` is forwarded. Views dashboards therefore can't combine "this version" with "also show no-version runs," which is the whole point of `includeNoVersion`. If intentional, add a comment; otherwise plumb it through like `ReleaseDashboard.svelte`.

Will check if needed

> ### 7. PR description disagrees with the code
> 
> PR body says _"switched to `IssueService(dry_run=True)` since `get()` only queries the local DB."_ The diff keeps plain `IssueService()`. Either revert the description or add the change — `dry_run=True` would skip JIRA/GitHub network clients and is a legitimate perf win here.

Fix

> ### 8. No test for the N+1 behaviour
> 
> The test suite (`test_issue_version_filter.py`) is great on correctness (22 parametrized cases + release-level test) but has no assertion about how many queries fire. A test that patches `cluster.session.execute_async` or `get_versions_by_run_ids` and counts calls would catch any regression of the above fix. Worth adding once #1–#3 are resolved.

Wont add it, asserting number of queries is not a good test, it will be extremely fragile

> ## 🟢 Minor / nits
> ### 9. `get_versions_by_run_ids` selects `id` redundantly
> 
> You already know `rid` from the future tuple — the `SELECT id, scylla_version` and the `row["id"]` round-trip are unnecessary:
> 
> ```python
> query = cluster.prepare(f"SELECT scylla_version FROM {cls.table_name()} WHERE id = ?")
> for rid, future in futures:
>     rows = future.result()
>     if rows:
>         result[rid] = rows.one().get("scylla_version")
> ```
> 
> Saves a column fetch per row and simplifies the loop.

Will fix

> ### 10. `check_version` import location
> 
> `issue_utils.py:6` imports `check_version` from `service.stats`. It's a 6-line pure string util and this is the second place using it. Moving it into `argus/backend/util/` (or a shared `version_utils`) would avoid dragging `stats` as a dependency of the issue pipeline.

Will fix

> ### 12. Svelte `$effect` with manual previous-value tracking is a smell
> 
> `frontend/Github/Issues.svelte:345-353`. The `previousVersion` / `previousIncludeNoVersion` pattern works, but in Svelte 5 the idiomatic way is to either (a) consolidate `fetchIssues` logic into a `$derived` that depends on the reactive inputs, or (b) use `$effect` + `untrack()` for the first run. Low priority; what's there is correct.

Will investigate more

> ### 13. Removed `<div>Issues</div>` heading (`Issues.svelte:415`)
> 
> Intentional given the accordion now says "Issues" (changed from "All Issues" in the two dashboards). Confirm this isn't accidental.

Intentional, as it no longer is All issues

> ### 14. Docstring wording
> 
> `get_versions_by_run_ids` says _"Batch-fetch … using async parallel queries"_ — the call is parallel but not a single batch; saying "parallel per-run_id lookups" would be more honest and set reviewer expectations correctly.

Sure why not

pehala:
Overall the best option in the long term would be to have the binding be done on database level, but that is not something that I can do with any degree of confidence and it would be more work. I would rather improve this and have something

argus/backend/plugins/core.py
soyacz: looks we have in all the plugins, but it's not in `PluginModelBase` - so this is not safe.
btw. we have also `product_version` - @k0machi do you remember how this is used?

pehala:
Fixed according to review comments

soyacz (commented):
We could test it on staging - I don't see changes in models so should be safe.
If we face perf issues, we could consider storing product_version in `IssuesLink` table (and migrate). This way we could filter issues links that contain specific version without querying each run id.

besides, when posting such changes, I'd prefer to post plan first (with justification why we need this)

pehala:
> We could test it on staging - I don't see changes in models so should be safe. If we face perf issues, we could consider storing product_version in `IssuesLink` table (and migrate). This way we could filter issues links that contain specific version without querying each run id.
> 
> besides, when posting such changes, I'd prefer to post plan first (with justification why we need this)

Agree, definitely needs testing on staging.
Regarding justification I added it to the PR description, it is taken from the issue. TLDR is:
_I would like to get answer to the question: “What issue were found during investigation of version 2025.1.12” easily, it will help me go over the issues in a reasonable manner, without having to click on each run and gather issues from there._

Regarding plan, I can add it for the next time, would you like me to add it here as well?

soyacz:
> > We could test it on staging - I don't see changes in models so should be safe. If we face perf issues, we could consider storing product_version in `IssuesLink` table (and migrate). This way we could filter issues links that contain specific version without querying each run id.
> > besides, when posting such changes, I'd prefer to post plan first (with justification why we need this)
> 
> Agree, definitely needs testing on staging. Regarding justification I added it to the PR description, it is taken from the issue. TLDR is: _I would like to get answer to the question: “What issue were found during investigation of version 2025.1.12” easily, it will help me go over the issues in a reasonable manner, without having to click on each run and gather issues from there._
> 
> Regarding plan, I can add it for the next time, would you like me to add it here as well?

No need for that here, but just for the future - I prefer to review the plans - this is the spot where human review has the biggest impact. Code review is pretty good with AI then (and checking if code is according to plan)

pehala:
v3:
  * Rebased
  *  Added small frontend tests

argus/backend/plugins/core.py
pehala: If it is safe for now, can we keep it like this and add it to PluginModelBase in a followup?

frontend/Github/Issues.svelte
```diff
@@ -169,22 +171,29 @@
...
                 aggregateByIssue: new Number(aggregateByIssue),
-            }).toString();
+            };
+            if (productVersion) {
+                queryParams.productVersion = productVersion;
+                if (includeNoVersion) {
+                    queryParams.includeNoVersion = 1;
+                }
+            }
+            let params = queryString.stringify(queryParams).toString();
```
k0machi, line 196: toString isn't needed after stringify.

frontend/ReleaseDashboard/TestDashboard.svelte
```diff
@@ -375,8 +375,7 @@
         return [...enterpriseVersions, ...ossVersions, ...unrecognizedVersions];
     };
 
-    const fetchImages = async function() {
-        let response = await fetch(PANEL_MODES[dashboardObjectType].imagesRoute(dashboardObject));
+    const fetchImages = async function() {        let response = await fetch(PANEL_MODES[dashboardObjectType].imagesRoute(dashboardObject));
```
k0machi, line 378: Two statements on the same line

argus/backend/service/issue_utils.py
```diff
@@ -0,0 +1,69 @@
...
+        if plugin_name:
+            runs_by_plugin[plugin_name].add(link.run_id)
+
+    # Query each plugin's primary model only for its own run_ids
+    version_map: dict[UUID, str | None] = {}
+    for plugin_name, run_ids in runs_by_plugin.items():
+        plugin = AVAILABLE_PLUGINS.get(plugin_name)
+        if not plugin:
+            continue
+        resolved = plugin.model.get_versions_by_run_ids(run_ids)
```
k0machi, line 40: Have you checked the performance of this? This to me seems like it would execute something like 5000 queries (in concurrency of 50) on current `scylla-master`. If this endpoint behaviour is not the default and user has to specifically click something for it - then it is fine I guess. We can check on staging.

k0machi (commented):
Some minor comments

argus/backend/service/issue_utils.py
pehala: I did not, I do not have anything close to a production level of setup on my machine to verify, we would need staging for that. My expectation is that it scylla is built exactly this kind of amount queries and that async queries will make sure it doesnt block backend. 

The filtering happens only when we you change version which you filter and should not be on every opening of dashboard. 

I would try it out on staging and if we encounter provlem we can revert and redesign

soyacz:
@k0machi can you take it to staging?

k0machi:
Deployed on staging.

k0machi (approved):

pehala:
There are improvement to be made here (loading wheel seems to be the bgigest one) but on staging we verified that this is no causing any performance loss, it was slow before, it is slow now :D I would merge and create followups for the enhacements

PR #969, opened 2026-04-16
feat(cli): add ssh tunnel commands and strict cloudflare bypass

CodeLieutenant:
Implement Step 7b from docs/plans/ssh-tunnel-design by adding hidden argus ssh commands for key listing, key registration, and long-running tunnel connection using /api/v1/client/ssh/keys and /api/v1/client/ssh/tunnel. The connect flow now finds a free local port, enforces StrictHostKeyChecking=yes, and rotates ssh-keygen-managed ed25519 keys when the argus-expiry comment indicates expiration.

Cover the private-network CLI path by adding --disable-cloudflare and ARGUS_DISABLE_CLOUDFLARE to fully bypass cloudflared, CF cookies, and CF access headers while honoring ARGUS_AUTH_TOKEN (with ARGUS_TOKEN fallback). Add focused unit tests for command wiring, key lifecycle/expiry handling, API payloads, tunnel argument generation/readiness, and cloudflare bypass behavior.


cli/cmd/ssh.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	sshConnectCmd = &cobra.Command{
+		Use:   "connect",
+		Short: "Start local SSH tunnel and print local port",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			cmd.SilenceUsage = true
+			ctx := cmd.Context()
+			log := logging.For(LoggerFrom(ctx), "ssh-connect")
+			client := APIClientFrom(ctx)
```
k0machi, line 83: isn't this out of scope for this client? The argus python client should handle setting up ssh tunnels.

CodeLieutenant: Yes, it will do it as well, but i want something to test with, and why not to have it there, there is no reason not to support everything.

Also this is hidden command

k0machi: Okay. This can stay then.

soyacz:
Findings:
1. High: cli/internal/services/ssh.go:268-283 turns on StrictHostKeyChecking=yes, but the CLI never writes a dedicated known_hosts file or passes -o UserKnownHostsFile=.... The API model now includes HostKeyFingerprint in cli/internal/models/ssh.go:12-21, and the design doc explicitly says the client should generate a known_hosts entry from that fingerprint. As shipped, first connection on any machine that has not already trusted cfg.ProxyHost in the user/global known_hosts will fail with host key verification errors, so the tunnel setup path is broken for fresh environments and the backend-provided fingerprint is unused.
2. Medium: cli/cmd/ssh.go:33-49 adds ssh keys list without the SkipAuthRetryAnnotation, so wrapAllCommandsWithAuthRetry in cli/cmd/root.go:329-349 wraps it in the normal interactive re-auth flow. In the proxy host AuthorizedKeysCommand use case from the design doc, there is no TTY/browser available. If the token is missing or expired, runWithAuthRetry (cli/cmd/root.go:269-327) will attempt the Cloudflare login path instead of failing fast, which can hang or time out the SSH handshake instead of returning a clean non-zero exit to sshd.

k0machi:
Merge conflicts here as well.

k0machi (approved):
LGTM, but I am not 100% sure what ssh-tunnel command in CLI would be used for, aside from testing. Maybe this could be used for the proposed driver matrix rework.

soyacz (approved):

PR #970, opened 2026-04-16
fix(cli): prevent auth headers on cross-host log-download redirects

CodeLieutenant:
## Summary
- scope API client redirect auth re-attachment to same-host redirects only, so credentials are preserved for Argus/Cloudflare internal hops but not forwarded to cross-host targets
- explicitly drop auth headers/cookies for cross-host redirects to avoid leaking credentials and to prevent AWS S3 presigned download failures caused by mixed auth mechanisms
- add a regression test for `DoStream` that simulates Argus redirecting to a different host and asserts no `Authorization`, `Cookie`, or CF Access service headers reach the redirected request

## Why
Log download endpoints return a 302 to a presigned S3 URL. The CLI was re-attaching `Authorization` and cookies on every redirect, which caused S3 to reject the request with `400 InvalidArgument` ("Only one auth mechanism allowed").

## Validation
- `go test ./internal/api`
- `go test ./cmd`

## Notes
This keeps existing behavior for same-host redirects while making cross-host redirects safe and compatible with presigned object downloads.

### DIFF
--- a/cli/internal/api/api.go
+++ b/cli/internal/api/api.go
@@ -10,6 +10,7 @@ import (
 	"mime"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/scylladb/argus/cli/internal/models"
@@ -137,15 +138,9 @@ func New(rawBaseURL string, opts ...ClientOption) (*Client, error) {
 		o(c)
 	}
 
-	// Install a redirect policy that re-attaches auth credentials on every
-	// hop.  Go's default http.Client strips cookies set via Request.AddCookie
-	// and drops the Authorization header on cross-origin redirects, which
-	// breaks Cloudflare Access: CF responds with a 302 to its interstitial
-	// page, the redirect loses the CF_Authorization cookie / API token, and
-	// the CLI ends up with an HTML login page instead of JSON.
-	//
-	// This must be set after applying opts so that WithHTTPClient (used by
-	// tests) can override the entire http.Client including CheckRedirect.
+	// Install a redirect policy that re-attaches auth credentials on same-host
+	// redirects only. Cross-host redirects (e.g. to presigned S3 URLs) must not
+	// carry Authorization/Cookie headers.
 	if c.httpClient.CheckRedirect == nil {
 		c.httpClient.CheckRedirect = c.reattachAuthOnRedirect
 	}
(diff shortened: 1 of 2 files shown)


soyacz (approved):

k0machi:
Again, I will mention this since this was ignored on the previous PR, but you should not need to reattach cookies at all on non-same origin redirects - As you can see, this breaks other websites.

cli/internal/api/api.go
```diff
@@ -214,19 +209,25 @@ func (c *Client) attachAuth(req *http.Request) {
...
 func (c *Client) reattachAuthOnRedirect(req *http.Request, via []*http.Request) error {
 	if len(via) >= 10 {
 		return errors.New("stopped after 10 redirects")
 	}
-	c.attachAuth(req)
+
+	if strings.EqualFold(req.URL.Host, c.baseURL.Host) {
+		c.attachAuth(req)
+		return nil
+	}
```
k0machi, line 222: Again, this process is usually automatic.

k0machi (approved):

soyacz:
AI findings:
1. Medium: cli/internal/api/api.go:216 treats req.URL.Host and c.baseURL.Host as raw strings, so a redirect from https://argus.example.com/... to https://argus.example.com:443/... is misclassified as cross-host. In deployments where a proxy emits absolute Location headers with an explicit default port, this path will drop Authorization and cookies even though the redirect stays on the same Argus endpoint, so the Cloudflare auth failure this PR is fixing can still happen. The host check needs to compare normalized hostname + effective port, not the verbatim Host strings.

I think it's not relevant for us. Can be fixed in followup. Merging

k0machi:
> AI findings:
> 
> 1. Medium: cli/internal/api/api.go:216 treats req.URL.Host and c.baseURL.Host as raw strings, so a redirect from https://argus.example.com/... to https://argus.example.com:443/... is misclassified as cross-host. In deployments where a proxy emits absolute Location headers with an explicit default port, this path will drop Authorization and cookies even though the redirect stays on the same Argus endpoint, so the Cloudflare auth failure this PR is fixing can still happen. The host check needs to compare normalized hostname + effective port, not the verbatim Host strings.
> 
> I think it's not relevant for us. Can be fixed in followup. Merging

This has to be fixed by removing those parts, you are implementing same thing the go http client should be already doing.

soyacz:
> > AI findings:
> > 
> > 1. Medium: cli/internal/api/api.go:216 treats req.URL.Host and c.baseURL.Host as raw strings, so a redirect from https://argus.example.com/... to https://argus.example.com:443/... is misclassified as cross-host. In deployments where a proxy emits absolute Location headers with an explicit default port, this path will drop Authorization and cookies even though the redirect stays on the same Argus endpoint, so the Cloudflare auth failure this PR is fixing can still happen. The host check needs to compare normalized hostname + effective port, not the verbatim Host strings.
> > 
> > I think it's not relevant for us. Can be fixed in followup. Merging
> 
> This has to be fixed by removing those parts, you are implementing same thing the go http client should be already doing.

@CodeLieutenant please followup

PR #966, opened 2026-04-14
test(frontend): bootstrap Vitest and add utility function tests

pehala:
## Summary
- Bootstrap Vitest test infrastructure for the frontend (previously had zero test coverage)
- Add 82 unit tests across 8 test files covering all pure JS/TS utility modules
- Add `frontend-test` CI job that runs in parallel with existing backend tests

### Infrastructure
- `vitest` + `jsdom` added as devDependencies
- `vitest.config.js` created with jsdom environment and `frontend/**/*.test.{js,ts}` discovery
- `test` and `test:watch` scripts added to `package.json`
- `frontend-test` job appended to `.github/workflows/test.yml` (Node 22, `yarn install --frozen-lockfile`, `yarn test`)

### Test coverage
| Test file | Module | Functions tested |
|-----------|--------|-----------------|
| `TextUtils.test.js` | `Common/TextUtils.js` | `titleCase`, `lightenDarkenColor`, `determineLuma`, `subUnderscores`, `sanitizeSelector` |
| `DateUtils.test.js` | `Common/DateUtils.js` | `timestampToISODate`, `generateWeeklyScheduleDate/EndDate`, `permanentScheduleStart/End`, `startDate`, `endDate` |
| `RunUtils.test.js` | `Common/RunUtils.js` | All 7 package finders, `extractBuildNumber`, `filterInvestigated`, `filterNotInvestigated`, `createScreenshotUrl` |
| `StateManagement.test.js` | `Common/StateManagement.js` | `stateEncoder`, `stateDecoder` |
| `SelectUtils.test.js` | `Common/SelectUtils.js` | `filterUser` (unicode normalization, null handling) |
| `UserUtils.test.js` | `Common/UserUtils.js` | `getPicture`, `getUser` (ghost user fallback) |
| `TeamUtils.test.ts` | `Teams/TeamUtils.ts` | `createUserSelectList`, `TeamManagerAPIError` |
| `ArgusNotification.test.ts` | `Common/ArgusNotification.ts` | `from_json` factories, enum value contracts |

### Notable observations surfaced by tests
- `subUnderscores` uses `.replace()` not `.replaceAll()`, so only the first underscore is replaced (documented in test)
- `lightenDarkenColor` has inconsistent `Math.min` arity across channels (documented in test)

PR #971, opened 2026-04-16
fix: Fix Jira rendering

pehala:
### Summary

Currently, Jira issues are only semi-handled on frontend. For example, on dashboard they are rendered without title and with wrong icon like this:
before

After the patch it looks like this
after

This should be "bugfix" mostly, so I did not write a plan. It is fixing existing functionality rather than drastically changing or adding new stuff

### Changes

* Add JIRA information for similar events and graphed stats
   * This one might be removed, and it can be frontend only fix as it is more complex to verify correctness
* Unify frontend handling of icon + summary via new IssueBadge library.
   * Future refactors can unify it even futher with GithubIssue and JiraIssue component
   * Jira icon is blue, because it looks better  
* Add test for utils functionality (Requires https://github.com/scylladb/argus/pull/966)

### DIFF
--- a/argus/backend/plugins/sct/service.py
+++ b/argus/backend/plugins/sct/service.py
@@ -12,6 +12,7 @@ from flask import current_app, g
 from cassandra.util import uuid_from_time
 from argus.backend.db import ScyllaCluster
 from argus.backend.models.github_issue import GithubIssue, IssueLink
+from argus.backend.models.jira import JiraIssue
 from argus.backend.models.web import ArgusEventTypes, ErrorEventEmbeddings, CriticalEventEmbeddings
 from argus.backend.models.argus_ai import SCTErrorEventEmbedding, SCTCriticalEventEmbedding
 from argus.backend.plugins.sct.testrun import SCTEvent, SCTEventSeverity, SCTJunitReports, SCTTestRun, SubtestType, SCTUnprocessedEvent, StressCommand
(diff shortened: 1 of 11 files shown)


k0machi:
All enumerations and types already exist on the frontend, there is no need to duplicate them (they can be imported) and you can dispatch based on them: https://github.com/scylladb/argus/blob/master/frontend/Github/Issues.svelte#L2-L138

I would also avoid mangling different issue types on backend and instead leave it to the frontend to dynamically dispatch appropriate issue container (as can be seen in that component).

k0machi:
Also check `Issues.svelte` and `argus.css` for common jira styles instead of duplicating them.

k0machi:
> I would also avoid mangling different issue types on backend and instead leave it to the frontend to dynamically dispatch appropriate issue container (as can be seen in that component).

Ah never mind, that was for parts missing the Jira issue fetch, not for rendering issues on frontend.

pehala:
> > I would also avoid mangling different issue types on backend and instead leave it to the frontend to dynamically dispatch appropriate issue container (as can be seen in that component).
> 
> Ah never mind, that was for parts missing the Jira issue fetch, not for rendering issues on frontend.

Sidenote: I think we should refactor Jira+Github handling into a generic class, so if we ever add more tracking tools (not likely but we should still be prepared for) or links in general, it will easy to add them. Right now you would have to edit all usages on backend (as we seen, they are a few)

pehala:
There is more duplication there :/ I will redo the PR probably as the refactor with proper plan, since the simple "fix" is still 300 lines across multiple files

pehala:
Closing in favour of https://github.com/scylladb/argus/pull/973

PR #973, opened 2026-04-17
docs(plans): add plan for unifying GitHub and Jira issue support

pehala:
## Context

Argus tracks issues from two providers: GitHub and Jira. Both are modeled in the database
(`GithubIssue`, `JiraIssue`) and linked to test runs via a shared `IssueLink` table. A unified
`IssueService` facade dispatches CRUD operations to the correct provider based on the URL.

However, several backend code paths and most of the frontend were originally written for GitHub only,
with Jira support bolted on later. This has resulted in silent data loss (Jira issues dropped from
query results), broken rendering (Jira issues passed to GitHub-only components that access
nonexistent fields), and ~350 lines of copy-pasted code between two nearly identical Svelte
components.

### DIFF
--- a/docs/plans/unify-issue-tracking.md
+++ b/docs/plans/unify-issue-tracking.md
@@ -24,23 +24,22 @@ issue IDs from `IssueLink`. Any Jira issues linked to runs in these contexts are
 they exist in the database but never appear in the API response.
 
 The Jira URL parser (`service/jira_service.py:83`) uses a regex hardcoded to
-`scylladb.atlassian.net`, rejecting issue URLs from any other Atlassian Cloud instance.
+`scylladb.atlassian.net`, even though the Jira server URL is already configurable via
+`config["JIRA_SERVER"]` (line 38). The regex should be derived from the configured server URL
+instead of being hardcoded.
 
-### Frontend: Release dashboard renders Jira issues as broken GitHub cards
+### Frontend: Dead code and broken rendering in release dashboard area
 
-The release dashboard has a complete rendering pipeline that is GitHub-only:
+`ReleaseGithubIssues.svelte` and `TestWithIssuesCard.svelte` form a GitHub-only rendering pipeline
+(`ReleaseGithubIssues` → `TestWithIssuesCard` → `GithubIssue`), but they are **dead code** — not
+imported or used anywhere. The actual release dashboard (`ReleaseDashboard.svelte:87`) uses the
+generic `<Issues>` component which properly handles both subtypes. These dead files should be
+deleted.
 
-1. `ReleaseGithubIssues.svelte` fetches from `/api/v1/release/issues` — the backend returns **both**
-   GitHub and Jira issues (confirmed: `stats.py:181` queries both `GithubIssue` and `JiraIssue`).
-2. Issues are passed to `TestWithIssuesCard.svelte` (line 76), which unconditionally renders every
-   issue via `<GithubIssue {issue} />` (line 13).
-3. `GithubIssue.svelte` accesses `issue.repo`, `issue.number`, `issue.owner`, `issue.title`,
-   `issue.url` — none of which exist on Jira issues (which use `key`, `summary`, `project`,
-   `permalink`). The result is broken cards showing `undefined/undefined#undefined`.
-
-The same problem exists in `TestPopoutSelector.svelte` (lines 164-185), which hardcodes the GitHub
-icon and renders `{issue.owner}/{issue.repo}#{issue.number}` — producing `undefined/undefined#undefined`
-for Jira issues.
+However, `TestPopoutSelector.svelte` (lines 164-185) **is** actively used (by
+`ReleaseDashboard.svelte:123` and `ViewTestDashboard.svelte:32`) and has the same problem: it
+hardcodes the GitHub icon and renders `{issue.owner}/{issue.repo}#{issue.number}` — producing
+`undefined/undefined#undefined` for Jira issues.
 
 ### Frontend: Graphed stats badges assume GitHub shape
 
@@ -79,7 +78,7 @@ hardcoded to GitHub fields. Other places that need compact issue display (e.g.,
 ### Frontend: Misleading "Github" names
 
 Several files are named "Github" but handle (or should handle) both types:
-`ViewGithubIssues.svelte`, `ReleaseGithubIssues.svelte`, `ViewTypes.js` key `githubIssues`. Shared
+`ViewGithubIssues.svelte`, `ViewTypes.js` key `githubIssues`. Shared
 type definitions and helpers live in `frontend/Github/Issues.svelte`.
 
 ---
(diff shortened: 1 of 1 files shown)


pehala:
Sidenote: creating this plan required more work than for SCT, @soyacz what about migrating the writing-plan skill from SCT to make it more unified and easier to write?

docs/plans/unify-issue-tracking.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Backend: Jira issues silently dropped in two code paths
+
+The SCT similar-runs feature (`plugins/sct/service.py:700-704`) and the graphed stats widget
+(`service/views_widgets/graphed_stats.py:110-114`) only query the `GithubIssue` table when resolving
+issue IDs from `IssueLink`. Any Jira issues linked to runs in these contexts are silently dropped —
+they exist in the database but never appear in the API response.
+
+The Jira URL parser (`service/jira_service.py:83`) uses a regex hardcoded to
+`scylladb.atlassian.net`, rejecting issue URLs from any other Atlassian Cloud instance.
```
k0machi: This was intended, actually. You can swap it with the config property if you want.

docs/plans/unify-issue-tracking.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Frontend: Release dashboard renders Jira issues as broken GitHub cards
+
+The release dashboard has a complete rendering pipeline that is GitHub-only:
+
+1. `ReleaseGithubIssues.svelte` fetches from `/api/v1/release/issues` — the backend returns **both**
+   GitHub and Jira issues (confirmed: `stats.py:181` queries both `GithubIssue` and `JiraIssue`).
+2. Issues are passed to `TestWithIssuesCard.svelte` (line 76), which unconditionally renders every
+   issue via `<GithubIssue {issue} />` (line 13).
```
k0machi: It's an unused component, replaced with just `Issues.svelte`

docs/plans/unify-issue-tracking.md
```diff
@@ -0,0 +1,359 @@
...
+3. `GithubIssue.svelte` accesses `issue.repo`, `issue.number`, `issue.owner`, `issue.title`,
```
k0machi: Yes, because they are separated and dispatched by `Issues.svelte`

docs/plans/unify-issue-tracking.md
```diff
@@ -0,0 +1,359 @@
...
+   `issue.url` — none of which exist on Jira issues (which use `key`, `summary`, `project`,
+   `permalink`). The result is broken cards showing `undefined/undefined#undefined`.
+
+The same problem exists in `TestPopoutSelector.svelte` (lines 164-185), which hardcodes the GitHub
+icon and renders `{issue.owner}/{issue.repo}#{issue.number}` — producing `undefined/undefined#undefined`
+for Jira issues.
```
k0machi: This one wasn't updated - it should import the logic from `Issues.svelte`

docs/plans/unify-issue-tracking.md
```diff
@@ -0,0 +1,359 @@
...
+## Context
+
+Argus tracks issues from two providers: GitHub and Jira. Both are modeled in the database
+(`GithubIssue`, `JiraIssue`) and linked to test runs via a shared `IssueLink` table. A unified
+`IssueService` facade dispatches CRUD operations to the correct provider based on the URL.
+
+However, several backend code paths and most of the frontend were originally written for GitHub only,
+with Jira support bolted on later. This has resulted in silent data loss (Jira issues dropped from
+query results), broken rendering (Jira issues passed to GitHub-only components that access
+nonexistent fields), and ~350 lines of copy-pasted code between two nearly identical Svelte
```
k0machi, line 12: Specifically for svelte components and modelling two separate models with different fields was to account for potential divergences in issue APIs, while they could've been the same, the semantic doesn't actually work, e.g. what is Jira's `key` in relation to Github's owner, repo and number fields.

docs/plans/unify-issue-tracking.md
pehala: Will remove

docs/plans/unify-issue-tracking.md
pehala: Each component can be rendered differently if we want, important is that the conditional rendering is done in one place. The Github and Jira should share majority of the code and diverge only when needed to, this can be done via common parent in OOP, probably has a way how to do it in similar fashion in svelte as well

docs/plans/unify-issue-tracking.md
pehala: Not sure what you mean, This component can work with only Github, so it is clearly wrong right?

k0machi: Well, as I mentioned, this particular component hiearchy is unused.

docs/plans/unify-issue-tracking.md
k0machi: They do share the majority of the code in `Issues.svelte` - However they currently duplicate the API functions, which should be extracted into a module context of Issues.svelte

docs/plans/unify-issue-tracking.md
pehala: So should I remove the component?

docs/plans/unify-issue-tracking.md
pehala: Wont remove :D https://github.com/scylladb/argus/blob/3168803c62b460cab5d8a56fd1452f12ff917496/argus/backend/service/jira_service.py#L82 There is indeed a hardcoded regex used. The config value exists but it is not used here

docs/plans/unify-issue-tracking.md
pehala: Changed so the components stay but the duplicates code is extrated

docs/plans/unify-issue-tracking.md
pehala: remove the components here

docs/plans/unify-issue-tracking.md
pehala: #### 3c. `ReleaseDashboard/TestPopoutSelector.svelte` (lines 164-185) — **Fixes broken popout**
Even better it will use the new component issueBadge

docs/plans/unify-issue-tracking.md
k0machi: Yes, that's what I meant

docs/plans/unify-issue-tracking.md
k0machi: I think it's safe to remove, I see 0 references to it.

docs/plans/unify-issue-tracking.md
```diff
@@ -0,0 +1,359 @@
...
+
+---
+
+### Phase 3: Frontend — Fix broken rendering and update consumers
+
+#### 3a. `Github/Issues.svelte` (lines 464-469)
+
+Replace the `{#if issue.subtype == "github"} <GithubIssue> {:else} <JiraIssue> {/if}` branching
+with a single `<IssueCard {issue} ... />`.
+
```
k0machi, line 225: This seems to me that it's just burying the same `if` statement into another component, you can use a component dispatch instead (define in the same ts file)
```js
const IssueComponents = {
  "jira": JiraIssue, 
  "github": GithubIssue,
  "unknown": UnknownIssue, // can be a simple <pre>{JSON.stringify(issue)</pre> component
};
```

Then in the each block you can dispatch:

```svelte
{@const Component = IssueComponents[issue.subtype] ?? IssueComponents.unknown}
<Component {issue} />
```

docs/plans/unify-issue-tracking.md
```diff
@@ -0,0 +1,359 @@
...
+#### 3b. `ReleaseDashboard/TestPopoutSelector.svelte` (lines 164-185) — **Fixes broken popout**
+
+Replace the hardcoded `<Fa icon={faGithub} />` and `{issue.owner}/{issue.repo}#{issue.number}` with
+either `<IssueBadge {issue} />` or inline use of the polymorphic helpers:
+
+- Icon: `issue.subtype === "github" ? faGithub : faJira`
+- Identifier: `getKey(issue)`
+- Title: `getTitle(issue)`
+- URL: `getUrl(issue)`
```
k0machi, line 234: I suggest defining either a mini-mode for both `GithubIssue` and `JiraIssue` or a mini component that will receive the correct subtype.

docs/plans/unify-issue-tracking.md
```diff
@@ -0,0 +1,359 @@
...
+Replace the inline badge markup (lines 18-26) with `{#each issues as issue (issue.id)} <IssueBadge
+{issue} /> {/each}`. This fixes the undefined `issue.number` key, broken state coloring, and
+`#undefined` display for Jira issues.
+
+#### 3f. `Views/Widgets/ViewGraphedStats/Interfaces.ts` (lines 15-20)
+
+Replace the single flat `Issue` interface with a discriminated union matching the main app's type
+system (`IssueTypes.ts`). The graphed stats `Issue` should mirror the same `GithubSubtype` /
+`JiraSubtype` pattern with a shared base:
+
```
k0machi, line 262: No, it should import from the same proposed TS module.

soyacz (commented):
I wonder if we could have just a single table for issues - both Jira and GH (with 'type' field).
each could be handled separately depending on the type wherever is needed, otherwise we would have just single object treated the same (on frontend would be the same, as backend only would see the difference).

docs/plans/unify-issue-tracking.md
pehala: > This seems to me that it's just burying the same `if` statement into another component, you can use a component dispatch instead (define in the same ts file)
> 
> ```js
> const IssueComponents = {
>   "jira": JiraIssue, 
>   "github": GithubIssue,
>   "unknown": UnknownIssue, // can be a simple <pre>{JSON.stringify(issue)</pre> component
> };
> ```
> 
> Then in the each block you can dispatch:
> 
> ```svelte
> {@const Component = IssueComponents[issue.subtype] ?? IssueComponents.unknown}
> <Component {issue} />
> ```

That is IF that is just not hidden and will be copied to every consumer. I do not think this is a good solution.
New component makes it easier to use in every component and if we decide to change anything we wont have to change all of the consumers as all will be hideen inside this simple "redirect" component

docs/plans/unify-issue-tracking.md
pehala: I would do `<IssueBadge {issue} />` with all the backend handling done only inside

pehala:
> I wonder if we could have just a single table for issues - both Jira and GH (with 'type' field).
> each could be handled separately depending on the type wherever is needed, otherwise we would have just single object treated the same (on frontend would be the same, as backend only would see the difference).

You mean unify the types on the backend instead?

docs/plans/unify-issue-tracking.md
k0machi: You can wrap that into the smaller component as well.

k0machi (approved):

docs/plans/unify-issue-tracking.md
pehala: Reworked

pehala:
v3:
  * In section 3b, consider only issueBadge solution

k0machi (approved):

k0machi:
Should I merge it? @pehala

pehala:
> Should I merge it? @pehala 

I dont plan on doing anymore changes

soyacz:
> > I wonder if we could have just a single table for issues - both Jira and GH (with 'type' field).
> > each could be handled separately depending on the type wherever is needed, otherwise we would have just single object treated the same (on frontend would be the same, as backend only would see the difference).
> 
> You mean unify the types on the backend instead?

yes, I meant to send just "issue" to the front, having just a 'type' field. Most of the logic should not require type (especially on the front), only backend will handle things differently (and only in limited scope - e.g. fetching will differ)

pehala:
> yes, I meant to send just \"issue\" to the front, having just a 'type' field. Most of the logic should not require type (especially on the front), only backend will handle things differently (and only in limited scope - e.g. fetching will differ)

The github and jira issue have different additional fields which we might want to use on frontend, for issue specific rendering, so I do not know if it would work as nicely. I might be missing something

PR #978, opened 2026-04-22
fix: enable local development without Jira, GitHub, and email credentials

pehala:
## Problem Statement

Running Argus locally requires valid credentials for three external integrations — Jira, GitHub issue tracking, and SMTP email — even when none of them are needed for development. This creates unnecessary friction for local setup and makes it impossible to run the backend with only ScyllaDB available.

Three distinct issues contribute to this:

1. **Dead code causing a `KeyError` on every request:** `ArgusService.__init__` eagerly reads `GITHUB_ACCESS_TOKEN` from config via a hard `[]` lookup and assigns it to `self.github_headers`. This attribute is never referenced anywhere else in the class, making it dead code that silently crashes the server if the key is absent.

2. **No config-level way to disable Jira/GitHub clients:** `JiraService` and `GithubService` both support a `dry_run=True` mode that skips instantiating remote clients, but there is no way to activate this through configuration. Developers have to patch calling code by hand.

3. **Email credentials required at every request:** `Email.__init__` unconditionally calls `_retrieve_credentials()`, which reads all five `EMAIL_*` config keys via hard lookups. This is triggered on every request through `NotificationManagerService`, which is instantiated by both `ArgusService` and `TestRunService`. There is no way to disable email through configuration.

## Summary

This PR makes all three external integrations independently opt-out through configuration flags, so a developer can start the full backend with only ScyllaDB available.

### Changes

- **Remove dead `github_headers` assignment** from `ArgusService.__init__` (`argus/backend/service/argus_service.py`). The attribute was set but never read, and caused a `KeyError` on startup when `GITHUB_ACCESS_TOKEN` was absent.

- **Add `GITHUB_ENABLED` and `JIRA_ENABLED` config flags** (`argus/backend/service/issue_service.py`). Both default to `true`. When set to `false`, the corresponding remote client is not instantiated — no credentials are required, and local DB reads continue to work normally. Existing code that passes `dry_run=True` explicitly (e.g. tests) is unaffected.

- **Make email credential read lazy** (`argus/backend/util/send_email.py`). Move `_retrieve_credentials()` from `__init__` into `_connect()` so that `EMAIL_*` keys are only read when an SMTP connection is actually being established, not at service construction time.

- **Add `EMAIL_ENABLED` config flag** (`argus/backend/service/notification_manager.py`). Defaults to `true`. When `false`, `EmailNotificationServiceSender` is not instantiated at all — no credentials are read. In-app DB-backed notifications (`ArgusDBNotificationSaver`) are unaffected.

- **Remove dead test fixtures** (`argus/backend/tests/conftest.py`). The `github_service` and `jira_service` fixtures instantiated live clients without `dry_run=True` despite never being used by any test. Removed entirely.

- **Update test app config** (`argus/backend/tests/conftest.py`). Replace fake `EMAIL_*` placeholder values with `EMAIL_ENABLED: False`.

- **Update `argus_web.example.yaml` and `docs/dev-setup.md`** to document all three flags and provide a minimal credential-free local dev config.

### Minimal local dev config (no external credentials required)

```yaml
GITHUB_ENABLED: false
JIRA_ENABLED: false
EMAIL_ENABLED: false
```

### DIFF
--- a/argus/backend/service/argus_service.py
+++ b/argus/backend/service/argus_service.py
@@ -51,10 +51,6 @@ class ArgusService:
         self.session = database_session if database_session else ScyllaCluster.get_session()
         self.database = ScyllaCluster.get()
         self.notification_manager = NotificationManagerService()
-        self.github_headers = {
-            "Accept": "application/vnd.github.v3+json",
-            "Authorization": f"token {current_app.config['GITHUB_ACCESS_TOKEN']}"
-        }
         self.build_id_and_url_statement = self.database.prepare(
             f"SELECT build_id, build_job_url, test_id FROM {SCTTestRun.table_name()} WHERE id = ?"
         )  # TODO: transfer to PluginModelBase
--- a/argus/backend/service/issue_service.py
+++ b/argus/backend/service/issue_service.py
@@ -1,4 +1,5 @@
 from uuid import UUID
+from flask import current_app
 from argus.backend.models.github_issue import GithubIssue, IssueLink
 from argus.backend.service.github_service import GithubService
 from argus.backend.service.issue_utils import build_version_map, filter_links_by_version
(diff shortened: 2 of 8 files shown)


pehala:
Another one of the little annoyances I encountered while developing patches for Argus, should make development setup easier, while not compromising production code

k0machi (approved):

k0machi:
@pehala Conflicts in this PR

pehala:
> @pehala Conflicts in this PR

rebased

k0machi (approved):

PR #979, opened 2026-04-22
docs(plans): add frontend rollup warning remediation plan

pehala:
## Summary
- add a plan document for the current `yarn rollup -c` warning set based on the actual build output
- separate repo-owned warning cleanup from out-of-scope Sass/build-pipeline and third-party warning work
- document that the `Views.svelte` enum warning looks like a false positive and that some state warnings should be reviewed before adding `$state(...)`
- recommend globally ignoring the interactive-element a11y warning family instead of remediating those patterns in this plan
- add a separate follow-up phase for evaluating a Rollup-to-Vite migration

## Validation
- ran `yarn rollup -c` to capture the current warning set for analysis
- updated `docs/plans/frontend-rollup-warning-remediation.md`

### DIFF
--- a/docs/plans/frontend-rollup-warning-remediation.md
+++ b/docs/plans/frontend-rollup-warning-remediation.md
@@ -1,0 +1,304 @@
+# Frontend Rollup Warning Remediation
+
+## Context
+
+Running `yarn rollup -c` completes successfully, but emits a large number of frontend warnings from
+Svelte compilation, TypeScript-to-Svelte type imports, and the Sass/build pipeline.
+
+The current objective is to reduce repo-owned warning noise without taking on intrusive build-system
+changes.
+
+---
+
+## Problem Statement
+
+The frontend build currently mixes three different classes of warnings:
+
+1. Repo-owned Svelte warnings that point to concrete markup, accessibility, or reactivity issues in
+   application code.
+2. Type import and missing export warnings that indicate incorrect cross-file type usage.
+3. Sass and build-pipeline deprecation warnings caused by the current Rollup/PostCSS/Sass stack.
+
+The first two groups are actionable inside the existing app structure and should be cleaned up.
+The Sass/build-pipeline warnings should be documented but treated as out of scope for this work,
+because removing them cleanly requires a broader tooling change rather than a localized fix.
+
+---
+
+## Build Findings
+
+The latest `yarn rollup -c` run produced the following notable warning groups.
+
+### 1. Repo-owned reactivity and state usage warnings
+
+These are the highest-signal warnings because they can point to stale UI state or updates that do
+not trigger correctly under Svelte 5 rune semantics.
+
+- `frontend/Views/Views.svelte`
+  `Tab` is reported as updated without `$state(...)`, but this appears to be a false positive from
+  the enum compilation pattern and should not be treated as a real remediation item.
+- `frontend/TestRun/TestRun.svelte`
+  `activeTab` is referenced in a way that only captures its initial value.
+- `frontend/ReleasePlanner/DutyPlanner.svelte`
+  `newScheduleDate` is referenced in a way that only captures its initial value.
+- `frontend/ReleasePlanner/ReleasePlanCopyForm.svelte`
+  `selectedRelease` is referenced in a way that only captures its initial value.
+- `frontend/TestRun/DriverMatrixTestRun.svelte`
+  `activeTab` warning.
+- `frontend/TestRun/Sirenada/SirenadaTestRun.svelte`
+  `activeTab` warning.
+- `frontend/TestRun/Generic/GenericTestRun.svelte`
+  `activeTab` warning.
+- `frontend/TestRun/Components/Cell.svelte`
+  `value` is referenced in a way that captures only the initial value.
+- `frontend/TestRun/EventsTab.svelte`
+  `eventContainer` is updated but not declared with `$state(...)`.
+- `frontend/Views/Widgets/ViewHighlights/ViewHighlights.svelte`
+  `newGroupName` and `groupCreationError` are reported as non-reactive updates, but these should be
+  reviewed before introducing `$state(...)` because they may not need reactive declaration.
+
+### 2. Repo-owned markup and accessibility warnings
+
+Most of these are straightforward cleanup items: missing labels, icon-only buttons without labels,
+and a small number of syntax or dead-code warnings.
+
+- Missing `aria-label` or text on buttons:
+    - `frontend/Alerts/AlertWidget.svelte`
+    - `frontend/ReleasePlanner/ReleasePlan.svelte`
+    - `frontend/WorkArea/JobConfigureModal.svelte`
+    - `frontend/Common/ModalWindow.svelte`
+    - `frontend/TestRun/Jenkins/JenkinsCloneModal.svelte`
+    - `frontend/TestRun/Jenkins/JenkinsBuildModal.svelte`
+    - `frontend/Profile/NotificationCommentWrapper.svelte`
+    - `frontend/Teams/TeamDetail.svelte`
+    - `frontend/Teams/TeamShort.svelte`
+    - `frontend/AdminPanel/ViewWidget.svelte`
+    - `frontend/AdminPanel/ViewListItem.svelte`
+    - `frontend/ReleasePlanner/Schedule.svelte`
+    - `frontend/WorkArea/JobConfigureModal.svelte`
+- Labels not associated with controls:
+    - `frontend/TestRun/ArtifactTab.svelte`
+    - `frontend/TestRun/ResultsGraphs.svelte`
+    - `frontend/Views/WidgetSettingTypes/CheckValue.svelte`
+    - `frontend/Views/WidgetSettingTypes/StringValue.svelte`
+    - `frontend/Views/WidgetSettingTypes/IntegerValue.svelte`
+    - `frontend/TestRun/Jenkins/CloneTargetSelector.svelte`
+    - `frontend/TestRun/Jenkins/DummyCheckParam.svelte`
+- Invalid self-closing non-void HTML tags:
+    - `frontend/ReleasePlanner/Schedule.svelte`
+- Quoted component/custom element attributes that will stringify in future Svelte versions:
+    - `frontend/Teams/TeamMember.svelte`
+- Empty block:
+    - `frontend/Views/Widgets/PytestWidget/PytestFlatHelper.svelte`
+- Unused CSS selectors:
+    - `frontend/TestRun/PackagesInfo.svelte`
+    - `frontend/Discussion/Comment.svelte`
+    - `frontend/Discussion/CommentEditor.svelte`
+    - `frontend/TestRun/Components/Filters.svelte`
+    - `frontend/Views/Widgets/ViewGraphedStats.svelte`
+
+### 3. Incorrect type import / missing export warnings
+
+These are correctness issues in how types are imported from `.svelte` files or adjacent modules.
+
+- `frontend/Views/Widgets/ViewGraphedStats/StackTracePreview.svelte`
+  imports `TestRun` from `./Interfaces`, but `Interfaces.ts` does not export it.
+- `frontend/Views/Widgets/ViewGraphedStats/StatusBadge.svelte`
+  same missing `TestRun` export.
+- `frontend/Views/Widgets/PytestWidget/PytestResultRow.svelte`
+  imports `PytestResult` from `ViewPytestOverview.svelte`.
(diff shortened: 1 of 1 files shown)


pehala:
@k0machi This is more of a maintenance work, the warnings can sometimes point to real issues and they do make sense to resolve. I will remove anything you consider either wrong or with too much of a risk.

The most annoying warnings are the JS API deprecated, which are intentionally left out here and which will be resolved only by switching to Vite instead of rolloup, which will be an intrusive change (I think we should do it still, just a bit later)

k0machi:
> @k0machi This is more of a maintenance work, the warnings can sometimes point to real issues and they do make sense to resolve. I will remove anything you consider either wrong or with too much of a risk.

> The most annoying warnings are the JS API deprecated, which are intentionally left out here and which will be resolved only by switching to Vite instead of rolloup, which will be an intrusive change (I think we should do it still, just a bit later)

Vite change is planned, somewhere after fastapi migrations due to splitting frontend into a SvelteKit app. We can of course expedite just switching to vite, it's not that big of a change:

* Change to vite
* Install same plugins as we have for rollup
* Convert the config
* Check tree-shaking settings
* On existing deployments: clear the public/dist folder
* Ensure that templates still refer to the same bundle names (not needed if names are preserved)

docs/plans/frontend-rollup-warning-remediation.md
```diff
@@ -0,0 +1,277 @@
...
+- Empty block:
+    - `frontend/Views/Widgets/PytestWidget/PytestFlatHelper.svelte`
+- Unused CSS selectors:
+    - `frontend/TestRun/PackagesInfo.svelte`
+    - `frontend/Discussion/Comment.svelte`
+    - `frontend/Discussion/CommentEditor.svelte`
+    - `frontend/TestRun/Components/Filters.svelte`
+    - `frontend/Views/Widgets/ViewGraphedStats.svelte`
+
+### 3. Incorrect type import / missing export warnings
```
k0machi, line 100: Okay to do.

docs/plans/frontend-rollup-warning-remediation.md
```diff
@@ -0,0 +1,277 @@
...
+- `frontend/TestRun/Generic/GenericTestRun.svelte`
+  `activeTab` warning.
+- `frontend/TestRun/Components/Cell.svelte`
+  `value` is referenced in a way that captures only the initial value.
+- `frontend/TestRun/EventsTab.svelte`
+  `eventContainer` is updated but not declared with `$state(...)`.
+- `frontend/Views/Widgets/ViewHighlights/ViewHighlights.svelte`
+  `newGroupName` and `groupCreationError` are updated but not declared with `$state(...)`.
+
+### 2. Repo-owned markup and accessibility warnings
```
k0machi, line 60: Refer to comments below

docs/plans/frontend-rollup-warning-remediation.md
```diff
@@ -0,0 +1,277 @@
...
+
+Most of these are straightforward cleanup items: missing labels, icon-only buttons without labels,
+or clickable non-interactive elements.
+
+- Missing `aria-label` or text on buttons:
```
k0machi, line 65: Honestly I'd silence that on project level, I don't think we'll ever need to support screen readers...

docs/plans/frontend-rollup-warning-remediation.md
```diff
@@ -0,0 +1,277 @@
...
+    - `frontend/WorkArea/JobConfigureModal.svelte`
+- Labels not associated with controls:
+    - `frontend/TestRun/ArtifactTab.svelte`
+    - `frontend/TestRun/ResultsGraphs.svelte`
+    - `frontend/Views/WidgetSettingTypes/CheckValue.svelte`
+    - `frontend/Views/WidgetSettingTypes/StringValue.svelte`
+    - `frontend/Views/WidgetSettingTypes/IntegerValue.svelte`
+    - `frontend/TestRun/Jenkins/CloneTargetSelector.svelte`
+    - `frontend/TestRun/Jenkins/DummyCheckParam.svelte`
+- Click handlers on non-interactive elements:
```
k0machi: This could be a big one - The cleanest way of dealing with this is converting them to buttons and such but this might shift the current markup.

docs/plans/frontend-rollup-warning-remediation.md
```diff
@@ -0,0 +1,277 @@
...
+    - `frontend/TestRun/Jenkins/JenkinsCloneModal.svelte`
+    - `frontend/TestRun/Jenkins/JenkinsBuildModal.svelte`
+    - `frontend/Profile/NotificationCommentWrapper.svelte`
+    - `frontend/Teams/TeamDetail.svelte`
+    - `frontend/Teams/TeamShort.svelte`
+    - `frontend/AdminPanel/ViewWidget.svelte`
+    - `frontend/AdminPanel/ViewListItem.svelte`
+    - `frontend/ReleasePlanner/Schedule.svelte`
+    - `frontend/WorkArea/JobConfigureModal.svelte`
+- Labels not associated with controls:
```
k0machi, line 79: Should probably be silenced - This is again something to help accessibility and the association is checked via `id` and `for` attributes, which aren't needed in our cases

docs/plans/frontend-rollup-warning-remediation.md
```diff
@@ -0,0 +1,277 @@
...
+- `frontend/TestRun/Sirenada/SirenadaTestRun.svelte`
+  `activeTab` warning.
+- `frontend/TestRun/Generic/GenericTestRun.svelte`
+  `activeTab` warning.
+- `frontend/TestRun/Components/Cell.svelte`
+  `value` is referenced in a way that captures only the initial value.
+- `frontend/TestRun/EventsTab.svelte`
+  `eventContainer` is updated but not declared with `$state(...)`.
+- `frontend/Views/Widgets/ViewHighlights/ViewHighlights.svelte`
+  `newGroupName` and `groupCreationError` are updated but not declared with `$state(...)`.
```
k0machi: These two might not need declaration of $state()

docs/plans/frontend-rollup-warning-remediation.md
```diff
@@ -0,0 +1,277 @@
...
+
+The latest `yarn rollup -c` run produced the following notable warning groups.
+
+### 1. Repo-owned reactivity and state usage warnings
+
+These are the highest-signal warnings because they can point to stale UI state or updates that do
+not trigger correctly under Svelte 5 rune semantics.
+
+- `frontend/Views/Views.svelte`
+  `Tab` is updated but not declared with `$state(...)`.
```
k0machi: `Tab` is an enum and is never updated. This is most likely a linter bug.

docs/plans/frontend-rollup-warning-remediation.md
```diff
@@ -0,0 +1,277 @@
...
+The Sass/build-pipeline warnings should be documented but treated as out of scope for this work,
+because removing them cleanly requires a broader tooling change rather than a localized fix.
+
+---
+
+## Build Findings
+
+The latest `yarn rollup -c` run produced the following notable warning groups.
+
+### 1. Repo-owned reactivity and state usage warnings
```
k0machi, line 32: This will need careful checking of what is being done.

docs/plans/frontend-rollup-warning-remediation.md
```diff
@@ -0,0 +1,277 @@
...
+    - `frontend/TestRun/Components/ScreenshotModal.svelte`
+    - `frontend/TestRun/DriverMatrixBreakdown.svelte`
+    - `frontend/Teams/TeamMember.svelte`
+- Invalid self-closing non-void HTML tags:
+    - `frontend/ReleasePlanner/Schedule.svelte`
+- Quoted component/custom element attributes that will stringify in future Svelte versions:
+    - `frontend/Teams/TeamMember.svelte`
+- Empty block:
+    - `frontend/Views/Widgets/PytestWidget/PytestFlatHelper.svelte`
+- Unused CSS selectors:
```
k0machi, line 93: Easy fix.

docs/plans/frontend-rollup-warning-remediation.md
```diff
@@ -0,0 +1,277 @@
...
+- `frontend/Views/Widgets/ViewGraphedStats/StatusBadge.svelte`
+  same missing `TestRun` export.
+- `frontend/Views/Widgets/PytestWidget/PytestResultRow.svelte`
+  imports `PytestResult` from `ViewPytestOverview.svelte`.
+- `frontend/Profile/PlannedJob.svelte`
+  imports `PlannedTest` from `PlannedJobs.svelte`.
+- `frontend/ReleaseDashboard/TestDashboard.test.ts`
+  imports named exports from `TestDashboard.svelte` that are not exported as module exports.
+
+### 4. Build/tooling warnings observed but not targeted here
```
k0machi, line 115: For Sass warnings I'd double check by upgrading rollup, sass and svelte and other packages to see if it's truly the case.

docs/plans/frontend-rollup-warning-remediation.md
pehala: I partially agree, if it is not much work I think it is better to add the the lables, given it is not much work.

docs/plans/frontend-rollup-warning-remediation.md
pehala: Yeah, I will remove this one

k0machi:
In general I suggest this: We need to update our frontend stack to the current locked versions according to our package.json and see what changes in the compilation output, then we can deal with actual problems.

docs/plans/frontend-rollup-warning-remediation.md
pehala: We should do best effort on accessibility IMO, especially since having it be accessible it easy enough to do and doesnt affect markup

pehala:
> In general I suggest this: We need to update our frontend stack to the current locked versions according to our package.json and see what changes in the compilation output, then we can deal with actual problems.

I did that and AI analysis ended with: "Yeah even the new versions of our stack use deprecated API". Given it was AI, it might be worth revisiting, but I would be inclided to believe especially since vite migration is something we want to do

pehala:
> Vite change is planned, somewhere after fastapi migrations due to splitting frontend into a SvelteKit app. We can of course expedite just switching to vite, it's not that big of a change:
> 
>     * Change to vite
> 
>     * Install same plugins as we have for rollup
> 
>     * Convert the config
> 
>     * Check tree-shaking settings
> 
>     * On existing deployments: clear the public/dist folder
> 
>     * Ensure that templates still refer to the same bundle names (not needed if names are preserved)

I will think about it more and might contribute a followup with exactly that: Switch stack to vite and keep things in flask unchanged. I worry most about production deployment, the command to compile frontnend will change so we will need to coordinate, but if that is not a big problem I think I will be able to do it.

docs/plans/frontend-rollup-warning-remediation.md
pehala: Agreed, it is hard to judge it for me, until I see it changed, so maybe we can just let AI fix it and remove it from the implementation PR if it bad

k0machi:
> 
> I will think about it more and might contribute a followup with exactly that: Switch stack to vite and keep things in flask unchanged. I worry most about production deployment, the command to compile frontnend will change so we will need to coordinate, but if that is not a big problem I think I will be able to do it.

In theory it's just `yarn vite` instead of `yarn rollup`, so it's no big deal.

docs/plans/frontend-rollup-warning-remediation.md
k0machi: It's a lot of extra noise in the code for no benefit is my primary objection.

docs/plans/frontend-rollup-warning-remediation.md
k0machi: The problem is then you need to enforce `id` uniqueness across ALL components, otherwise you'll get browser warnings about duplicate `id` attributes :)

pehala: That doesnt seem to be a problem to me. I would suggest trying out and removing if it is too many chaanges

docs/plans/frontend-rollup-warning-remediation.md
pehala: I dont think one html param per button is a lot of extra noise, I would again suggest trying it out (doing it with AI is free :D) and we can put it out from the implementation PR if we deem it too noisy

pehala:
Closing in favour of https://github.com/scylladb/argus/pull/986, as the discussion was mostly concluded here and the implementation showed that is was reasonable.

PR #981, opened 2026-04-22
perf(dashboard): cache release stats

pehala:
## Summary

Three independent performance levers targeting the release dashboard, which was making 3 slow serial API calls on every page load (~6s total). All measurements on a **single-node local ScyllaDB cluster with ~500K seeded runs across 6 tests** (`dev-db/bulk_seed_runs.py`).

| Endpoint | Before | After (cold) | After (warm cache) |
|---|---|---|---|
| `GET /api/v1/release/stats/v2` | ~1.7s | ~1.7s | **~5–12ms** |
| `GET /api/v1/release/<id>/versions` | ~2.1s | ~2.1s | **~7ms** |
| `GET /api/v1/release/<id>/images` | ~2.1s | ~2.1s | **~6ms** |
| **Total dashboard load** | **~6s** | **~6s (first load)** | **~20ms** |

---

### Lever A — CL=ONE read profile for stats fan-out (`39001c9`)

Fan-out queries in `ReleaseStatsCollector` switched from default QUORUM to the `read_fast` CL=ONE execution profile. ~5% improvement on a single-node local cluster; more meaningful on production multi-node RF>1 deployments where it avoids coordinator round-trips to a replica.

### Lever B — Release stats snapshot cache (`15f4891`)

Persist the serialized `ReleaseStats.to_dict()` result in a new `argus_release_stats_snapshot` Scylla table keyed by `(release_id, filter_key)`. Cache hit returns in ~5ms instead of ~1.7s.

Cache invalidation is **explicit, event-driven, and version-scoped**: each plugin model calls `invalidate_release_snapshot()` (inherited from `PluginModelBase`) on `submit_run()` and `finish_run()`. Only snapshots matching the run's `scylla_version` and the all-versions aggregate (`v=::*`) are dropped — other version snapshots stay warm.

Also fixes a bug where `fetchReleaseStats()` and `fetchViewStats()` in the frontend were hardcoding `force: new Number(true)` on every request (including the 5-minute auto-refresh), which meant the cache was written but never read.

### Lever C — Denormalized index tables for distinct versions and images (`1e2735c`)

Two new tables — `argus_release_distinct_versions` and `argus_release_distinct_images` — hold the set of distinct values per release. Version/image filter dropdown reads now hit a single-partition index returning O(5–50) rows instead of scanning the full SCT table (~500K rows, ~2.1s).

Write side-effects are owned by each plugin model (`index_version()` on `PluginModelBase`; SCT-specific `index_image()` on `SCTTestRun`). A fallback to the original GSI scan is in place when the index partition is empty. Backfill via `scripts/migration/migration_2026-04-22.py` (idempotent).

---

### Automated tests

```bash
# Start single-node ScyllaDB
docker compose -f dev-db/docker-compose.yml up -d

# Seed ~500K runs
uv run python dev-db/bulk_seed_runs.py --runs 500000 --release <release-name>

# Run the test suite (31 tests, requires running ScyllaDB)
uv run pytest argus/backend/tests/test_stats_snapshot.py -v
# 31 passed in ~24s

# After deploying to an existing environment, run migration once:
uv run scripts/migration/migration_2026-04-22.py
```

### Manual UI testing

Start the dev server and open a release dashboard with a seeded release:

```bash
FLASK_ENV=development FLASK_APP=argus_backend:start_server CQLENG_ALLOW_SCHEMA_MANAGEMENT=1 uv run flask run
# In a separate terminal:
ROLLUP_ENV=development yarn rollup -c --watch
```

Things to verify in the browser:

1. **Initial page load** — open a release dashboard, observe the stats table loads. Open DevTools Network tab and confirm `stats/v2`, `versions`, and `images` all return quickly on the second load (~5–20ms vs ~2s before).
2. **Version filter** — switch between versions in the filter dropdown; each selection should return near-instantly from cache after the first load.
3. **Refresh button** — click the manual refresh button (⟳); confirm it bypasses the cache (`force=true`) and recomputes fresh stats, then subsequent loads are fast again.
4. **Auto-refresh** — leave the page open for 5+ minutes; confirm the auto-refresh no longer forces a full recompute on every tick (was broken before this fix).
5. **Release deletion** — delete a release via the admin panel and confirm no orphaned rows remain in `argus_release_stats_snapshot`, `argus_release_distinct_versions`, or `argus_release_distinct_images`.

## Files changed

| File | Change |
|---|---|
| `argus/backend/models/web.py` | `ReleaseStatsSnapshot`, `ReleaseDistinctVersions`, `ReleaseDistinctImages` models |
| `argus/backend/plugins/core.py` | `invalidate_release_snapshot()` (version-scoped), `index_version()` on `PluginModelBase` |
| `argus/backend/plugins/sct/testrun.py` | `index_image()` (renamed public); call sites in `submit_run`/`finish_run` |
| `argus/backend/plugins/driver_matrix_tests/model.py` | `submit_run`/`finish_run` side-effect hooks |
| `argus/backend/plugins/generic/model.py` | `submit_run`/`finish_run` side-effect hooks |
| `argus/backend/service/stats.py` | Snapshot read/write in `ReleaseStatsCollector.collect()` |
| `argus/backend/service/client_service.py` | Thin dispatcher — no side-effect logic |
| `argus/backend/service/release_manager.py` | `delete_release()` cleans all three new table types |
| `frontend/ReleaseDashboard/TestDashboard.svelte` | Stop hardcoding `force=true` on every load |
| `scripts/migration/migration_2026-04-22.py` | Idempotent backfill for index tables |
| `dev-db/bulk_seed_runs.py` | Helper script to seed runs for performance testing |
| `argus/backend/tests/test_stats_snapshot.py` | 31 tests (unit + integration via real ScyllaDB) |
| `docs/plans/dashboard-performance-optimization.md` | Design doc |

### DIFF
--- a/argus/backend/plugins/core.py
+++ b/argus/backend/plugins/core.py
@@ -173,7 +173,8 @@ class PluginModelBase(Model):
         for step in range(0, ceil(len(build_ids) / step_size)):
             start_pos = step*step_size
             next_slice = build_ids[start_pos:start_pos+step_size]
-            futures.append(cluster.session.execute_async(query=query, parameters=(next_slice,)))
+            futures.append(cluster.session.execute_async(query=query, parameters=(next_slice,),
+                                                         execution_profile="read_fast"))
 
         return futures
 
--- a/argus/backend/plugins/core.py
+++ b/argus/backend/plugins/core.py
@@ -4,7 +4,6 @@ from datetime import datetime, UTC
 from math import ceil
 from uuid import UUID
 from time import time
-
 from cassandra.cqlengine import columns
 from cassandra.cqlengine.models import Model
 from cassandra.cqlengine.usertype import UserType
@@ -19,7 +18,9 @@ from argus.backend.models.web import (
     ArgusScheduleGroup,
     ArgusSchedule,
     ArgusScheduleTest,
-    ArgusScheduleAssignee
+    ArgusScheduleAssignee,
+    ReleaseStatsSnapshot,
+    ReleaseDistinctVersions,
 )
 from argus.backend.util.common import chunk
 from argus.common.enums import TestInvestigationStatus, TestStatus
@@ -45,6 +46,7 @@ class PluginModelBase(Model):
     build_job_url = columns.Text()
     build_number = columns.Integer()
     product_version = columns.Text(index=True)
+    scylla_version = columns.Text()
 
     # Test Logs Collection
     logs = columns.List(value_type=columns.Tuple(columns.Text(), columns.Text()))
@@ -276,6 +278,27 @@ class PluginModelBase(Model):
 
     def sut_timestamp(self, sut_package_name) -> float:
         raise NotImplementedError()
+
+    def invalidate_release_snapshot(self) -> None:
+        if not self.release_id:
+            return
+        try:
+            version = self.scylla_version or ""
+            version_prefix = f"v={version}::"
+            all_versions_prefix = "v=::"
+            for snapshot in ReleaseStatsSnapshot.filter(release_id=self.release_id).all():
+                if snapshot.filter_key.startswith(version_prefix) or snapshot.filter_key.startswith(all_versions_prefix):
+                    snapshot.delete()
+        except Exception:
+            LOGGER.warning("Failed to invalidate stats snapshot for release %s", self.release_id, exc_info=True)
+
+    def index_version(self) -> None:
+        if not self.release_id or not self.scylla_version:
+            return
+        try:
+            ReleaseDistinctVersions.create(release_id=self.release_id, version=self.scylla_version)
+        except Exception:
+            LOGGER.warning("Failed to index version %s for release %s", self.scylla_version, self.release_id, exc_info=True)
 
 
 class PluginInfoBase:
--- a/argus/backend/plugins/sct/testrun.py
+++ b/argus/backend/plugins/sct/testrun.py
@@ -216,19 +216,19 @@ class SCTTestRun(PluginModelBase):
         req = SCTTestRunSubmissionRequest(**request_data)
         run = cls.from_sct_config(req=req)
         run.invalidate_release_snapshot()
-        run._index_image(run)
+        run.index_image(run)
         return run
 
     @classmethod
     def get_distinct_product_versions(cls, release: ArgusRelease) -> list[str]:
+        rows = list(ReleaseDistinctVersions.filter(release_id=release.id).all())
+        if rows:
+            return sorted([r.version for r in rows], reverse=True)
+        # Fallback: index not yet populated — scan GSI directly
         cluster = ScyllaCluster.get()
-        statement = cluster.prepare(f"SELECT scylla_version FROM {
-                                    cls.table_name()} WHERE release_id = ?")
-        rows = cluster.session.execute(
-            query=statement, parameters=(release.id,))
-        unique_versions = {r["scylla_version"]
-                           for r in rows if r["scylla_version"]}
-
+        statement = cluster.prepare(f"SELECT scylla_version FROM {cls.table_name()} WHERE release_id = ?")
+        result = cluster.session.execute(query=statement, parameters=(release.id,))
+        unique_versions = {r["scylla_version"] for r in result if r["scylla_version"]}
         return sorted(list(unique_versions), reverse=True)
 
     @classmethod
@@ -248,13 +248,14 @@ class SCTTestRun(PluginModelBase):
 
     @classmethod
     def get_distinct_cloud_images_for_release(cls, release: ArgusRelease):
+        rows = list(ReleaseDistinctImages.filter(release_id=release.id).all())
+        if rows:
+            return sorted([r.image_id for r in rows], reverse=True)
+        # Fallback: index not yet populated — scan GSI + deserialize UDTs directly
         cluster = ScyllaCluster.get()
-        statement = cluster.prepare(f"SELECT cloud_setup FROM {
-                                    cls.table_name()} WHERE release_id = ?")
-        rows = cluster.session.execute(
-            query=statement, parameters=(release.id,))
-        unique_images = {cls.get_image(r) for r in rows if cls.get_image(r)}
-
+        statement = cluster.prepare(f"SELECT cloud_setup FROM {cls.table_name()} WHERE release_id = ?")
+        result = cluster.session.execute(query=statement, parameters=(release.id,))
+        unique_images = {cls.get_image(r) for r in result if cls.get_image(r)}
         return sorted(list(unique_images), reverse=True)
 
     @classmethod
@@ -419,7 +420,7 @@ class SCTTestRun(PluginModelBase):
         self.end_time = datetime.utcnow()
         self.invalidate_release_snapshot()
         self.index_version()
-        self._index_image(self)
+        self.index_image(self)
 
     def submit_logs(self, logs: list[dict]):
         for log in logs:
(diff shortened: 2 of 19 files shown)


pehala:
This is a big change, but it made 500K rows instant locally. Important is that it is cached on server side, so single cache warmup will last until new run appears/is deleted. Given that release dashboard are not updated that regularly unless there is active RC in testing and people are doing a lot of one time tests, this should be massive savings. Master duty view is updated once per day so it will also massively benefit.

pehala:
I am intentionally submitting both plan and code at once, because I need the code to verify the improvements and also the whole PR started as a PoC. For initial review, I would like to focus on general idea verification and design implications.

The PR should be functional enough to try on staging (if we can easily redo the DB if we want to try something else)

argus/backend/plugins/sct/testrun.py
```diff
@@ -474,6 +482,18 @@ def get_run_response(cls, run_id: UUID) -> dict | None:
...
 
+    @staticmethod
+    def index_image(run: 'SCTTestRun') -> None:
+        if not run.release_id:
+            return
+        try:
+            image_id = run.cloud_setup and run.cloud_setup.db_node and run.cloud_setup.db_node.image_id
+            if not image_id:
+                return
+            ReleaseDistinctImages.if_not_exists().create(release_id=run.release_id, image_id=image_id)
```
k0machi: I would replace with upsert since checking `if not exists` here is pointless due to all fields being exactly the same.

argus/backend/plugins/sct/testrun.py
```diff
@@ -204,18 +204,21 @@ def load_test_run(cls, run_id: UUID) -> 'SCTTestRun':
...
     def submit_run(cls, request_data: dict) -> 'SCTTestRun':
         req = SCTTestRunSubmissionRequest(**request_data)
-        return cls.from_sct_config(req=req)
+        run = cls.from_sct_config(req=req)
+        run.invalidate_release_snapshot()
+        run.index_image(run)
+        return run
 
     @classmethod
     def get_distinct_product_versions(cls, release: ArgusRelease) -> list[str]:
```
k0machi, line 223: It seems only SCT currently implements this new logic. Is that planned? I would suggest updating these methods on other plugins as well.

argus/backend/plugins/core.py
```diff
@@ -261,6 +261,31 @@ def finish_run(self, payload: dict = None):
     def sut_timestamp(self, sut_package_name) -> float:
         raise NotImplementedError()
 
+    def invalidate_release_snapshot(self) -> None:
+        # Lazy import to avoid circular dependency: web.py → stats.py → plugins → core.py
+        from argus.backend.models.web import ReleaseStatsSnapshot
```
pehala: inline import...

argus/backend/plugins/core.py
```diff
@@ -261,6 +261,31 @@ def finish_run(self, payload: dict = None):
...
+            version_prefix = f"v={version}::"
+            all_versions_prefix = "v=::"
+            for snapshot in ReleaseStatsSnapshot.filter(release_id=self.release_id).all():
+                if snapshot.filter_key.startswith(version_prefix) or snapshot.filter_key.startswith(all_versions_prefix):
+                    snapshot.delete()
+        except Exception:
+            LOGGER.warning("Failed to invalidate stats snapshot for release %s", self.release_id, exc_info=True)
+
+    def index_version(self) -> None:
+        from argus.backend.models.web import ReleaseDistinctVersions
```
pehala: inline import...

argus/backend/plugins/sct/testrun.py
pehala: Some of the new methods are used in other plugins as well

pehala:
@k0machi @soyacz would it be possible to deploy this PoC on the staging to verify its usefulness on production grade data?

k0machi:
> @k0machi @soyacz would it be possible to deploy this PoC on the staging to verify its usefulness on production grade data?

I'll do it, however, question - is there currently a way to refresh this cache without having a run change something in a release? E.g. does the cache have a timestamp on it?

pehala:
> I'll do it, however, question - is there currently a way to refresh this cache without having a run change something in a release? E.g. does the cache have a timestamp on it?

Cache should be invalidated automatically only, there is no manual clearance mechanism.

k0machi:
> > I'll do it, however, question - is there currently a way to refresh this cache without having a run change something in a release? E.g. does the cache have a timestamp on it?
> 
> Cache should be invalidated automatically only, there is no manual clearance mechanism.

What I mean by this is what triggers the cache build in this PR - i.e. can I deploy it on staging and see it actually build the cache automatically?

pehala:
> What I mean by this is what triggers the cache build in this PR - i.e. can I deploy it on staging and see it actually build the cache automatically?

Yes, it will be automatic

pehala:
v2:
  * Remove cache control from the dashboard API endpoints
  * Add scylla_version to PluginBase, as all plugins have it defined regardless
     * Remove it from the plugins 
  * Rebased

argus/backend/plugins/core.py
```diff
@@ -276,6 +279,27 @@ def finish_run(self, payload: dict = None):
     def sut_timestamp(self, sut_package_name) -> float:
         raise NotImplementedError()
 
+    def invalidate_release_snapshot(self) -> None:
+        if not self.release_id:
+            return
+        try:
+            version = self.scylla_version or ""
+            version_prefix = f"v={version}::"
+            all_versions_prefix = "v=::"
```
k0machi, line 288: I feel like this could be extracted into a helper file to make it more organized.

argus/backend/service/stats.py
```diff
@@ -15,12 +16,16 @@
...
 from argus.backend.models.web import ArgusRelease, ArgusGroup, ArgusScheduleTest, ArgusTest, \
-    ArgusTestRunComment, ArgusUserView
+    ArgusTestRunComment, ArgusUserView, ReleaseStatsSnapshot
 from argus.backend.db import ScyllaCluster
 
 LOGGER = logging.getLogger(__name__)
 
 
+def snapshot_filter_key(version: str | None, image_id: str | None, include_no_version: bool, limited: bool) -> str:
+    return f"v={version or ''}::img={image_id or ''}::nov={int(include_no_version)}::lim={int(limited)}"
```
k0machi, line 26: Ditto

k0machi (approved):
LGTM, minor comment

pehala:
## What changed in the latest push

Added full-partition snapshot cache invalidation to all code paths that mutate dashboard-visible data.

### New helper: `invalidate_release_snapshots(release_id)`

Added a module-level function in `argus/backend/models/web.py`. It deletes all `ReleaseStatsSnapshot` rows for the given `release_id` partition in one sweep. Exceptions are caught and logged as warnings so a cache miss never surfaces to the caller.

This complements the existing `PluginModelBase.invalidate_release_snapshot()` which does a narrower, version-scoped wipe used only for run lifecycle events (`submit_run` / `finish_run`). The full-partition wipe is necessary for structural or metadata changes because they affect every filter-key combination stored in the snapshot partition.

### Invalidation call sites added

| File | Methods |
|------|---------|
| `service/release_manager.py` | `toggle_test_enabled`, `toggle_group_enabled`, `create_group`, `create_test`, `delete_group`, `delete_test`, `update_group`, `update_test`, `set_release_state`, `set_release_dormancy`, `set_release_perpetuality`, `edit_release`, `batch_move_tests` |
| `service/testrun.py` | `change_run_status`, `change_run_investigation_status`, `change_run_assignee`, `post_run_comment`, `delete_run_comment`, `update_run_comment`, `ignore_jobs` |
| `service/jira_service.py` | `submit_issue`, `delete_issue` |
| `service/github_service.py` | `submit_issue`, `delete_issue` |
| `service/planner_service.py` | `create_plan`, `update_plan`, `copy_plan`, `delete_plan`, `complete_plan` |

`delete_release` already wiped the full snapshot partition as part of its cleanup; no change needed there.

pehala:
> ## What changed in the latest push
> 
> Added full-partition snapshot cache invalidation to all code paths that mutate dashboard-visible data.

More invalidations. In general when using this pattern the cache invalidation is tricky and we might miss a few places, it is something to be wary of

k0machi (approved):

PR #982, opened 2026-04-23
add more backend tests

k0machi:
- **test(driver-matrix-api): add controller integration tests**
  Cover all four endpoints in argus/backend/plugins/driver_matrix_tests/controller.py
  (/result/submit, /result/fail, /env/submit, /test_report) plus two error paths.
  Verification is JSON-first via the paired /test_report endpoint; ORM fallback is
  used only for /result/fail and /env/submit since neither failure_message nor
  environment_info are exposed by any GET endpoint.
  
  Also extends docs/plans/api_tests.md with a Controller Coverage Matrix tracking
  the per-controller integration-test rollout.
  

- **test(driver-matrix-api): drop ORM fallback for fail/env verification**
  Replace the DriverTestRun.get(...) fallback in test_submit_driver_matrix_failure
  and test_submit_driver_matrix_env with a paired GET against
  /api/v1/run/driver-matrix-tests/<run_id>. PluginModelBase.get_run_response
  returns dict(run.items()) which already exposes test_collection,
  environment_info, and scylla_version, so JSON-only verification is sufficient.
  
  Also updates the controller coverage matrix in docs/plans/api_tests.md to
  remove the ORM-fallback caveat for this module.
  

- **test(sct-api): add controller tests for events, stress commands, similar runs**
  

- **docs(plans): mark sct-api iteration 2 complete**
  

- **test(client-api): add controller tests for run lifecycle, pytest, render report**
  

- **test(testrun-api): add controller tests for run lifecycle, comments, activity, pytest**
  

- **test(release-api): add controller tests for releases, schedules, groups and tests**
  Covers /version, /releases, release detail/versions/images/pytest/activity,
  planner data + comments, schedules submit/update/delete/comment/assignee,
  group + test assignees, /release/stats/v2, /release/create, /groups, /tests,
  group/test details, /test/<id>/set_plugin and /test-info.
  
  Adds GITHUB_ACCESS_TOKEN to the test config since ArgusService.__init__
  hard-references it via current_app.config[...].
  

- **test(api): add controller tests for users, tokens, polls and proxy endpoints**
  Covers /users, /user/token (stability), /user/jobs, /user/planned_jobs,
  /test_runs/poll (with additionalRuns dedup), /test_run/poll (single + unknown),
  /artifact/resolveSize (HTTP HEAD mocked), /zeus/<endpoint> error paths and the
  deprecated /test_run/comment/get round-trip via the comments submit endpoint.
  

- **test(api): add controller tests for graph views and /test-results**
  Covers /create-graph-view, /update-graph-view (success + unknown-id) with
  round-trip verification through /test-results, plus /test-results GET
  (default + date-range + missing testId) and HEAD (404 when no results).
  

- **test(admin-api): add controller tests for admin release CRUD**
  Covers /admin/api/v1/release/{create,set_perpetual,set_state,set_dormant,edit,delete}
  plus /releases/get. State changes verified via /api/v1/release/<id>/details
  (no ORM fallback). Includes JSON content-type validation.
  

- **test(admin-api): add controller tests for admin group/test CRUD**
  Covers /group/{create,update,delete} including delete-with-cascade and
  delete-with-relocation, /test/{create,update,delete,batch_move},
  /groups/get + /tests/get plus group/test state toggles. Mutations verified
  through /api/v1/group|test/<id>/details, no ORM fallback.
  

- **test(admin-api): add controller tests for admin user endpoints**
  

- **test(notification-api): add controller tests for notification endpoints**
  

- **test(view-api): add controller tests for user view CRUD and resolution endpoints**
  

- **test(team-api): add controller tests for team CRUD and membership endpoints**
  

- **test(planner-api): add controller tests for plan CRUD, copy eligibility and gridview**
  

- **test(widgets): add controller tests for summary widget endpoints**
  

- **test(widgets): add controller tests for pytest widget endpoints**
  

- **test(widgets): add controller tests for nemesis stats widget endpoint**
  

- **test(widgets): add controller tests for graphs widget endpoint**
  

- **test(widgets): add controller tests for graphed_stats widget endpoints**
  

- **test(auth): add controller tests for auth blueprint endpoints**
  

- **test(profile): add controller tests for profile update and picture storage endpoints**
  

- **docs(api-tests): mark highlights widget iteration complete (pre-existing coverage verified)**
  

- **test(infra): add shared external-service mocks and widget seeding fixtures**
  Phase 1 of the closeout effort for partial controller test suites:
  
  - Extend conftest.py with config keys for GitHub OAuth, Jira, Jenkins,
    AWS/S3 and Cloudflare Access so service constructors don't crash in
    tests. CLOUDFLARE_ACCESS_JWK_CLIENT is set to a SimpleNamespace stub
    so no JWKS HTTP fetch happens.
  - Add argus/backend/tests/_helpers/external_mocks.py exposing fixtures
    (mock_issue_service, mock_github_callback, mock_jenkins_service,
    mock_s3, mock_cf_access_payload) that patch the highest stable
    service-method boundary so tests never touch the network.
  - Add argus/backend/tests/widgets/conftest.py with seven function-scoped
    seeding fixtures (sct_run, seeded_view_with_run, sct_run_with_nemesis,
    generic_results_for_run, graph_view_for_test, seeded_pytest_row,
    linked_github_issue) sufficient to exercise the full data-shape
    branches of the five view-widget controllers.
  

- **test(widgets): add full-data shape tests for view widgets**
  Closes the 'partial' status on rows 15-19 of the controller coverage
  matrix by exercising the populated branches of each widget endpoint
  using the shared seeding fixtures from widgets/conftest.py:
  
  - summary: versioned_runs surfaces the seeded SCT run's version key;
    runs_results returns the seeded P99 read cell value/status.
  - pytest: a seeded pytest_v2 row (with user_fields) appears in
    result_filter hits + pieChart counts; user_fields endpoint returns
    the persisted SCYLLA_MODE entry.
  - nemesis_stats: a finalized SCTNemesis on a seeded run appears in
    nemesis_data with the run's scylla-server version.
  - graphs: seeding an ArgusGraphView surfaces the test name in
    tests_details.
  - graphed_stats: seeded run appears in test_runs; finalized nemesis
    appears in nemesis_data; runs_details returns version + linked
    GithubIssue via IssueLink.
  
  Fixture: package payload now includes date/revision_id/build_id (used
  by versioned_runs version key '<version>-<date>-<revision>') and
  test_method on sct_config (required for versioned_runs to surface
  the run).
  

- **test(testrun_api): add issues, Jenkins, S3 and terminate_stuck_runs coverage**
  Closes the 'partial' status on row 4 of the controller coverage matrix:
  - Issues endpoints (submit, event-scoped submit, get with full query
    arg parsing, get-missing-key error, delete) use mock_issue_service
    which patches IssueService in the controller module so neither
    PyGithub nor JIRA HTTP clients run.
  - All 9 Jenkins routes (params, build, queue_info +missing, clone
    targets/groups/create/build/settings/validate) use the
    mock_jenkins_service fixture.
  - S3-backed routes (log download, screenshot proxy) use mock_s3 to
    patch TestRunService.get_log/proxy_stored_s3_image; both controllers
    redirect(302) to the patched URL.
  - /terminate_stuck_runs runs against the live SCT model (empty
    result; only verifies response shape).
  
  Conftest now registers _helpers.external_mocks as a pytest_plugins
  entry so all subdirs can use the shared mocks without local imports.
  Fixed mock_s3 default for proxy_stored_s3_image to return a URL
  string (matches what the controller redirects to).
  

- **test(planner): add /plan/copy and /plan/trigger coverage**
  

- **test(auth): add password login, CF JWT happy path and full impersonation coverage**
  

- **test(profile): cover /profile/oauth/github callback and /profile/create flows**
  

- **test(view_api): smoke-test /views/stats empty/widget paths**
  

- **test(testrun_api): clean up seeded pytest rows to avoid cross-suite pollution**


soyacz:
ai review:
All 7 agents done. Here's the consolidated review:

---

## PR #982 Review — Backend Tests

### Critical

**`conftest.py` L191-194: Auth bypass may be broken at the root level.**
`load_logged_in_user` is likely registered as `app.before_request(load_logged_in_user)` — a reference, not a name lookup. Patching the module-level name after registration has no effect on the registered hook. If so, the real hook runs on every request. Verify this is working at all; add a test that confirms `g.user` is actually set per-request.

**`conftest.py` L197-203: `g.user` on app context doesn't survive into request contexts.**
Flask `g` is request-scoped. Setting it in `app_context` fixture doesn't propagate into `flask_client` requests. Either the auth path is broken or there's a different mechanism setting `g.user` per-request that should be documented.

**`test_graphs_widget.py`: Graph rendering code path never exercised.**
`graph_view_for_test` fixture seeds a `GraphView` but no `ArgusGenericResultData`. Controller only appends to `view_data` when graphs are non-empty — so `res["response"][test_id]` is always `[]`. The entire `get_test_graphs → filter → render` branch is untested dead code from the suite's perspective. Fix: seed result data so rendered graphs list is non-empty.

**`test_pytest_widget.py` L7-9: Tests a stub.**
`GET /api/v1/views/widgets/pytest/view` unconditionally returns `0`. Test is validating a placeholder, not a feature. Remove or mark `xfail` with explanation.

**`test_testrun_api.py` L449-463, L506-517: Mock-only tests.**
`test_issues_submit_invokes_service` and `test_issues_delete_invokes_service` assert `mock.assert_called_once()` with IssueService fully mocked. Nothing real is tested — proves controller wires the call, not that controller is correct. Add response body assertion or DB state verification.

---

### Recurring patterns (affects most files)

**No unauthenticated tests anywhere.**
Every file uses a pre-authenticated `flask_client`. `@api_login_required` and `@check_roles` decorators are never exercised in denying state. A single misconfigured decorator would go undetected. Needs at least one test per module hitting a guarded endpoint without auth.

**Fixtures that create via the endpoint under test** (`admin_release`, `admin_group_via_api`, `admin_test_via_api`).
If the endpoint is broken, all dependent tests fail as fixture errors, not test failures — harder to diagnose. Fix: use service layer for setup, endpoint for assertions.

**Empty-list smoke tests** (`test_release_versions_returns_list`, `test_user_jobs_returns_list`, `test_planned_jobs_returns_list`, `test_view_versions_for_empty_view`, `test_view_images_for_empty_view`, etc.).
Assert `isinstance(response, list)` on a freshly-created entity with no data. Only proves no 500, not correct behavior. Seed data and assert non-empty, or name them `_smoke` and leave a comment.

**User row leaks** (team_api, admin_users, testrun_api, notification_api).
`_make_user()` / `regular_user` fixture calls throughout without teardown. Rows accumulate in session-scoped DB. Not a correctness issue today, but `test_admin_users_returns_dict_without_password` and listing tests will see leaked users over time.

**Broad cleanup via `all().delete()`** (`test_team_api.py`, `test_planner_api.py` `cleanup_teams`/`cleanup_plans`).
Full-table scan deletes affect all data in the session-scoped DB, including rows created by other test modules. Track created IDs and delete only those.

**Duplicate `empty_view` fixture** in `test_graphed_stats_widget.py`, `test_graphs_widget.py`, `test_nemesis_stats_widget.py`, `test_summary_widget.py`.
Four identical fixture bodies differing only in prefix string. Move to `widgets/conftest.py` as a factory fixture.

---

### File-specific highlights

**`external_mocks.py`**: `mock_issue_service` yields the class mock (`mocks[0]`), not the instance. Callers must use `mock_issue_service.return_value.submit.assert_called()` — unintuitive. Yield `instance` directly.

**`test_admin_release_api.py`**: `test_admin_edit_release` doesn't verify `name` field is immutable after edit. `test_admin_delete_release` doesn't verify listing endpoint drops the entry.

**`test_admin_users_api.py`**: `test_admin_toggle_admin_self_forbidden` doesn't verify roles are unchanged after rejection. `test_admin_delete_admin_user_forbidden` doesn't assert promote actually succeeded before testing deletion rejection.

**`test_auth.py`**: `g.user = None` trick (L112) likely doesn't affect in-request `g`. No test for wrong password path. `load_logged_in_user` patch may have same root issue as conftest.

**`test_driver_matrix_api.py`**: `_build_xml` parses `passed="1"`, `failures="0"` but no test reads these back. `test_submit_driver_matrix_result_idempotent` doesn't verify data was written at all on first submit.

**`test_graphs_api.py`**: `test_test_results_with_date_range` asserts only `status == "ok"` — date filtering not verified at all. Could return unfiltered data and pass.

**`test_users_jobs_api.py`**: `test_resolve_artifact_size_non_s3_uses_http_head` — `mock_head.assert_called_once()` redundant; the response body assertion already proves it was called. The S3 branch of `resolve_artifact_size` is never tested.

**`test_testrun_api.py`**: All Jenkins tests (L524-623) only verify mock call forwarding and response shape against mocked return values. Label as wiring tests or add real state assertions.

**`test_view_api.py`**: `test_view_stats_for_empty_view_returns_ok` and `test_view_stats_with_widget_id_filter_returns_ok` — shape-only assertions (`isinstance(..., dict)`). Stats path is the most complex part of the view service; deserves key-level assertions.

**`test_planner_api.py`**: Happy path for `trigger_jobs` with a plan that HAS tests and should call Jenkins — never tested. Only the "no matching plan" and "no plans to trigger" branches are covered.

**`test_team_api.py` L168**: `team_member` fixture in function signature but never used in body. Remove.

**`widgets/conftest.py` L183-235**: `generic_results_for_run` bypasses service layer via direct `.save()`. If service enforces validation or sets derived fields, fixture creates data the service would never produce — widget tests may pass on artificial data.

**`widgets/conftest.py` L307-315**: `linked_github_issue` hardcodes `issue.number = 12345`. Two test runs hitting the same `(owner, repo, number)` will collision if there's a uniqueness constraint. Derive from uuid.

argus/backend/tests/_helpers/external_mocks.py
pehala, line 1: Conftest should be used for this. If you find out that the conftest would be too big, it indicates that we need a better test structure so we can nest fixtures in other conftests for specific areas where needed

argus/backend/tests/_helpers/external_mocks.py
```diff
@@ -0,0 +1,158 @@
...
+exercise real controller + Flask plumbing while never touching the network.
+
+Usage::
+
+    def test_something(flask_client, mock_issue_service):
+        mock_issue_service.return_value.submit.return_value = {"id": "abc"}
+        flask_client.post("/api/v1/issues/submit", json={...})
+        mock_issue_service.return_value.submit.assert_called_once()
+"""
+from __future__ import annotations
```
pehala: Not needed, if typing.TYPE_CHECKING is the preferred way

argus/backend/tests/_helpers/external_mocks.py
```diff
@@ -0,0 +1,158 @@
...
+
+    patchers = [patch(target, return_value=instance) for target in _ISSUE_SERVICE_PATCH_TARGETS]
+    mocks = [p.start() for p in patchers]
+    try:
+        # Return the first mock; ``.return_value`` is shared across all targets
+        # because we passed the same instance.
+        yield mocks[0]
+    finally:
+        for p in patchers:
+            p.stop()
```
pehala: use with

argus/backend/tests/_helpers/external_mocks.py
```diff
@@ -0,0 +1,158 @@
...
+    instance = MagicMock(name="IssueServiceInstance")
+    # Sensible defaults so tests that don't customize still get JSON-serializable
+    # responses.
+    instance.submit.return_value = {"id": "00000000-0000-0000-0000-000000000000"}
+    instance.submit_for_sct_event.return_value = {"id": "00000000-0000-0000-0000-000000000000"}
+    instance.get.return_value = []
+    instance.delete.return_value = True
+
+    patchers = [patch(target, return_value=instance) for target in _ISSUE_SERVICE_PATCH_TARGETS]
+    mocks = [p.start() for p in patchers]
```
pehala: not needed for single patch

argus/backend/tests/_helpers/external_mocks.py
```diff
@@ -0,0 +1,158 @@
...
+    instance.get_queue_info.return_value = {"why": None, "url": "http://jenkins.test/job/1/"}
+    instance.latest_build.return_value = {"number": 1}
+    instance.get_advanced_settings.return_value = {}
+    instance.clone_job.return_value = "cloned/job/path"
+    instance.verify_job_settings.return_value = True
+    instance.get_clone_targets.return_value = []
+    instance.get_clone_groups.return_value = []
+    instance.change_advanced_settings.return_value = True
+
+    patchers = [patch(target, return_value=instance) for target in _JENKINS_PATCH_TARGETS]
```
pehala: two with statements (one statement with two context managers) would work as well

argus/backend/tests/admin_api/test_admin_group_test_api.py
```diff
@@ -0,0 +1,336 @@
...
+
+import pytest
+
+
+ADMIN_PREFIX = "/admin/api/v1"
+API_PREFIX = "/api/v1"
+
+
+def _post(flask_client, path: str, payload: dict):
+    return flask_client.post(path, data=json.dumps(payload), content_type="application/json")
```
pehala: repeated across the entire module

argus/backend/tests/admin_api/test_admin_group_test_api.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+def _group_details(flask_client, group_id: str) -> dict:
+    resp = flask_client.get(f"{API_PREFIX}/group/{group_id}/details")
+    assert resp.status_code == 200
```
pehala: Assertions should not be hidden in functions, it is needesly obfuscated

argus/backend/tests/admin_api/test_admin_group_test_api.py
```diff
@@ -0,0 +1,336 @@
...
+
+def test_admin_get_groups_includes_created(flask_client, admin_release, admin_group_via_api):
+    resp = flask_client.get(
+        f"{ADMIN_PREFIX}/groups/get", query_string={"releaseId": str(admin_release.id)}
+    )
+    assert resp.status_code == 200, resp.data
+    body = resp.json
+    assert body["status"] == "ok"
+    ids = {g["id"] for g in body["response"]}
+    assert admin_group_via_api["id"] in ids
```
pehala: You should always assert full on respones, it reads better and it makes the test better

pehala (commented):
I did a quick review only, stopping after the first 3 files. I think the test have insufficient quality and should be improved. They do test what we need to, but are hard to read with suboptimal structure, making expansion harder.

Also if we provide good tests, which AI can read, it will produce much better tests in the future

The issues I pointed out are systematic patterns and are most likely repeated across the whole PR.

I would probably recommend doing smaller part of the tests first, which would undergo detailed review for structure, etc. and then based on that created bigger PR with rest of them, which should be much easier to review

argus/backend/tests/admin_api/test_admin_group_test_api.py
```diff
@@ -0,0 +1,348 @@
...
+
+# ---------------------------------------------------------------------------
+# /group/create + /groups/get + /group/details (round-trip)
+# ---------------------------------------------------------------------------
+
+def test_admin_create_group_round_trip(flask_client, admin_release, admin_group_via_api):
+    resp = flask_client.get(f"{API_PREFIX}/group/{admin_group_via_api['id']}/details")
+    assert resp.status_code == 200
+    body = resp.json
+    assert body["status"] == "ok"
```
pehala, line 86: isnt this the same as status_code? (applies for many other tests)

argus/backend/tests/admin_api/test_admin_group_test_api.py
```diff
@@ -0,0 +1,348 @@
...
+    assert body["response"]["name"] == admin_group_via_api["name"]
+    assert str(body["response"]["release_id"]) == str(admin_release.id)
```
pehala, line 88: I would assert entire response json, if possible. It make it easier to understand and provides a stronger test ((applies for many other tests))

argus/backend/tests/client_api/test_client_api.py
```diff
@@ -0,0 +1,255 @@
...
+- GET  /api/v1/client/testrun/<run_type>/<run_id>/get                        (get_run)
+- POST /api/v1/client/testrun/<run_type>/<run_id>/heartbeat                  (run_heartbeat)
+- GET  /api/v1/client/testrun/<run_type>/<run_id>/get_status                 (run_get_status)
+- POST /api/v1/client/testrun/<run_type>/<run_id>/set_status                 (run_set_status)
+- POST /api/v1/client/testrun/<run_type>/<run_id>/update_product_version    (run_update_product_version)
+- POST /api/v1/client/testrun/<run_type>/<run_id>/logs/submit                (run_submit_logs)
+- POST /api/v1/client/testrun/<run_type>/<run_id>/finalize                   (run_finalize)
+- POST /api/v1/client/testrun/pytest/result/submit                           (submit_pytest_result)
+- GET  /api/v1/client/testrun/pytest/<test_name>/stats/<field>/<aggr>        (get_pytest_test_field_stats)
+- POST /api/v1/client/testrun/report                                         (render_email_report)
```
soyacz, line 16: client api tests are not complete.
We load also client endpoints by this:
```
for plugin in AVAILABLE_PLUGINS.values():
    if plugin.controller:
        bp.register_blueprint(plugin.controller)
```

soyacz (approved):
I'm ok with these tests.
One thing about missing client tests - plugins (actually the most important ones) can go to separate PR

argus/backend/tests/widgets/conftest.py
```diff
@@ -0,0 +1,335 @@
...
+        package_version="6.0.0",
+        package_date="20260101",
+        package_revision_id="deadbeef",
+        test_method="widget_seed_module.WidgetSeedTest.test_widget",
+    )
+
+    try:
+        SCTTestRun.get(id=run_id).delete()
+    except SCTTestRun.DoesNotExist:
+        pass
```
pehala: Use finalizer instead, it will run even if the second POST fails, which this cleanup wont.

argus/backend/tests/widgets/conftest.py
```diff
@@ -0,0 +1,335 @@
...
+    yield {
+        "name": inserted["name"],
+        "id": inserted["id"],
+        "iso_id": inserted["id"].isoformat(),
+        "status": "passed",
+    }
+
+    # Best-effort cleanup; pytest tables are partition-keyed by name so
+    # a delete-by-PK is safe.
+    from argus.backend.models.pytest import PytestResultTable, PytestUserField
```
pehala: Inner imports should be removed

argus/backend/tests/auth/test_auth.py
```diff
@@ -0,0 +1,214 @@
...
+def secret_key(argus_app):
+    """auth endpoints rely on flask sessions which require a secret key."""
+    original = argus_app.secret_key
+    argus_app.secret_key = "test-secret-key"
+    yield
+    argus_app.secret_key = original
+
+
+@pytest.fixture
+def saved_g_user():
```
pehala, line 20: I would expect we create our own user and use that instead, that way the test is truly independent. Using global is in opinion fragile, and I think running tests in parallel wont work with it.

argus/backend/tests/auth/test_auth.py
```diff
@@ -0,0 +1,214 @@
...
+    user = User()
+    user.id = uuid.uuid4()
+    user.username = f"pw-user-{uuid.uuid4().hex[:8]}"
+    user.email = f"{user.username}@scylladb.com"
+    user.password = generate_password_hash(raw_password)
+    user.roles = ["ROLE_USER"]
+    user.save()
+
+    original_methods = current_app.config.get("LOGIN_METHODS", [])
+    current_app.config["LOGIN_METHODS"] = ["password"]
```
pehala, line 111: You should not edit per test config in individual tests, it might fail tests running it parallel

argus/backend/tests/auth/test_auth.py
```diff
@@ -0,0 +1,214 @@
...
+    original_g_user = g.user
+    g.user = None  # force the unauthenticated branch in /auth/login
```
pehala, line 113: Dont understand why you need to it , shoiuldnt test alwys start with a blank slate?

argus/backend/tests/team_api/test_team_api.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+@pytest.fixture(autouse=True)
+def cleanup_teams():
+    yield
+    for team in list(Team.all()):
+        team.delete()
+
+
+def _create_team(flask_client, name=None, leader_id=None, members=None):
```
pehala: make this a factory fixture and you can clean after yourself in it too

argus/backend/tests/sct_api/test_sct_api.py
```diff
@@ -367,3 +367,240 @@ def test_submit_and_get_junit_report(flask_client, sct_run_id):
...
+    body = resp.json
+    assert body["status"] == "ok"
+    events = body["response"]
+    assert isinstance(events, list)
+    assert len(events) == 1
+    event = events[0]
+    assert event["severity"] == "ERROR"
+    assert event["message"] == "Something went wrong on node-1"
+    assert event["event_type"] == "DatabaseLogEvent"
+    assert event["node"] == "node-1"
```
pehala: This should be a one assert with a dict, here you can clearly see how this is harder to read than
assert events = 
```suggestion
    assert events = [{
    "severity": ERROR",
    "message": "Something went wrong on node-1",
    "event_type": "DatabaseLogEvent",
    "node": "node-1"    
    }]
```

argus/backend/tests/notification_api/test_notification_api.py
```diff
@@ -0,0 +1,159 @@
...
+        source_type=source_type.value,
+        source_id=source_id or uuid.uuid4(),
+        title=title,
+        content=content,
+    ).save()
+
+
+@pytest.fixture
+def cleanup_notifications():
+    _clear_notifications(g.user.id)
```
pehala: If you have done the test isolation properly, no cleanup before the tests should be needed.

argus/backend/tests/api/test_users_jobs_api.py
```diff
@@ -0,0 +1,286 @@
...
+        assert resp.json["status"] == "error"
+    finally:
+        if saved_host is not None:
+            argus_app.config["ZEUS_HOST"] = saved_host
+
+
+def test_zeus_proxy_without_token_errors(flask_client, argus_app):
+    saved_host = argus_app.config.get("ZEUS_HOST")
+    saved_token = argus_app.config.pop("ZEUS_TOKEN", None)
+    argus_app.config["ZEUS_HOST"] = "localhost:9999"
```
pehala, line 231: You should not do this inside of a test, this should be done via a fixture and the isolation of the tests should be ensure. I suspect each test that requires config changes should have its own argus instances

argus/backend/tests/main_api/test_profile.py
```diff
@@ -0,0 +1,301 @@
...
+from argus.backend.models.web import User, WebFileStorage
+
+
+@pytest.fixture(autouse=True)
+def secret_key(argus_app):
+    """Profile endpoints rely on flask sessions / flash which require a secret key."""
+    original = argus_app.secret_key
+    argus_app.secret_key = "test-secret-key"
+    yield
+    argus_app.secret_key = original
```
pehala, line 17: Create your own key if you need it for a test.

pehala (commented):
I do not think the tests are ready. I have a few concerns:
1) Asserts should be cleaner IMO, especially for JSON responses we should compare entire dicts, instead of picking individual fields only. It leads to much more readable test with clear expecations and the test tests more, which is a good thing also.
2) Tests are not sandboxed properly. E.g. Usage of global user and or objects, or changing configs in middle of the test. Did you verify that it can run with xdist?
3) You have a lot of cleanup fixtures and reverting back changes, which should not be needed. Each fixture can and should cleanup after itself.

PR #984, opened 2026-04-24
feature(client): add stable SSH tunnel data-path routing

CodeLieutenant:
## Summary
- add a dedicated client SSH tunnel stack (`tunnel_models`, `tunnel_state`, `tunnel_api`, `tunnel_ssh`) and a compatibility facade in `argus/client/tunnel.py`.
- route Argus client data-path requests through `http://127.0.0.1:<local_port>` when tunnel mode is enabled, with strict temp-file host key verification (`ed25519` + `ecdsa`, no RSA), free-port allocation, reconnect, request-level retry, and warning-based fallback.
- thread `use_tunnel` through client constructors and both CLIs (`generic` + `driver-matrix`) and add focused tunnel tests covering API retries, strict host checking, bind conflicts, cooldown retry behavior, and request-level recovery.

## Validation
- `uv run pytest argus/client/tests`
- `uv run python -m py_compile argus/client/base.py argus/client/tunnel.py argus/client/tunnel_api.py argus/client/tunnel_models.py argus/client/tunnel_ssh.py argus/client/tunnel_state.py argus/client/generic/cli.py argus/client/driver_matrix_tests/cli.py argus/client/generic/client.py argus/client/driver_matrix_tests/client.py argus/client/sct/client.py argus/client/sirenada/client.py argus/client/tests/test_tunnel.py`

### DIFF
--- a/argus/client/base.py
+++ b/argus/client/base.py
@@ -53,6 +53,12 @@ class ArgusClient:
     def close(self) -> None:
         self.session.close()
 
+    def __enter__(self) -> "ArgusClient":
+        return self
+
+    def __exit__(self, exc_type, exc, tb) -> None:
+        self.close()
+
     @property
     def auth_token(self) -> str:
         return self._auth_token
@@ -105,10 +111,19 @@ class ArgusClient:
         }
 
     def get(self, endpoint: str, location_params: dict[str, str] = None, params: dict = None) -> requests.Response:
-        url = self.get_url_for_endpoint(endpoint=endpoint, location_params=location_params)
+        url = self.get_url_for_endpoint(
+            endpoint=endpoint,
+            location_params=location_params
+        )
         LOGGER.debug("GET Request: %s, params: %s", url, params)
-        response = self.session.get(url=url, params=params, headers=self.request_headers, timeout=self._timeout)
+        response = self.session.get(
+            url=url,
+            params=params,
+            headers=self.request_headers,
+            timeout=self._timeout
+        )
         LOGGER.debug("GET Response: %s %s", response.status_code, response.url)
+
         return response
 
     def post(
@@ -118,10 +133,20 @@ class ArgusClient:
         params: dict = None,
         body: dict = None,
     ) -> requests.Response:
-        url = self.get_url_for_endpoint(endpoint=endpoint, location_params=location_params)
+        url = self.get_url_for_endpoint(
+            endpoint=endpoint,
+            location_params=location_params
+        )
         LOGGER.debug("POST Request: %s, params: %s, body: %s", url, params, body)
-        response = self.session.post(url=url, params=params, json=body, headers=self.request_headers, timeout=self._timeout)
+        response = self.session.post(
+            url=url,
+            params=params,
+            json=body,
+            headers=self.request_headers,
+            timeout=self._timeout
+        )
         LOGGER.debug("POST Response: %s %s", response.status_code, response.url)
+
         return response
 
     def submit_run(self, run_type: str, run_body: dict) -> requests.Response:
--- a/argus/client/base.py
+++ b/argus/client/base.py
@@ -1,31 +1,18 @@
 import re
 import logging
-import os
-import threading
-import time
-import atexit
 from dataclasses import asdict
 from typing import Any, Type
 from uuid import UUID
 
 import requests
-from requests.adapters import HTTPAdapter
-from urllib3.util.retry import Retry
 
 from argus.common.enums import TestStatus
-from argus.client.tunnel import (
-    SSHTunnel,
-    TunnelConfig,
-    delete_cached_tunnel_state,
-    resolve_tunnel_config_with_reason,
-)
+from argus.client.session import create_session
 from argus.client.generic_result import GenericResultTable
 from argus.client.sct.types import LogLink
 
 JSON = dict[str, Any] | list[Any] | int | str | float | bool | Type[None]
 LOGGER = logging.getLogger(__name__)
-TUNNEL_COOLDOWN_SECONDS = 30
-DEFAULT_TUNNEL_MONITOR_INTERVAL_SECONDS = 5.0
 
 
 class ArgusClientError(Exception):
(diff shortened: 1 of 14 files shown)


CodeLieutenant:
Follow-up included in latest commit (`ecf3188`): added a background SSH tunnel monitor thread in `ArgusClient` to proactively detect dead tunnels and reconnect, plus `close()` lifecycle cleanup for monitor/tunnel/session resources.

argus/client/driver_matrix_tests/cli.py
```diff
@@ -16,22 +16,36 @@ def cli():
...
+        extra_headers=extra_headers,
+        use_tunnel=use_tunnel,
+    )
     client.submit_driver_result(
         driver_name=metadata["driver_name"], driver_type=metadata["driver_type"], raw_junit_data=base64.encodebytes(raw_xml))
     LOGGER.info("Done.")
 
 
-def _submit_driver_failure_internal(api_key: str, base_url: str, run_id: str, metadata_path: str, extra_headers: dict):
+def _submit_driver_failure_internal(api_key: str, base_url: str, run_id: str, metadata_path: str,
```
soyacz, line 37: I'm not sure if we need tunnels for cli clients like this.
It's just one request and that's it - creating tunnel takes longer and brings no benefits

argus/client/base.py
```diff
@@ -62,6 +85,206 @@ def __init__(self, auth_token: str, base_url: str, api_version="v1", extra_heade
...
             self.session.headers.update(extra_headers)
 
+        if self._use_tunnel:
+            self._start_tunnel_monitor()
+
+    @staticmethod
+    def _resolve_use_tunnel(use_tunnel: bool | None) -> bool:
+        if use_tunnel is not None:
+            return use_tunnel
+        return os.environ.get("ARGUS_USE_TUNNEL", "").strip().lower() in ("1", "true", "yes", "on")
```
soyacz: env variables are rather used to override default behavior

argus/client/base.py
```diff
@@ -62,6 +85,206 @@ def __init__(self, auth_token: str, base_url: str, api_version="v1", extra_heade
...
+
+    def _tunnel_monitor_loop(self) -> None:
+        while not self._monitor_stop_event.wait(self._tunnel_monitor_interval):
+            try:
+                self._monitor_tunnel_health_once()
+            except Exception:  # noqa: BLE001
+                LOGGER.debug("SSH tunnel monitor loop failed", exc_info=True)
+
+    def _monitor_tunnel_health_once(self) -> None:
+        if not self._use_tunnel:
```
soyacz: we should not use tunell monitor loop at all in that case.
Btw. this will kill the cpu as there's no sleep in the loop

argus/client/tunnel_models.py
```diff
@@ -0,0 +1,88 @@
...
+    private_key: Path
+    public_key: Path
+    key_meta: Path
+    config_cache: Path
+
+
+def parse_datetime(value: str) -> datetime:
+    if not value:
+        raise ValueError("datetime value is required")
+    normalized = value.replace("Z", "+00:00")
```
k0machi: I don't understand why this needs to be done.

argus/client/tunnel_models.py
```diff
@@ -0,0 +1,88 @@
...
+    parsed = datetime.fromisoformat(normalized)
+    if parsed.tzinfo is None:
+        return parsed.replace(tzinfo=UTC)
+    return parsed.astimezone(UTC)
+
+
+def to_iso_z(dt: datetime) -> str:
+    return dt.astimezone(UTC).isoformat().replace("+00:00", "Z")
```
k0machi: Ditto.

soyacz (changes requested):
Generally I don't like this solution. Too many changes in the base argus client.
Tunnel should be implemented more transparently, without many changes to the client code itself.
I propose to reuse `session` of `ArgusClient` to be normal session we had before, or `TunneledSession` when we use tunnel.
Creating TunneledSession should handle state of the tunnel, replacing url etc.
it must implement all the methods current `session` object uses. Actually, we could introduce new `ArgusSession` class that would select proper implementation based on config (strategy pattern).

argus/client/tunnel_models.py
```diff
@@ -0,0 +1,88 @@
...
+
+@dataclass(frozen=True, slots=True)
+class TunnelConfig:
+    proxy_host: str
+    proxy_port: int
+    proxy_user: str
+    target_host: str
+    target_port: int
+    host_key_fingerprint: str
+    expires_at: datetime | None = None
```
k0machi, line 71: In fact this is datetime? Why do we need to do all these format changes?

argus/client/tunnel_models.py
k0machi: Seems to be working fine with either format:

Image

argus/client/tunnel_models.py
```diff
@@ -0,0 +1,88 @@
...
+    key_id: str | None = None
+    tunnel_id: str | None = None
+
+    @classmethod
+    def from_api_response(cls, response: dict[str, Any]) -> "TunnelConfig":
```
k0machi: Replace type hint with TypedDict. You can reflect off of that for missing fields.

argus/client/tunnel_api.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+    return TunnelConfig.from_api_response(response)
+
+
+def _call_tunnel_api(method: str, url: str, auth_token: str, payload: dict[str, Any] | None) -> dict[str, Any]:
+    headers = {
+        "Authorization": f"token {auth_token}",
+        "Accept": "application/json",
+        "Content-Type": "application/json",
+    }
```
k0machi: Wouldn't it better to receive ArgusClient instance since it implements most of this logic already?

argus/client/tunnel_ssh.py
```diff
@@ -0,0 +1,305 @@
...
+        self.shutdown()
+        try:
+            known_hosts_path = self._prepare_known_hosts_file(config)
+        except TunnelClientError as exc:
+            reason = f"strict host verification failed: {exc}"
+            LOGGER.warning("SSH tunnel %s", reason)
+            return None, reason
+
+        for attempt in range(1, MAX_PORT_BIND_ATTEMPTS + 1):
+            reserve_socket, local_port = self._reserve_local_port()
```
k0machi, line 63: This potentially has a Time to use problem, but shouldn't be a problem,  I don't see a process suddenly getting reserved port yanked away in the several milliseconds it would take to establish the tunnel.

argus/client/tunnel_ssh.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def _register_atexit(self) -> None:
+        if self._atexit_registered:
+            return
+        atexit.register(self.shutdown)
+        self._atexit_registered = True
+
+    def _build_ssh_command(self, config: TunnelConfig, local_port: int, known_hosts_path: Path) -> list[str]:
+        return [
+            "ssh",
```
k0machi: Should we make this an absolute path or a list of allowed paths? On the off chance the default PATH executable is something else.

k0machi (commented):
Minor comments

argus/client/base.py
k0machi: It has a wait interval using a stop event in `_tunnel_monitor_loop`.

argus/client/generic/cli.py
```diff
@@ -46,26 +53,40 @@ def submit_run(api_key: str, base_url: str, id: str, build_id: str, build_url: s
...
 @click.option("--base-url", default="https://argus.scylladb.com", help="Base URL for argus instance")
+@click.option("--use-tunnel/--no-use-tunnel", default=None, help="Route API calls through SSH tunnel")
 @click.option("--version", help="Scylla version to filter plans by", default=None, required=False)
 @click.option("--plan-id", help="Specific plan id for filtering", default=None, required=False)
 @click.option("--release", help="Release name to filter plans by", default=None, required=False)
 @click.option("--job-info-file", required=True, help="JSON file with trigger information (see detailed docs)")
 @click.option("--extra-headers", default={}, type=click.UNPROCESSED, callback=validate_extra_headers, help="extra headers to pass to argus, should be in json format", envvar='ARGUS_EXTRA_HEADERS')
-def trigger_jobs(api_key: str, base_url: str, job_info_file: str, version: str, plan_id: str, release: str, extra_headers: dict | None = None):
-    client = ArgusGenericClient(auth_token=api_key, base_url=base_url, extra_headers=extra_headers)
+def trigger_jobs(api_key: str, base_url: str, use_tunnel: bool | None, job_info_file: str, version: str,
```
soyacz, line 82: as I said, we don't need tunnels for single request - it misses its purpose and adds complexity. 
I'd prefer having it removed from client cli commands

argus/client/driver_matrix_tests/cli.py
```diff
@@ -16,22 +16,34 @@ def cli():
     pass
 
 
-def _submit_driver_result_internal(api_key: str, base_url: str, run_id: str, metadata_path: str, extra_headers: dict):
+def _submit_driver_result_internal(api_key: str, base_url: str, run_id: str, metadata_path: str,
```
soyacz, line 19: these formatting changes are not needed, better to remove to reduce PR size

argus/client/tunnel_state.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def generate_keypair_if_needed(paths: TunnelStatePaths) -> None:
+    if is_key_valid(paths):
+        return
+
+    if _generate_keypair_with_cryptography(paths):
+        return
+
+    if shutil.which("ssh-keygen") is None:
```
CodeLieutenant, line 62: i'll check

soyacz (commented):
Generally looks ok - I'd prefer removing redundant formatting changes, client cli using tunnels
besides, AI review:
Bugs / correctness (all low severity)
- TunneledSession not actually requests.Session-compatible (session.py:63,244) — only get/post/headers/mount/close implemented. Any future use of put/request/hooks/cookies will AttributeError. Either subclass requests.Session or narrow the docstring.
- Monitor thread outlives client (session.py:81) — daemon thread polls forever; no CLI calls client.close(). Combined with #1 above, tunnel-on-CLI keeps re-spawning SSH.
- Lock can stall request thread up to ~100s (session.py:187-192) — monitor's _ensure_tunnel holds RLock during reconnect. Behaviorally correct, worth documenting.
- _call_tunnel_api assumes dict response (tunnel_api.py:175) — .get on list/scalar raises AttributeError, escapes the TunnelClientError catch in resolve_tunnel_config_with_reason. Add isinstance(..., dict) check.
- Cache wiped on every transient establish failure (session.py:153) — delete_cached_tunnel_state() regenerates keypair on each 30s cooldown. Consider only invalidating keypair when server reports unknown key.
- Duplicated retry/fallback logic in get/post (session.py:194-226) — copy-paste candidate.
Style
- tunnel_models.py:48 uses TypedDict __required_keys__ dunder — add a comment.
- tunnel_state.py:91 chmods public key 0o600 (conventionally 0o644).
- _TunnelApiResponse annotation now used for both API responses and cache payloads — slightly inaccurate.
Test gaps
test_tunnel.py doesn't cover the new TunneledSession recovery logic introduced in e5d8d8a:
- retry-then-backoff path in get/post
- monitor thread lifecycle
- delete_cached_tunnel_state on establish failure

soyacz (approved):

argus/client/tunnel_ssh.py
CodeLieutenant: i dont thinks so either, especially on a machine that only does one thing, possible if someone opens a tcp connection and port becomes reserved, but that also takes longer then a microseconds between check and open

soyacz:
(approve conditional - pls address comments)

argus/client/generic/cli.py
CodeLieutenant: so in generic and driver we dont need it?

CodeLieutenant:
> (approve conditional - pls address comments)

updating it ASAP

argus/client/session.py
```diff
@@ -0,0 +1,263 @@
...
+        if value <= 0:
+            raise ValueError("interval must be positive")
+        return value
+    except ValueError:
+        LOGGER.warning(
+            "Invalid ARGUS_TUNNEL_MONITOR_INTERVAL=%r, using default %.1fs",
+            raw,
+            _DEFAULT_MONITOR_INTERVAL,
+        )
+        return _DEFAULT_MONITOR_INTERVAL
```
k0machi, line 45: Is this really needed?

k0machi (approved):
Overall LGTM, minor question.

CodeLieutenant:
:green_circle: https://argus.scylladb.com/tests/scylla-cluster-tests/479c6fef-dd2d-42eb-99cc-7a6d61b16989

Test proving everything is working:

```
15:59:08  Initializing Argus connection...
15:59:08  Skipping setting up argus events: You should create default EventsProcessRegistry first
15:59:09  SSH tunnel established: proxy=TUNNEL_IP:22, user=argus-proxy, key_id=abfb9977-05a5-4785-b7e1-9e8b691495e0, local_port=54769
```

soyacz:
@CodeLieutenant do you know why there are two lines like:
```
< t:2026-05-15 14:02:33,543 f:session.py      l:193  c:argus.client.session p:INFO  > SSH tunnel established: proxy=13.51.200.137:22, user=argus-proxy, key_id=17c7e527-7c88-40b5-a977-4334c38d874d, local_port=57271
[...]
< t:2026-05-15 14:02:35,203 f:session.py      l:193  c:argus.client.session p:INFO  > SSH tunnel established: proxy=13.51.200.137:22, user=argus-proxy, key_id=17c7e527-7c88-40b5-a977-4334c38d874d, local_port=34759
```
maybe we create client twice in Argus?
Anyway, it tunnel should be reused

k0machi:
Also, I think we should add distinguishing the traffic method into this PR as well - tag the traffic using SSH tunnel with a custom header called `X-Argus-Proxied-Connection: 1` and maybe add the metric collection to the backend metrics module.

soyacz:
> Also, I think we should add distinguishing the traffic method into this PR as well - tag the traffic using SSH tunnel with a custom header called `X-Argus-Proxied-Connection: 1` and maybe add the metric collection to the backend metrics module.

@CodeLieutenant please address this one and is good to go

soyacz:
I found the reason of double client creation: https://scylladb.atlassian.net/browse/SCT-377

k0machi:
> I found the reason of double client creation: https://scylladb.atlassian.net/browse/SCT-377

Check my note there as well - might be necessary to be kept as is, in this case - please make sure the client does correct clean up when going out of scope.

argus/client/session.py
CodeLieutenant: yes

CodeLieutenant:
> > Also, I think we should add distinguishing the traffic method into this PR as well - tag the traffic using SSH tunnel with a custom header called `X-Argus-Proxied-Connection: 1` and maybe add the metric collection to the backend metrics module.
> 
> @CodeLieutenant please address this one and is good to go

I dont think we need it, since i've added long time ago

https://github.com/scylladb/argus/pull/984/changes#diff-5fcf011f6c5fa1ae67525290f612093bd08a4a1e2c587eec54d4678826e7ed7dR244-R247

```
headers = {
    "User-Agent": "argus-client-ssh-tunnel",
    "X-SSH-Tunnel-Origin": self._tunnel_config.proxy_host,
    "X-Tunnel-Established-At": self._tunnel_established_at or "",
}
```

CodeLieutenant:
🟢 https://argus.scylladb.com/tests/scylla-cluster-tests/395f89ac-38a8-407c-b69e-4e37332c93c2 - GCE Run
🟢 https://argus.scylladb.com/tests/scylla-cluster-tests/b454ee5f-b24f-405c-93a7-a45511a86921 - Azure Run

🟢 https://argus.scylladb.com/tests/scylla-cluster-tests/21a165a4-b4f5-4895-a375-9ef40569432b - Test Without SSH Tunnel

I could not start Oracle Cloud tests, there is no capacity.

k0machi (approved):

PR #985, opened 2026-04-27
feat(ssh-tunnel): admin CRUD, key registration fixes, and provisioning script

CodeLieutenant:
## Summary

- **Backend**: store full `known_hosts` entry (discovered via `ssh-keyscan`) instead of raw SHA256 fingerprint, so the Go client can write a valid `known_hosts` file; remove `host_key_fingerprint` from the create payload entirely; add tunnel config delete endpoint; allow `Admin` role on `GET /client/ssh/keys`; fix `check_roles` to return JSON 403 for API views instead of silently redirecting; split `ProxyTunnelConfigDTO` / `ProxyTunnelCreateResponseDTO` so `api_token` never leaks into list/get responses
- **CLI**: add full admin CRUD for proxy tunnel configs and SSH key management (`ssh tunnel list/get/create/set-active/delete`, `ssh keys inspect/delete`); fix deadlock when SSH exits early (Permission denied) by using a non-blocking drain on `waitErrC`; run SSH subprocess in its own process group for clean Ctrl+C handling; `ssh connect` now always registers the public key on every invocation
- **Provisioning**: rename `config.sh` → `tunnel-server-setup.sh`; embed `ARGUS_TOKEN` directly in `argus-authorized-keys` (sourcing an env file does not work under `AuthorizedKeysCommandUser nobody`); all config via flags (`--token`, `--target-host`, `--target-port`, etc.)

## Test plan

- [x] `argus ssh tunnel create --host <ip> --target-host <private-ip>` creates config and returns `api_token`
- [x] `argus ssh tunnel list` / `get` / `set-active` / `delete` work as admin
- [x] `argus ssh keys list` returns authorized keys for both Admin and SSHTunnelServer users
- [x] `argus ssh keys inspect` shows registered keys after `ssh connect`
- [x] `argus ssh connect` registers key on every call and exits cleanly on Permission denied (no hang)
- [x] Ctrl+C on `ssh connect` terminates immediately
- [x] `sudo bash tunnel-server-setup.sh --token <tok> --target-host <ip>` provisions the proxy host end-to-end
- [x] `sudo -u nobody /usr/local/bin/argus-authorized-keys` returns registered public keys on the proxy host


argus/backend/controller/admin_api.py
```diff
@@ -417,6 +418,14 @@ def save_proxy_tunnel_config():
     }
 
 
+@bp.route("/proxy-tunnel/config/<string:tunnel_id>", methods=["DELETE"])
+@check_roles(UserRoles.Admin)
+@api_login_required
+def delete_proxy_tunnel_config(tunnel_id: str):
+    TunnelService().delete_proxy_tunnel_config(UUID(tunnel_id))
```
k0machi, line 425: Should probably also invalidate the keys for this tunnel.

argus/backend/service/tunnel_service.py
```diff
@@ -382,6 +415,24 @@ def list_proxy_tunnel_configs(self, active_only: bool | None = None) -> list[Pro
...
+    def delete_proxy_tunnel_config(self, tunnel_id: UUID | str) -> None:
+        """Permanently delete a proxy tunnel config and its associated service user."""
+        if not isinstance(tunnel_id, UUID):
+            tunnel_id = UUID(str(tunnel_id))
+        try:
+            config = ProxyTunnelConfig.get(id=tunnel_id)
+        except ProxyTunnelConfig.DoesNotExist as exc:
+            raise TunnelServiceException(f"Proxy tunnel config {tunnel_id} not found") from exc
+
+        if config.service_user_id:
```
k0machi, line 434: This can be dangerous - I suggest just invalidating the token or removing the SSH Tunnel role. IIRC This user is created automatically when a proxy tunnel is created, so it should be safe in 90% of the cases and I think you can't manually assign a user there.

cli/cmd/ssh.go
```diff
@@ -176,6 +184,25 @@ func init() {
...
+			seen[line] = struct{}{}
+			out = append(out, line)
+		}
+	}
+	if len(out) == 0 {
+		return ""
+	}
+	return strings.Join(out, "\n") + "\n"
+}
+
```
k0machi, line 205: Is it necessary to dedup it twice?

argus/backend/controller/admin_api.py
CodeLieutenant: we dont have keys for tunnel, every tunnel has every key, invalidating some keys will mess up other, keys are per user

cli/cmd/ssh.go
CodeLieutenant: unfortunatly yes, ssh failed if we have duplicated keys

k0machi (approved):

PR #987, opened 2026-04-29
docs(agents): add key files table and dev-setup reference to AGENTS.md

pehala:
Helps with orienting. I noticed agents tend to not read dev-setup and spend many actions trying to rediscover how to run backend and frontend


soyacz (approved):
I wonder if we should not create symlink to 'CLAUDE.md' which is used by both opencode and Claude.

pehala:
> I wonder if we should not create symlink to 'CLAUDE.md' which is used by both opencode and Claude.

Might be good idea, the files should be in sync, however skills optimized for one dont have to be as great for the other, but I suspect that will be only a minor issue

PR #988, opened 2026-04-29
feat(admin-panel): add SSH tunnel management UI

CodeLieutenant:
Adds a new admin section for managing proxy tunnel configs and registered SSH keys, wired into the AdminPanel sidebar. The component talks to the existing /admin/api/v1/proxy-tunnel/* and /admin/api/v1/ssh/keys endpoints, surfaces the one-time service-user API token on create, and supports enable/disable + delete on configs and revoke on keys.

Depends on #985


frontend/AdminPanel/AdminPanel.svelte
```diff
@@ -13,17 +14,20 @@
         users: UserManager,
         releases: ReleaseManager,
         views: ViewsManager,
+        "ssh-tunnel": ProxyTunnelManager,
     };
 
+    const routeLabel = (route) => (route === "ssh-tunnel" ? "SSH Tunnel" : titleCase(route));
```
k0machi: I suggest either using spaces in the key itself (so `"SSH Tunnel": ProxyTunnelManager`) or adding an exportable prop called `documentTitle` into each manager component.

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,514 @@
...
+        </div>
+    {/if}
+    {#if lastNotice}
+        <div class="alert alert-success d-flex align-items-center">
+            <div>{lastNotice}</div>
+            <div class="ms-auto">
+                <button class="btn btn-sm" aria-label="Dismiss notice" onclick={() => (lastNotice = "")}><Fa icon={faTimes} /></button>
+            </div>
+        </div>
+    {/if}
```
k0machi: Use `sendMessage` error reporting as it already includes both the toast and the console to view these. Also, for exceptions that are not well known it might be better to also output trace to the console

k0machi:
Can I see the visual demo of this interface?

soyacz:
@CodeLieutenant don't forget about this one

CodeLieutenant:
https://github.com/user-attachments/assets/a3c82bec-3ab9-44f4-a4b0-9aba412bbefd

frontend/AdminPanel/AdminPanel.svelte
```diff
@@ -73,6 +76,16 @@
...
                     </button>
                 </li>
+                <li class="list-group-item">
+                    <button
+                        class:btn-primary={currentRoute == "ssh-tunnel"}
+                        class:btn-outline-primary={currentRoute != "ssh-tunnel"}
+                        class="btn h-100 w-100"
+                        onclick={(e) => handleRouteClick("ssh-tunnel", e)}
+                    >
+                        SSH Tunnel
```
k0machi: Can be `routeTitle` as well

frontend/AdminPanel/AdminPanel.svelte
```diff
@@ -73,6 +76,16 @@
                         Views
```
k0machi: but do make sure to update them all

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,524 @@
...
+        if (!data) {
+            throw new Error(`HTTP ${response.status} on ${url}`);
+        }
+        if (data.status !== "ok") {
+            const message =
+                data.message ?? data.response?.arguments?.[0] ?? data.response?.message ?? `HTTP ${response.status}`;
+            throw new Error(message);
+        }
+        return data.response;
+    };
```
k0machi: Seems pretty general - I'd add signatures and move to the generic functions in the Common directory.

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,524 @@
+<script>
```
k0machi: Not a typescript module - all new components should use TypeScript

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,524 @@
...
+
+        {#await configsPromise}
+            <div class="p-3 text-center text-muted">
+                <span class="spinner-border spinner-border-sm"></span> Loading proxy tunnels...
+            </div>
+        {:then configs}
+            {#if !configs || configs.length === 0}
+                <div class="p-3 text-center text-muted">No proxy tunnel configs.</div>
+            {:else}
+                <div class="table-responsive">
```
k0machi: I would not use `table` at all here and reimplent using flexbox - check other places that use table-like elements, like Pytest.svelte

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,524 @@
...
+        {#await keysPromise}
+            <div class="p-3 text-center text-muted">
+                <span class="spinner-border spinner-border-sm"></span> Loading keys...
+            </div>
+        {:then keys}
+            {#if !keys || keys.length === 0}
+                <div class="p-3 text-center text-muted">No active SSH keys registered.</div>
+            {:else}
+                <div class="table-responsive">
+                    <table class="table table-sm align-middle mb-0">
```
k0machi: ditto

k0machi (changes requested):

frontend/Common/ProxyUtils.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (data.status !== "ok") {
+        const message =
+            (data as any).message ??
+            (data.response as any)?.arguments?.[0] ??
+            (data.response as any)?.message ??
+            `HTTP ${response.status}`;
+        throw new Error(message);
+    }
+    return data.response;
+};
```
k0machi: Not a ProxyUtil

frontend/Common/ProxyUtils.ts
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Truncate a string, keeping `head` chars from the start and `tail` chars from the end,
+ * joined by an ellipsis. Returns "—" for empty/null values.
+ */
+export const truncate = (value: string | null | undefined, head = 14, tail = 8): string => {
+    if (!value) return "—";
+    const trimmed = String(value).trim();
+    if (trimmed.length <= head + tail + 1) return trimmed;
+    return `${trimmed.slice(0, head)}…${trimmed.slice(-tail)}`;
+};
```
k0machi: Also doesn't belong in this file

frontend/Common/ProxyUtils.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export const copyToClipboard = async (value: string, label: string, context = "copyToClipboard"): Promise<void> => {
+    try {
+        await navigator.clipboard.writeText(value);
+        sendMessage("success", `${label} copied to clipboard.`, context);
+    } catch (error: any) {
+        console.error(context, error);
+        sendMessage("error", `Could not copy ${label.toLowerCase()}: ${error.message}`, context);
+    }
+};
```
k0machi: I think this function might already exist in the codebase

frontend/Common/ProxyUtils.ts
```diff
@@ -0,0 +1,67 @@
+import { sendMessage } from "../Stores/AlertStore";
+
+export interface ApiResponse<T = unknown> {
+    status: string;
+    response: T;
+    message?: string;
+}
```
k0machi: Also not a proxy util

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,531 @@
...
+    .config-row .col-status  { flex: 0 0 8%;  min-width: 0; }
+    .config-row .col-actions { flex: 0 0 16%; min-width: 0; }
+
+    /* Key table columns */
+    .key-row .col-fp         { flex: 0 0 28%; min-width: 0; }
+    .key-row .col-tunnel     { flex: 0 0 16%; min-width: 0; }
+    .key-row .col-keyuser    { flex: 0 0 16%; min-width: 0; }
+    .key-row .col-created    { flex: 0 0 14%; min-width: 0; }
+    .key-row .col-expires    { flex: 0 0 14%; min-width: 0; }
+    .key-row .col-keyactions { flex: 0 0 12%; min-width: 0; }
```
k0machi: I don't like this specificity, it should be enough to do work with standard flex tools available everywhere, e.g. using `mx-n` or `px-n` utility components for padding.

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,531 @@
...
+            {:else}
+                <!-- Header row -->
+                <div class="d-flex align-items-center px-3 py-1 border-bottom text-muted small fw-semibold key-row">
+                    <div class="col-fp font-monospace">Fingerprint</div>
+                    <div class="col-tunnel">Tunnel</div>
+                    <div class="col-keyuser">User</div>
+                    <div class="col-created">Created</div>
+                    <div class="col-expires">Expires</div>
+                    <div class="col-keyactions ms-auto text-end">Actions</div>
+                </div>
```
k0machi: It doesn't have to be a table, could be just a collection of rows with a custom element

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,531 @@
...
+                <!-- Header row -->
+                <div class="d-flex align-items-center px-3 py-1 border-bottom text-muted small fw-semibold config-row">
+                    <div class="col-host font-monospace">Host</div>
+                    <div class="col-user">Proxy User</div>
+                    <div class="col-target font-monospace">Target</div>
+                    <div class="col-hostkey">Host Key</div>
+                    <div class="col-svcuser">Service User</div>
+                    <div class="col-status">Status</div>
+                    <div class="col-actions ms-auto text-end">Actions</div>
+                </div>
```
k0machi: Ditto

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,531 @@
...
+    const formatDate = (raw: string | null | undefined): string => {
+        if (!raw) return "—";
+        const ts = Date.parse(raw);
+        if (Number.isNaN(ts)) return raw;
+        return timestampToISODate(ts, true);
+    };
+
+    const formIsValid = $derived(
+        Boolean(createForm.host && createForm.target_host && createForm.proxy_user && createForm.port && createForm.target_port),
+    );
```
k0machi: This doesn't really validate the form - all it checks is if the value is truthy, nothing stops me here from putting garbage into `target_host` for example. You can check mechanisms I use for forms in `JenkinsRebuildForm` for one way to do validation potentially. Or just omit client-side validation altogether and use bootstrap validation utilities

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,531 @@
...
+            selectedKey = null;
+        }
+    };
+
+    const formatDate = (raw: string | null | undefined): string => {
+        if (!raw) return "—";
+        const ts = Date.parse(raw);
+        if (Number.isNaN(ts)) return raw;
+        return timestampToISODate(ts, true);
+    };
```
k0machi: `timestampToISODate` function exists.

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,531 @@
...
+            deleteServiceUser = false;
+        }
+    };
+
+    const submitDeleteKey = async () => {
+        if (!selectedKey) return;
+        try {
+            working = true;
+            await fetchJson(`${ADMIN_API}/ssh/keys/${selectedKey.key_id}`, { method: "DELETE" });
+            sendMessage("success", "SSH key revoked.", "ProxyTunnelManager::submitDeleteKey");
```
k0machi: I would avoid sending toasts on successes, they are quite annoying when you see a lot of them.

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,531 @@
...
+                        class="btn btn-primary w-75 me-2"
+                        disabled={working || !formIsValid}
+                        onclick={submitCreate}
+                    >
+                        {#if working}<span class="spinner-border spinner-border-sm me-2"></span>{/if}
+                        Save & Provision
+                    </button>
+                    <button class="btn btn-secondary w-25" disabled={working} onclick={resetForm}>Cancel</button>
+                </div>
+            </div>
```
k0machi: This probably should be a separate component. Also, using row-column bootstrap system with just `col-md-12` is unnecessary, you can use normal block flow there.

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,531 @@
...
+                        bind:checked={deleteServiceUser}
+                    />
+                    <label class="form-check-label" for="deleteServiceUser">
+                        Also delete the service user
+                    </label>
+                </div>
+                {#if deleteServiceUser}
+                    <div class="alert alert-warning py-2 mt-2 small mb-0">
+                        The service user <code>proxy-tunnel-{deleteTarget.host}</code> will be permanently removed.
+                    </div>
```
k0machi, line 234: `proxy_user` can be specified in the create form, which means this hardcoding can desync.

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,531 @@
...
+                    <div class="col-fp font-monospace">Fingerprint</div>
+                    <div class="col-tunnel">Tunnel</div>
+                    <div class="col-keyuser">User</div>
+                    <div class="col-created">Created</div>
+                    <div class="col-expires">Expires</div>
+                    <div class="col-keyactions ms-auto text-end">Actions</div>
+                </div>
+                <ul class="list-group list-group-flush rounded">
+                    {#each keys as key (key.key_id)}
+                        <li class="list-group-item">
```
k0machi: ul > li unnecessary.

k0machi (changes requested):

frontend/Common/ProxyUtils.ts
k0machi: Also I think there's a CSS utility class in bootstrap for this, `text-truncate`

https://getbootstrap.com/docs/5.2/helpers/text-truncation/

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,531 @@
...
+    let toggleTarget = $state<TunnelConfig | null>(null);
+
+    let confirmDeleteConfig = $state(false);
+    let deleteTarget = $state<TunnelConfig | null>(null);
+    let deleteServiceUser = $state(false);
+
+    let activeFilter = $state("all");
+
+    const loadConfigs = async (): Promise<TunnelConfig[]> => {
+        const params = new URLSearchParams();
```
k0machi: `query-string` package.

argus/backend/controller/admin_api.py
```diff
@@ -422,8 +422,9 @@ def save_proxy_tunnel_config():
 @check_roles(UserRoles.Admin)
 @api_login_required
 def delete_proxy_tunnel_config(tunnel_id: str):
-    TunnelService().delete_proxy_tunnel_config(UUID(tunnel_id))
-    return {"status": "ok", "response": {"deleted": True}}
+    delete_user = request.args.get("delete_user", "").lower() == "true"
```
k0machi: Not quite idiomatic in my opinion, sending `1` for true and `0` for false is usually how we do thing in our API inside query string values

k0machi:
May I see a demo of the current iteration?

frontend/AdminPanel/ProxyTunnelManager.svelte
```diff
@@ -0,0 +1,439 @@
...
+            {:else}
+                <div class="row g-0 align-items-center px-3 py-1 border-bottom text-muted small fw-semibold">
+                    <div class="col-2">Host</div>
+                    <div class="col-2">Proxy User</div>
+                    <div class="col-2">Target</div>
+                    <div class="col-2">Host Key</div>
+                    <div class="col-1">Svc User</div>
+                    <div class="col-1">Status</div>
+                    <div class="col-2 text-end">Actions</div>
+                </div>
```
k0machi: Still the same problem from my perspective - this is too many columns. I would re-arrange this element into a multiline custom one of some sort that is similar to what we have for example in pytest or driver matrix tests.

PR #992, opened 2026-05-11
chore(plugins/sct): Migration script for legacy events and tab reoder

k0machi:
This commit adds a migration script for SCTTestRun.events field,
extracting plaintext events and processing them into fully parsed
SCTEvent table events, as well as putting any ERROR and CRITICAL events
found into SCTUnprocessedEvent event queue.

Additionally, this commit includes a script to drop events,
allocated_resources and nemesis_data columns from the primary SCTTestRun
table, to be executed after confirming the migration.

Fixes ARGUS-55

### DIFF
--- a/scripts/migration/migration_2026-05-08_drop_columns.py
+++ b/scripts/migration/migration_2026-05-08_drop_columns.py
@@ -1,0 +1,29 @@
+import logging
+
+from argus.backend.db import ScyllaCluster
+from argus.backend.plugins.sct.testrun import SCTTestRun
+from argus.backend.util.logsetup import setup_application_logging
+
+
+setup_application_logging(log_level=logging.INFO)
+LOGGER = logging.getLogger(__name__)
+DB = ScyllaCluster.get()
+
+COLUMNS_TO_DROP = ["allocated_resources", "nemesis_data", "events"]
+
+
+def migrate():
+    table_name = SCTTestRun.table_name()
+    LOGGER.warning("Starting migration: dropping columns %s from %s...", COLUMNS_TO_DROP, table_name)
+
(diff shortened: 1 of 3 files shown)


soyacz (commented):
as we spoke: `SCTUnprocessedEvent` not needed - new events are long enough and in all supported branches

scripts/migration/migration_2026-05-08_drop_columns.py
```diff
@@ -0,0 +1,29 @@
...
+from argus.backend.db import ScyllaCluster
+from argus.backend.plugins.sct.testrun import SCTTestRun
+from argus.backend.util.logsetup import setup_application_logging
+
+
+setup_application_logging(log_level=logging.INFO)
+LOGGER = logging.getLogger(__name__)
+DB = ScyllaCluster.get()
+
+COLUMNS_TO_DROP = ["allocated_resources", "nemesis_data", "events"]
```
soyacz, line 12: we can't drop these columns until we remove old ways of publishing events (i think resources and nemesis data should not be affected)

soyacz (commented):
Just a remark, that we should not drop events table until we patch `argus.backend.plugins.sct.service.SCTService.submit_events`
for the safety, we should already drop saving old events (leaving api for compatibility)

k0machi:
> Just a remark, that we should not drop events table until we patch `argus.backend.plugins.sct.service.SCTService.submit_events`
> for the safety, we should already drop saving old events (leaving api for compatibility)

Of course, I am currently thinking if we should completely dummy this function out or try to convert sent events into new ones - skipping that if we already have new events in the DB.

PR #995, opened 2026-05-12
fix(cli-release): extract inline workflow scripts and fix first-release tag resolution

CodeLieutenant:
grep -v exits 1 when no lines match; with bash pipefail enabled this killed the Resolve previous CLI tag step on the first release. Fix by wrapping grep in a subshell with || true. Also extract all inline bash blocks into versioned scripts under cli/scripts/ for easier local testing.


k0machi:
Does this also fix the issue of GoReleaser expecting non-prefixed git tag?

CodeLieutenant:
> Does this also fix the issue of GoReleaser expecting non-prefixed git tag?

it is now

k0machi (approved):

PR #999, opened 2026-05-13
fix(plugins/sct): Disable old events submit endpoint

k0machi:
Since we are dropping those columns, disable the old submit events
endpoint by making it a no-op.


soyacz (commented):
LGTM

soyacz (approved):

k0machi:
Also need to remove the tests for this endpoint before merge.

soyacz (approved):

PR #1000, opened 2026-05-13
improvement(cli): Add Pytest Results command

k0machi:
This PR adds both a backend API endpoint and a Go Client command to
fetch test results globally using the provided user filters, same as
frontend implementation for the Pytest View Widget.

### DIFF
--- a/argus/backend/controller/views_widgets/pytest.py
+++ b/argus/backend/controller/views_widgets/pytest.py
@@ -41,6 +41,17 @@ def get_view_pytest_results(view_id: str):
         "response": res
     }
 
+@bp.route("/pytest/results", methods=["GET"])
+@api_login_required
+def get_pytest_results():
+    service = PytestViewService()
+    res = service.result_filter()
+    return {
+        "status": "ok",
+        "response": res
+    }
+
+
 @bp.route("/pytest/<path:test_name>/<string:id>/fields", methods=["GET"])
 @api_login_required
 def get_user_fields_for_test(test_name: str, id: str):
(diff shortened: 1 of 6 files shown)


CodeLieutenant (approved):

soyacz:
## Code Review Summary: PR #1000 – Add Pytest Results command

### Bugs

1. **`limit` is a string, used as slice index** (`argus/backend/service/views_widgets/pytest.py:122,192`)
   `request.args.get("limit", 500)` returns a string when the query param is present, but the default is an int. At line 192, `results[:limit]` will raise `TypeError` because Python can't slice with a string. The new CLI always sends `limit=500`, so this will now consistently trigger. Fix: `limit = int(request.args.get("limit", 500))`.

2. **`before`/`after` type inconsistency** (`argus/backend/service/views_widgets/pytest.py:133-138,190`)
   These variables are reassigned from strings to `datetime` objects when provided, then passed to `prepare_bar_chart`. Pre-existing issue, but the CLI makes it easier to hit.

### Minor Issues

3. **`TestID` silently dropped from table output** (`cli/internal/models/tests.go`)
   The `PytestFilterHit` struct has a `TestID` field with a `table:"Test ID"` tag, but `Headers()` and `Rows()` don't include it. Data is fetched but not displayed.

4. **No database-level LIMIT** – All matching rows are fetched from ScyllaDB, then sliced in Python. Pre-existing behavior, but the new CLI endpoint makes it more accessible.

k0machi:
1. Is actually a pre-existing bug. Will fix.
2. Intentional and not a bug.
3. Intentional.
4. Intentional.

soyacz (approved):

soyacz:
@k0machi can you release new cli version?

k0machi:
> @k0machi can you release new cli version?

Will do, I'll be deploying recent changes soon along with Python and Cli versions

PR #1001, opened 2026-05-20
feature(ssh-tunnel): improve sshd config for better connection handling

CodeLieutenant:
Our connections were dropping after 1h of use when testing. After these modifications, SSH tunnel is stable for the whole test duration (unless some network error happens, but we cannot control that).

### DIFF
--- a/scripts/tunnel-server-setup.sh
+++ b/scripts/tunnel-server-setup.sh
@@ -90,21 +90,30 @@ ChallengeResponseAuthentication no
 KbdInteractiveAuthentication no
 PermitRootLogin no
 PubkeyAuthentication yes
+ClientAliveInterval 60
+ClientAliveCountMax 3
+TCPKeepAlive yes
+UsePAM yes
+
+MaxAuthTries 3
+MaxSessions 32
+MaxStartups 50:30:200
+LoginGraceTime 30
+StrictModes yes
 
 Match User ${PROXY_USER}
     AllowTcpForwarding local
+    PermitTunnel yes
+    AllowStreamLocalForwarding yes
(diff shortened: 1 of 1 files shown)


scripts/tunnel-server-setup.sh
```diff
@@ -90,21 +90,30 @@ ChallengeResponseAuthentication no
 KbdInteractiveAuthentication no
 PermitRootLogin no
 PubkeyAuthentication yes
+ClientAliveInterval 60
+ClientAliveCountMax 3
```
k0machi, line 94: I'd set this to 10.

scripts/tunnel-server-setup.sh
```diff
@@ -90,21 +90,30 @@ ChallengeResponseAuthentication no
...
+TCPKeepAlive yes
+UsePAM yes
+
+MaxAuthTries 3
+MaxSessions 32
```
k0machi, line 99: Are you sure? We might have more than 32 clients accessing the proxy. I'd set it to 100

CodeLieutenant: This is not the number of TCP Connection as that is the ulimit size (basically the number of opened files), this is number of sessions per TCP Connection. We can even go lower, like 16, and that would be enough, but I went to larger number to make sure that it works. Putting it to 128 or something like that would just increase the memory usage without giving us any benefits.

scripts/tunnel-server-setup.sh
CodeLieutenant: This is the number where if 3 in a row clients fail to get heartbeat, we should have it lower so that the number of connection that are down, should not clog the server with dead connections.

k0machi (approved):

PR #1002, opened 2026-05-20
fix(service/planner): Diff-based plan updates

k0machi:
This commit addresses a problem where multiple people updating same plan
would override each other's changes. Previously the plan would be
updated by overriding all the fields based on received state. This is
now replaced by a diff-based method, where editor sends updated fields
only and the backend logic combines this into a single plan.

In this case, any removals are resolved by the initial update, while
updates are resolved by latter updates.

### DIFF
--- a/argus/backend/service/planner_service.py
+++ b/argus/backend/service/planner_service.py
@@ -4,7 +4,7 @@ import datetime
 import json
 from collections import defaultdict
 from copy import deepcopy
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from functools import reduce
 from typing import Any, Optional, TypedDict
 from uuid import UUID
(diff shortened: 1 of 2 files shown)


k0machi:
@soyacz ping

soyacz:
from AI, not sure if this is critical, but worth looking at:

### 🔴 Critical — type mismatch breaks removals + dedup

Model stores UUIDs, not strings:
```
tests/groups/participants = List(value_type=columns.UUID)
assignee_mapping          = Map(UUID, UUID)
```
So `plan.tests` returns `list[UUID]`. But diff fields (`tests_remove`, `tests_add`, etc.) arrive from JSON as **strings**.

`planner_service.py:208-216`:
```python
current_tests = list(plan.tests or [])      # [UUID(...), ...]
remove_set = set(plan_request.tests_remove)  # {"abc-...", ...}  strings
current_tests = [t for t in current_tests if t not in remove_set]  # UUID not in str-set → NEVER matches
...
for t in plan_request.tests_add:
    if t not in existing_test_set:           # str not in UUID-set → ALWAYS true → appends dup
```
Consequences:
- **Removals silently no-op.** Removed tests/groups/participants stay in plan.
- **Dedup against existing fails** → concurrent add of same id creates duplicate entry.
- Same bug in `assignee_mapping_remove`/`_set` (`:241-243`) — str keys popped against UUID keys → no-op; mixed str/UUID keys left in map.

Whole feature premise (safe merge) defeated. **Fix:** normalize to UUID on both sides before set ops, e.g. `remove_set = {UUID(x) for x in plan_request.tests_remove}` and convert `_add` ids too. Must run/test against real Scylla — this won't surface in a quick read.

---

### 🟡 Bugs / risk

- **`view_id` precedence soup** `:255`
  ```python
  if plan_request.view_id != str(plan.view_id) if plan.view_id else True:
  ```
  Parses as `(a != b) if plan.view_id else True`. Works by luck, unreadable. Rewrite explicit:
  ```python
  if not plan.view_id or plan_request.view_id != str(plan.view_id):
  ```
- **`completed`/`ends_at` half-wired.** Backend now applies them (`:191-194`) but frontend `computeDiff`/`takeSnapshot` never send them. Dead branch on server, can't be set from UI. Either wire frontend or drop the fields.
- **Snapshot diverges on concurrent edit.** `takeSnapshot()` after save records *frontend* state, not server's merged result. Next diff computed vs stale snapshot → can resurrect items another user removed. Inherent to approach but document it.

---

### 🟡 Diff hygiene

~80% of `+194/-73` is autoformatter noise (black/autopep8) on untouched lines. Ugly multiline f-strings:
```python
f"Found existing plan {
    existing.name} (...)"
```
Pollutes blame, buries the real change. **Split formatting into separate commit** or revert unrelated reformatting.

---

### 🟡 Tests

`changedFiles: 2`, zero tests. New merge logic (remove-then-add, dedup, mapping cleanup, concurrent ordering) is exactly the kind of code that needs unit coverage. Add tests — they'd have caught the UUID/str bug.

soyacz (approved):
LGTM

k0machi:
> from AI, not sure if this is critical, but worth looking at:

Will double check and merge after

PR #1004, opened 2026-05-20
feature(client): JSONL request replay log (client-side, phase 1)

CodeLieutenant:
## Summary

Implements the client-side portion (Phase 1 + Phase 4 plumbing) of the request replay design in [`docs/plans/request_replay.md`](docs/plans/request_replay.md). Every `ArgusClient.post()` is now recorded to an append-only JSONL journal so failed submissions can be replayed once Argus recovers.

- **`argus/client/replay_log.py`** — `ReplayLog` + `ReplayRecord`. Single daemon writer thread fed by a `queue.Queue`; request threads never touch the file. Compact JSON (no spaces, `ensure_ascii=False`), batched flushes (up to 64 records per `flush()`).
- **`argus/client/base.py`** — `ArgusClient.__init__` takes a required `log_dir`, optional `replay_only` flag, and explicit `run_id` kwarg. `post()` wraps the HTTP call in `self._replay_log.record(...)`. `replay_only` skips the HTTP call but still records the intended request and returns a stub 200 response so callers continue without error (no `MagicMock` anti-pattern).
- **Subclasses** (SCT, Generic, Sirenada, DriverMatrix) accept and forward `log_dir` + `replay_only`.
- **In-repo CLIs** (`argus/client/generic/cli.py`, `argus/client/driver_matrix_tests/cli.py`) gain a `--log-dir` option defaulting to cwd.
- **pytest-argus-reporter** gains `--argus-log-dir` (env `ARGUS_LOG_DIR`, default cwd).

Filename: `argus_replay_log_{run_id}_{unix_ns}.jsonl` — nanosecond resolution disambiguates parallel processes and back-to-back instantiation in the same process. `run_id` is sanitized to `[A-Za-z0-9_-]` so it cannot escape `log_dir`.

## Out of scope (follow-ups)

- Phase 2 (backend auto-finalize on terminal status, sirenada `results` dedup)
- Phase 3 (SCT removes its self-healing code; swaps `MagicMock` for `replay_only=True`)
- Phase 5 (`argus replay` CLI command)
- Phase 6 (server-side `POST /api/v1/client/replay/ingest` endpoint)

## Test plan

- [x] `uv run pytest argus/client/tests/` — 59 passed (1 pre-existing unrelated `test_package.py` env error)
- [x] New `argus/client/tests/test_replay_log.py` covers: JSONL format, thread safety (16 threads × 50 writes), filename convention + sanitization, multiple files per run, compact separators, exception capture, `replay_only` mode (POST stubbed, GET raises), `close()` idempotence, context-manager usage, and `ArgusClient.post()` success/HTTP-error/connection-error/2xx-with-`error`-body paths
- [ ] Smoke-test from an SCT job once Phase 3 lands

### DIFF
--- a/argus/client/base.py
+++ b/argus/client/base.py
@@ -1,6 +1,7 @@
 import re
 import logging
 from dataclasses import asdict
+from pathlib import Path
 from typing import Any, Type
 from uuid import UUID
 
@@ -9,6 +10,7 @@ import requests
 from argus.common.enums import TestStatus
 from argus.client.session import create_session
 from argus.client.generic_result import GenericResultTable
+from argus.client.replay_log import ReplayLog, ReplayLogOnlyResponse
 from argus.client.sct.types import LogLink
 
 JSON = dict[str, Any] | list[Any] | int | str | float | bool | Type[None]
@@ -34,24 +36,52 @@ class ArgusClient:
         FETCH_RESULTS = "/testrun/$type/$id/fetch_results"
         FINALIZE = "/testrun/$type/$id/finalize"
 
-    def __init__(self, auth_token: str, base_url: str, api_version="v1", extra_headers: dict | None = None,
-                 timeout: int = 60, max_retries: int = 3, use_tunnel: bool | None = None) -> None:
+    # Subclasses override ``test_type`` as a class attribute; ``run_id`` is
+    # set on the instance by subclass constructors. Both are surfaced in the
+    # replay-log filename.
+    test_type: str | None = None
+
+    def __init__(self, auth_token: str, base_url: str, log_dir: str | Path, api_version="v1",
+                 extra_headers: dict | None = None, timeout: int = 60, max_retries: int = 3,
+                 use_tunnel: bool | None = None, replay_log_only: bool = False,
+                 run_id: UUID | str | None = None) -> None:
         self._auth_token = auth_token
         self._base_url = base_url
         self._api_ver = api_version
         self._timeout = timeout
-        self.session = create_session(
-            auth_token=auth_token,
-            base_url=base_url,
-            use_tunnel=use_tunnel,
-            max_retries=max_retries,
+        self._replay_log_only = replay_log_only
+        # Set run_id on the instance so subclasses that read ``self.run_id``
+        # later see the explicit value, not the class-attribute default.
+        if run_id is not None:
+            self.run_id = run_id
+        # In replay-log-only mode no HTTP calls are made, so skip opening a
+        # session (and any SSH tunnel that might come with it).
+        if replay_log_only:
+            self.session = None
+        else:
+            self.session = create_session(
+                auth_token=auth_token,
+                base_url=base_url,
+                use_tunnel=use_tunnel,
+                max_retries=max_retries,
+            )
+            if extra_headers:
+                self.session.headers.update(extra_headers)
+
+        self._replay_log = ReplayLog(
+            log_dir=log_dir,
+            run_id=str(run_id) if run_id is not None else None,
+            test_type=self.test_type,
         )
 
-        if extra_headers:
-            self.session.headers.update(extra_headers)
+    @property
+    def replay_log_path(self) -> Path:
+        return self._replay_log.path
 
     def close(self) -> None:
-        self.session.close()
+        if self.session is not None:
+            self.session.close()
+        self._replay_log.close()
 
     def __enter__(self) -> "ArgusClient":
         return self
@@ -111,6 +141,12 @@ class ArgusClient:
         }
 
     def get(self, endpoint: str, location_params: dict[str, str] = None, params: dict = None) -> requests.Response:
+        # In replay-log-only mode no HTTP call is made; behave like a mock so
+        # callers (e.g. SCT tests that previously used ``MagicMock``) do not
+        # have to special-case GETs.
+        if self._replay_log_only:
+            LOGGER.debug("GET [replay-log-only] %s params: %s", endpoint, params)
+            return ReplayLogOnlyResponse(endpoint=endpoint)
         url = self.get_url_for_endpoint(
             endpoint=endpoint,
             location_params=location_params
@@ -133,21 +169,28 @@ class ArgusClient:
         params: dict = None,
         body: dict = None,
     ) -> requests.Response:
-        url = self.get_url_for_endpoint(
-            endpoint=endpoint,
-            location_params=location_params
-        )
-        LOGGER.debug("POST Request: %s, params: %s, body: %s", url, params, body)
-        response = self.session.post(
-            url=url,
-            params=params,
-            json=body,
-            headers=self.request_headers,
-            timeout=self._timeout
-        )
-        LOGGER.debug("POST Response: %s %s", response.status_code, response.url)
+        with self._replay_log.record("POST", endpoint, location_params, params, body) as rec:
+            if self._replay_log_only:
+                # Record the request so a future replay can re-send it, but
+                # skip the HTTP call. ``rec`` stays at success=False (default).
+                LOGGER.debug("POST [replay-log-only] %s body: %s", endpoint, body)
+                return ReplayLogOnlyResponse(endpoint=endpoint)
 
-        return response
+            url = self.get_url_for_endpoint(
+                endpoint=endpoint,
+                location_params=location_params
+            )
+            LOGGER.debug("POST Request: %s, params: %s, body: %s", url, params, body)
+            response = self.session.post(
+                url=url,
+                params=params,
+                json=body,
+                headers=self.request_headers,
+                timeout=self._timeout
+            )
+            LOGGER.debug("POST Response: %s %s", response.status_code, response.url)
+            rec.record(response)
+            return response
 
     def submit_run(self, run_type: str, run_body: dict) -> requests.Response:
         return self.post(endpoint=self.Routes.SUBMIT, location_params={"type": run_type}, body={
--- a/argus/client/replay_log.py
+++ b/argus/client/replay_log.py
@@ -1,0 +1,249 @@
+"""JSONL replay log for Argus client API calls.
+
+Every mutating (POST) request is recorded as a single JSONL line so that, when
+Argus is unavailable, the recorded calls can be replayed against the server
+once it recovers. See ``docs/plans/request_replay.md`` for the full design.
+
+Each request thread builds a :class:`ReplayRecord` in memory, runs the HTTP
+call inside the :meth:`ReplayLog.record` context manager, and the record is
+enqueued onto a :class:`queue.Queue` on exit. A single background writer
+thread drains the queue and writes to disk -- request threads never touch the
+file.
+"""
+from __future__ import annotations
+
+import atexit
+import itertools
+import json
+import logging
+import os
+import queue
+import re
+import threading
+import time
+from contextlib import contextmanager
+from dataclasses import asdict, dataclass, field
+from pathlib import Path
+from typing import IO, Any, Iterator
+
+from argus.common.utils import durable_sync
+
+_instance_counter = itertools.count()
+
+LOGGER = logging.getLogger(__name__)
+
+
+class _CloseSentinel:
+    """Singleton signal placed on the queue to stop the writer thread."""
+
+
+_CLOSE = _CloseSentinel()
+
+# Allow only characters that are unambiguously safe inside a filename.
+_UNSAFE_FILENAME_CHARS = re.compile(r"[^A-Za-z0-9_-]")
+
+
+def _sanitize_for_filename(value: str) -> str:
+    # Strip any path-significant characters (slashes, dots) so a hostile run-id
+    # cannot escape ``log_dir``. Dots are dropped entirely rather than mapped
+    # to ``_`` so ``..`` cannot survive the substitution.
+    return _UNSAFE_FILENAME_CHARS.sub("_", value) or "unknown"
+
+
+def _now_ns() -> int:
+    return time.time_ns()
+
+
+@dataclass
+class ReplayRecord:
+    """One Argus API call. Populated during the HTTP call, written on exit."""
+
+    method: str
+    endpoint: str
+    location_params: dict | None
+    params: dict | None
+    body: dict | None
+    test_type: str
+    ts: int = field(default_factory=lambda: _now_ns() // 1_000_000)
+    success: bool = False
+    error: str | None = None
+
+    def record(self, response: Any) -> None:
+        """Populate ``success`` / ``error`` from a ``requests.Response``.
+
+        Mirrors :meth:`ArgusClient.check_response` discriminator: only a
+        2xx with a JSON body and ``status == "ok"`` counts as success.
+        A 2xx HTML page (auth proxy, gateway error) or
+        ``{"status": "error", ...}`` is a failure.
+        """
+        if not 199 < response.status_code < 300:
+            self.error = f"HTTP {response.status_code}"
+            return
+        try:
+            payload = response.json()
+        except ValueError:
+            self.error = f"HTTP {response.status_code} non-JSON response"
+            return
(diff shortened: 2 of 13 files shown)


argus/client/tests/test_replay_log.py
```diff
@@ -0,0 +1,289 @@
...
+    assert records[0]["endpoint"] == ArgusClient.Routes.SUBMIT
+    assert records[0]["location_params"] == {"type": "test-type"}
+    assert records[0]["body"] == {"hello": "world"}
+    assert records[0]["success"] is True
+
+
+def test_argus_client_post_records_failure_on_non_2xx(requests_mock, tmp_path):
+    requests_mock.post(
+        "https://test.example.com/api/v1/client/testrun/test-type/submit",
+        json={"error": "boom"},
```
soyacz, line 188: such a test should not return "error" field - as now we are not sure which mechanism worked

argus/client/tests/test_replay_log.py
soyacz, line 1: add also a test when response is not json - should fail (we always return json, and we usually get pure html on auth issues).
Also, add test when request is not authenticated - we should also mark it as error.

pytest-argus-reporter/pytest_argus_reporter.py
```diff
@@ -183,11 +193,20 @@ def __init__(self, config):
 
     @cached_property
     def argus_client(self):
-        return ArgusGenericClient(
+        kwargs = dict(
```
soyacz: I'm not sure if we need that. We rather first upgrade argus client then pytest reporter. @k0machi WDYT?

pytest-argus-reporter/pytest_argus_reporter.py
```diff
@@ -164,6 +173,7 @@ def __init__(self, config):
         self.base_url = config.getoption("base_url")
         self.api_key = config.getoption("api_key")
         self.extra_headers = config.getoption("extra_headers")
+        self.log_dir = config.getoption("log_dir")
```
soyacz: we miss here `replay_only` which actually might be useful in pytest reporter - where we could collect artifacts from these tests and send at the end only in bulk.

soyacz (commented):
sevearl from me, and a few from AI:
## Review: PR #1004 — JSONL request replay log (client-side, phase 1)

**3 findings**, 1 medium, 1 low, 1 informational.

### 1. Bug: `success` detection checks wrong key — `base.py:212`

```python
success = "error" not in response.json()
```

The Argus API returns `{"status": "error", "response": {...}}` — the discriminator key is `"status"`, not `"error"`. This means server-side errors returned as HTTP 200 (which happens — see `error_handlers.py:49`) will be logged as `success: true`. The correct check:

```python
success = response.json().get("status") == "ok"
```

This aligns with how `check_response` (line 133) already validates responses.

**Severity: Medium.** Replay records will have incorrect success flags, which matters when later phases use the log to decide which requests to retry.

### 2. `replay_only` mode still creates an HTTP session — `base.py:79-84`

In `replay_only` mode, `create_session()` is called even though no HTTP requests will ever be made (GET raises, POST returns a stub). Wasteful but not broken.

**Severity: Low.**

### 3. `_ReplayOnlyResponse` lacks `request` attribute compatibility — `base.py:35`

`_ReplayOnlyResponse.request = None` works today because the stub's `status_code=200` and `json()` returning `{"status": "ok"}` means code paths accessing `response.request.method` are never reached. But it's a fragility point if callers change.

**Severity: Informational.**

argus/client/tests/test_replay_log.py
CodeLieutenant: Right — fixed in e14c756. Replaced the test with a real-shape one (`{"status": "error", "response": {...}}`) and added two more: non-JSON 200 (auth proxy HTML) and 401.

argus/client/tests/test_replay_log.py
CodeLieutenant: Added in e14c756: `test_argus_client_post_records_non_json_response_as_failure` covers the auth-proxy HTML case, and `test_argus_client_post_records_401_unauthenticated_as_failure` covers unauthenticated. Both end up with `success=False` and a descriptive `error` string.

pytest-argus-reporter/pytest_argus_reporter.py
CodeLieutenant: Agreed — reverted all pytest-argus-reporter changes from this PR in e14c756. Will land them in a follow-up PR after argus-alm is released so we don't need the `inspect.signature` shim.

pytest-argus-reporter/pytest_argus_reporter.py
CodeLieutenant: Good idea — will wire up `--argus-replay-only` and bulk upload-at-end in the reporter follow-up PR.

CodeLieutenant:
Tested with

https://github.com/scylladb/scylla-cluster-tests/pull/14778

🟢 https://jenkins.scylladb.com/job/scylla-staging/job/dusan/job/longevity-10gb-3h-gce-test/4/ (Email Sending Failed) - There is no argus run since Argus is disabled for this run to test when argus is down what would happen. Log is stored and with SCT PR there is command `download-logs`

```bash
./sct.py investigate download-logs 15bb6cad-1d24-4e81-800e-9c45d6ea7d06 --log-type event
```

argus/client/replay_log.py
```diff
@@ -0,0 +1,218 @@
...
+    # to ``_`` so ``..`` cannot survive the substitution.
+    return _UNSAFE_FILENAME_CHARS.sub("_", value) or "unknown"
+
+
+def _now_ns() -> int:
+    return time.time_ns()
+
+
+class ReplayRecord:
+    """One Argus API call. Populated during the HTTP call, written on exit."""
```
k0machi, line 59: Why not just a dataclass?

argus/client/replay_log.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+_CLOSE = _CloseSentinel()
+
+# Drain up to this many records from the queue per flush to amortize the
+# per-flush cost across bursts of requests.
+_BATCH_SIZE = 64
+# Compact JSON separators -- no spaces between tokens. Roughly 10-15% smaller
+# than the json.dumps default for the kind of structured bodies we record.
+_JSON_SEPARATORS = (",", ":")
```
k0machi: Don't see a point in making it constant since it is used in precisely one spot and is something obvious.

k0machi (approved):
Minor comments,  otherwise LGTM.

argus/client/tests/test_replay_log.py
```diff
@@ -0,0 +1,341 @@
...
+        body={"k": "v"},
+    )
+    assert response.status_code == 200
+    assert response.json() == {"status": "ok", "response": {}}
+    client.close()
+
+    records = _read_records(client.replay_log_path)
+    assert len(records) == 1
+    assert records[0]["success"] is False
+    assert "replay-only" in records[0]["error"]
```
soyacz: `replay-only` -> `replay-log-only`

argus/client/base.py
```diff
@@ -34,24 +58,52 @@ class Routes():
...
         self._auth_token = auth_token
         self._base_url = base_url
         self._api_ver = api_version
         self._timeout = timeout
-        self.session = create_session(
-            auth_token=auth_token,
-            base_url=base_url,
-            use_tunnel=use_tunnel,
-            max_retries=max_retries,
+        self._replay_only = replay_only
```
soyacz: `_replay_only`->`_replay_log_only`

argus/client/base.py
```diff
@@ -111,6 +163,11 @@ def request_headers(self):
         }
 
     def get(self, endpoint: str, location_params: dict[str, str] = None, params: dict = None) -> requests.Response:
+        if self._replay_only:
+            raise ArgusClientError(
```
soyacz: if we're going to use replay_only for Mocking, we should not raise (check what current Mock in SCT does)

argus/client/base.py
```diff
@@ -133,21 +190,51 @@ def post(
...
-            url=url,
-            params=params,
-            json=body,
-            headers=self.request_headers,
-            timeout=self._timeout
-        )
-        LOGGER.debug("POST Response: %s %s", response.status_code, response.url)
-
-        return response
+        with self._replay_log.record("POST", endpoint, location_params, params, body) as rec:
```
soyacz, line 172: I think this could be neater:
`rec` could get `record` method with accepting response as input - based on that do whole logic.
Also, if we use `_replay_only` it should actually skip running this body and just return mocked response.
This way, we wouldn't have to modify anything here to work with replay logger

soyacz:
@CodeLieutenant see Alexey and mine reviews & address

argus/client/base.py
```diff
@@ -133,21 +191,26 @@ def post(
...
-            timeout=self._timeout
-        )
-        LOGGER.debug("POST Response: %s %s", response.status_code, response.url)
-
-        return response
+        if self._replay_log_only:
+            LOGGER.debug("POST [replay-log-only] %s body: %s", endpoint, body)
+            return _ReplayLogOnlyResponse(endpoint=endpoint)
+
+        with self._replay_log.record("POST", endpoint, location_params, params, body) as rec:
```
soyacz, line 172: still I'd prefer it to be bundled inside `ReplayLog`. Upon init of this class we could say if it's replay only and skip calling inside of the context manager scope. Could be achieved by raising like this:
```
class SkipBlock(Exception):
    """Custom exception used to break out of the context body."""
    pass

class ConditionContext:
    def __init__(self, should_execute):
        self.should_execute = should_execute

    def __enter__(self):
        if not self.should_execute:
            print("Condition not met. Skipping the body!")
            raise SkipBlock()
        print("Condition met. Entering the body...")
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        # If our custom SkipBlock exception was raised, suppress it
        if exc_type is SkipBlock:
            return True  # Returning True swallows the exception
        
        # Let any other unexpected exceptions bubble up
        return False
```
(btw, we don't need then 'record' to be context manager, we could use instance of ReplayLog with `with` stmt
This way, whole feature would be skoped just to one class/ctx manager and we wouldn't leak this feature that much into 'base.py' (also dropping `_ReplayLogOnlyResponse` from here)

CodeLieutenant: I dont like this idea as it is using exceptions to control flow. Really hard to read and understand whats going on. Leaking it, I think is not a problem, making it fully encapsulated will

soyacz: ok, fair point. We could then move `_ReplayLogOnlyResponse` to `replay_log.py`.

soyacz (approved):
LGTM

k0machi (approved):
Generally looks okay, please take a look at SCT-377 to see if there are any arising issues from this.

soyacz:
**Findings:**

- **`close()`/`record()` TOCTOU (low).** `replay_log.py:199` checks `_closed` then enqueues; `close()` can set flag + push `_CLOSE` between check and `put()`. Writer exits on `_CLOSE`, late record sits in queue unwritten — lost silently. Rare (close races in-flight POST), tolerable per design (atexit drain), but real. If you care, drain remaining queue after join, or guard with lock.

- **`atexit` handler leak (low-med).** `replay_log.py:124` every `ReplayLog` does `atexit.register(self.close)` with no unregister. atexit holds strong ref for process life → instances never GC'd. Fine for CLI (one client, exits). Bad for long-lived process spawning many clients. Fix: `atexit.unregister(self.close)` in `close()`.

- **`_now_ns()` filename uniqueness (low).** `replay_log.py:116` two instances same process could collide if `time_ns()` returns equal value (coarse clock). `test_..._multiple_instances` asserts paths differ → theoretically flaky. Practically rare. Add counter/pid suffix to be safe.

- **Dead helper / inconsistency (nit).** `_now_ns()` defined and used for filename, but `ts` field (`replay_log.py:61`) inlines `time.time_ns() // 1_000_000`. Two ways to read clock. Cosmetic.

- **"fsync" comment misleading (nit).** `replay_log.py:152` says amortize fsync, but code only `f.flush()` (Python buf → OS, not disk). Reword to "flush" — durability claim in class docstring (`SIGKILL` loses queue) is otherwise honest.

## base.py

- **Breaking API change (med — call out).** `ArgusClient.__init__` `log_dir` is **required, no default** (`base.py:62`). Every external caller (SCT repo etc.) breaks until updated. In-repo CLIs default `"."`, but the constructor itself doesn't. Intentional per Phase 3 plan, but confirm downstream rollout is coordinated. Consider default `"."` to soften, or keep required if you want explicitness — just flag it loudly in changelog.

- **`_ReplayLogOnlyResponse.request = None` (low).** Stub passes `check_response` only on happy path (200 + status ok). If any caller hits the error branch with the stub, `check_response` (`base.py:141`) dereferences `response.request.method` → AttributeError. Stub never errors today, so latent. Fine for now.

## Tests — valid? Yes.

**Gaps (minor):**
- No test for close/record race (the TOCTOU above) — hard to test deterministically, acceptable.
- No test that `params` round-trips non-None (always `None` in cases). Trivial.
- Thread-safety test relies on writer draining before `close()` returns — passes because `close()` joins writer. Good, but timing-dependent by nature; current join makes it deterministic.

## Verdict

Approve with nits. Design sound, tests real and passing. Before merge address: (1) required-`log_dir` downstream coordination, (2) atexit unregister in `close()`. Rest are low-priority polish.

soyacz:
@CodeLieutenant let me know if you fix the above. Then we'll merge

CodeLieutenant:
@soyacz About this

> Breaking API change (med — call out). ArgusClient.__init__ log_dir is required, no default (base.py:62). Every external caller (SCT repo etc.) breaks until updated. In-repo CLIs default ".", but the constructor itself doesn't. Intentional per Phase 3 plan, but confirm downstream rollout is coordinated. Consider default "." to soften, or keep required if you want explicitness — just flag it loudly in changelog.

Should we make it not breaking or breaking it, is fine?

CodeLieutenant:
> Generally looks okay, please take a look at SCT-377 to see if there are any arising issues from this.

This is how we actually discovered this bug, first with SSH Client we discovered that there were 2, but we thought one was not that long lasting, but this replay log showed that both are long lived. This PR introduces nothing, same as SSH one, that affects the number of clients, bug is in SCT side

k0machi:
> > Generally looks okay, please take a look at SCT-377 to see if there are any arising issues from this.
> 
> This is how we actually discovered this bug, first with SSH Client we discovered that there were 2, but we thought one was not that long lasting, but this replay log showed that both are long lived. This PR introduces nothing, same as SSH one, that affects the number of clients, bug is in SCT side

No, I left a comment there talking about how replay log displays two lines of requests, one with success: true and other with `null`, is that intended? As I mentioned on the ticket it's not exactly a "bug", but we could centralize client, even though that's not really needed (unless you want a single replay log)

CodeLieutenant:
> > > Generally looks okay, please take a look at SCT-377 to see if there are any arising issues from this.
> > 
> > 
> > This is how we actually discovered this bug, first with SSH Client we discovered that there were 2, but we thought one was not that long lasting, but this replay log showed that both are long lived. This PR introduces nothing, same as SSH one, that affects the number of clients, bug is in SCT side
> 
> No, I left a comment there talking about how replay log displays two lines of requests, one with success: true and other with `null`, is that intended? As I mentioned on the ticket it's not exactly a "bug", but we could centralize client, even though that's not really needed (unless you want a single replay log)

Ohhh two log lines, well that I need to investigate

CodeLieutenant:
@k0machi I've found a bug that you described, it's not inside argus client, but on SCT side: `replay_log.py` and `base.py` with a "two-phase write" 

Pre-flight record with `success: null`, flushed synchronously before the HTTP call (so intent survives a kill mid-call)
Outcome record with success: true/false, written after

Will fix it once we have this one released, we can safely merge this

k0machi (approved):

PR #1005, opened 2026-05-21
fix(sct/service): Provide owner and repo fields for similar event issues

k0machi:
This commit fixes an issue where IssueBadge.svelte would not receive
owner and repo fields for github issues from SctSimilarEvents table.

Fixes ARGUS-143


soyacz (approved):
LGTM

PR #1007, opened 2026-05-22
feature(backend): replay-log ingest endpoint

CodeLieutenant:
## Summary

End-to-end implementation of the JSONL replay-log system from [`docs/plans/request_replay.md`](docs/plans/request_replay.md). Argus can now ingest replay logs captured by the Python client (#1004), and a new `argus run replay` CLI command bundles and uploads them. Verified end-to-end against a real SCT bundle: 418 records → 414 succeeded, 4 skipped, 0 failed.

This PR is based off master and is independent of #1004 — landing it alone gives Argus the ingest capability plus the CLI to drive it.

### Backend — `POST /api/v1/client/replay/ingest`

Accepts a `tar.zst` archive (also `application/zstd` / `application/octet-stream`), streams it through `zstandard` → `tarfile`, merges JSONL across every archive member, and re-applies each request **as an in-process proxy through `current_app.test_client()`**. The recorded request runs against the same controller, decorators (auth, error handling) and idempotency logic as a live client call. The Authorization header on the ingest request is forwarded so internal proxied calls inherit the original identity.

Server-side cross-cutting:

- **Ordering** — `submit_run` runs first (the run must exist before anything else touches it), terminal `set_status` runs last (back-fills `end_time`), heartbeats collapse to only the most recent record per `(type, id)`.
- **Skip list** — `/testrun/report/email` and `/planning/plan/trigger` would have external side effects; legacy `/sct/$id/events/submit` is superseded by the individual `/sct/$id/event/submit` records already in the same log; `finalize` is folded into terminal `set_status` per the plan and treated as a no-op.
- **Repeated-slash normalisation** — back-compat for an old SCT client typo (`/sct/$id//stress_cmd/submit`); the typo is also fixed in `argus/client/sct/client.py`.

Query params:

- `?dry_run=true` validates and counts without executing.
- `?create_missing_tests=true` opts into a pre-step that parses the run's `build_id` (`release/group/test`, mirroring the Jenkins folder layout used by `build_system_monitor`) and creates the matching `ArgusRelease` / `ArgusGroup` / `ArgusTest` rows when they don't already exist. When a run row already exists with empty categorical fields (e.g. from a partial replay before the flag was set), the helper also back-fills `test_id` / `release_id` / `group_id` so subsequent `submit_results` calls (partitioned by `test_id`) succeed. Curated instances leave the flag off and behave unchanged.

### CLI — `argus run replay`

Top-level command (under `argus run`) that bundles JSONL replay-log files into a tar.zst archive and streams them to the ingest endpoint.

Inputs are normalised in the CLI so the backend always receives the same shape:

- `--file <path>` accepts plain `.jsonl`, `.jsonl.zst[d]`, and `.tar.zst[d]` archives. tar.zst inputs are extracted and filtered to entries whose basename starts with `argus_replay_log_` so SCT event bundles can be passed directly.
- `--dir <path>` scans for `argus_replay_log_*.jsonl[.zst]` files, optionally filtered by `--run-id` (the filter also applies inside tar.zst archives).

Other flags:

- `--dry-run` — ask the server to validate without executing.
- `--target-url` — replay against a different Argus instance using the current keychain credentials.
- `--report text|json` — summary format; the text report renders the per-record errors table when non-empty.
- `--create-missing-tests` — pass the corresponding query param to the backend.

JSONL lines are pre-validated before packing (a single malformed line aborts with `filename:lineno: <parse error>`). The archive is streamed through an `io.Pipe` so memory usage stays flat regardless of input size.

## Architecture notes

The initial backend draft (still visible in the first commit, [`12b9fc8`](https://github.com/scylladb/argus/commit/12b9fc8)) used a per-endpoint dispatcher table mapping each route template to a `_dispatch_*` helper that called the corresponding internal service function directly. It was rewritten to the proxy approach (in [`855b954`](https://github.com/scylladb/argus/commit/855b954)) for two reasons:

1. The dispatcher table drifted from the controllers — `/sct/$id/resource/$name/terminate` had a controller but no dispatcher; `/sct/$id/events/submit` had no entry; the SCT client double-slash typo would have needed dispatch-side workarounds.
2. Every dispatcher reimplemented the controller's call signature, which is brittle and offers no upside.

The proxy approach delegates all of that to the real controllers via `app.test_client()` — no actual HTTP, no separate route↔service-method mapping to keep in sync, no replay-specific reimplementations.

## Out of scope (still ahead)

- **Phase 2** — `set_status` terminal auto-finalize + sirenada `results` dedup. Will land as a small follow-up.
- **Phase 3** — SCT self-healing removal (different repo).

## Test plan

- [x] `uv run pytest argus/backend/tests/replay/` — 36 passed in 0.04s (proxy dispatch, ordering, normalisation, skip-list, response handling, `create_missing_tests` pre-step, build-id parser)
- [x] `go test ./...` in `cli/` — all green
- [x] End-to-end against a real SCT bundle (`sct-runner-events-15bb6cad.tar.zst`):
  - `--file <tar.zst>` (mixed bundle, 15 entries) → 418 records, 414 succeeded, 4 skipped, 0 failed
  - `--dir <results> --run-id <uuid>` → same numbers
  - `--file <single .jsonl>` → 4 / 4 / 0 / 0
  - `--file <single .jsonl.zst>` → 4 / 4 / 0 / 0

Refs: https://scylladb.atlassian.net/browse/QAINFRA-43


argus/backend/controller/replay_api.py
```diff
@@ -0,0 +1,75 @@
...
+    try:
+        summary = ReplayService().ingest(archive, dry_run=dry_run)
+    except ReplayServiceError as exc:
+        return {
+            "status": "error",
+            "response": {
+                "exception": type(exc).__name__,
+                "arguments": [str(exc)],
+            },
+        }, 400
```
k0machi: Forward exceptions instead of returning error yourself

argus/backend/controller/replay_api.py
```diff
@@ -0,0 +1,75 @@
...
+
+    archive = request.get_data(cache=False)
+    if not archive:
+        return {
+            "status": "error",
+            "response": {
+                "exception": "EmptyRequest",
+                "arguments": ["request body is empty"],
+            },
+        }, 400
```
k0machi: Ditto

argus/backend/controller/replay_api.py
```diff
@@ -0,0 +1,75 @@
...
+    if content_type and content_type not in ACCEPTED_CONTENT_TYPES:
+        return {
+            "status": "error",
+            "response": {
+                "exception": "UnsupportedMediaType",
+                "arguments": [
+                    f"expected one of {sorted(ACCEPTED_CONTENT_TYPES)}, got {content_type!r}"
+                ],
+            },
+        }, 415
```
k0machi: Ditto

argus/backend/service/replay_service.py
```diff
@@ -0,0 +1,418 @@
...
+Each endpoint template recorded by the client is mapped to a dispatcher in
+:data:`ENDPOINT_DISPATCH` along with its idempotency class:
+
+* :class:`Idempotency.SAFE` -- replaying is harmless (last-write-wins or
+  natural dedup by key).
+* :class:`Idempotency.APPEND` -- the backend already dedupes appends on
+  payload match.
+* :class:`Idempotency.SKIP` -- has side effects outside Argus (email,
+  triggering CI), never replayed.
+"""
```
k0machi: I don't like that this service explicitly imports plugin services. I propose extending plugininfo class with extra information with those dispatchable endpoints and then providing `all_plugin_replayable_endpoints` function to collect them

argus/backend/service/replay_service.py
```diff
@@ -0,0 +1,418 @@
...
+from __future__ import annotations
```
k0machi: Not needed.

argus/backend/service/replay_service.py
```diff
@@ -0,0 +1,418 @@
...
+        results=body,
+    )
+
+
+def _dispatch_submit_config(location_params: dict, body: dict) -> None:
+    ClientService.submit_config(
+        run_id=location_params["id"],
+        config_name=body["name"],
+        config_content=body["content"],
+    )
```
k0machi: This can stay however, as this is part of core services in argus/backend/service

argus/backend/service/replay_service.py
```diff
@@ -0,0 +1,418 @@
...
+# Each entry maps the route template recorded by the client to the internal
+# service callable that re-applies it. Dispatcher signature:
+#
+#     dispatcher(location_params: dict, body: dict) -> None
+#
+# Dispatchers should raise on logical failure; the runner catches and records
+# the error in the summary.
+
+def _dispatch_submit_run(location_params: dict, body: dict) -> None:
+    ClientService().submit_run(run_type=location_params["type"], request_data=body)
```
k0machi: I'd consider trying to call the endpoint instead with custom context to make it truly 1:1 (and prevent deviation in case endpoint gets updated)

k0machi (changes requested):

argus/backend/service/replay_service.py
CodeLieutenant: Yes that was my initial idea, but how to do it without authentication? And do I have to bypass something else

argus/backend/controller/replay_api.py
```diff
@@ -0,0 +1,77 @@
...
+LOGGER = logging.getLogger(__name__)
+
+bp = Blueprint("replay_api", __name__, url_prefix="/replay")
+bp.register_error_handler(Exception, handle_api_exception)
+
+
+ACCEPTED_CONTENT_TYPES = frozenset({
+    "application/x-tar-zstd",
+    "application/zstd",
+    "application/octet-stream",
```
soyacz, line 44: tar.gz should be supported too

argus/backend/controller/replay_api.py
```diff
@@ -0,0 +1,77 @@
...
+    auth_header = request.headers.get("Authorization")
+
+    # ``ReplayServiceError`` inherits from ``APIException`` and is handled
+    # by the blueprint's error handler registered above.
+    summary = ReplayService(
+        auth_header=auth_header,
+        create_missing_tests=create_missing_tests,
+    ).ingest(archive, dry_run=dry_run)
+
+    LOGGER.info(
```
soyacz: this log msg should go to ReplayService

argus/backend/service/replay_service.py
```diff
@@ -0,0 +1,452 @@
...
+# Endpoint templates whose dispatch we never want to perform during replay.
+# These either touch systems outside Argus (email, CI) or are superseded
+# by other endpoints in the same log (the legacy bulk events submit).
+SKIP_ENDPOINTS: frozenset[str] = frozenset({
+    "/testrun/report/email",       # would send actual mail
+    "/planning/plan/trigger",      # would re-trigger CI jobs
+    # Legacy bulk submit -- the individual ``/sct/$id/event/submit`` records
+    # in the same log already carry the same data and replaying both would
+    # duplicate rows.
+    "/sct/$id/events/submit",
```
soyacz: can be removed - it's no-op anyway

soyacz (commented):
partial review

argus/backend/controller/replay_api.py
CodeLieutenant: Done

cli/cmd/replay.go
```diff
@@ -0,0 +1,343 @@
...
+	createMissingTests bool,
+	log zerolog.Logger,
+) (*models.ReplayIngestSummary, error) {
+	route := api.ReplayIngest
+	q := url.Values{}
+	if dryRun {
+		q.Set("dry_run", "true")
+	}
+	if createMissingTests {
+		q.Set("create_missing_tests", "true")
```
soyacz, line 241: I think default should be 'false' - so user knows that tries to upload non-existent test. 
When fails to find the test (when false) it should tell what release/group/test it will try to create and what is missing (release? group? test?)

CodeLieutenant: It is by default false, and it only creates the test, not the group or release.

argus/backend/service/replay_service.py
```diff
@@ -0,0 +1,624 @@
...
+
+        try:
+            response = client.open(
+                url,
+                method=method,
+                query_string=params or None,
+                json=body if body is not None else None,
+                headers=headers,
+            )
+        except Exception as exc:  # noqa: BLE001 -- isolate per-record
```
soyacz, line 561: can this happen? don't we catch all the failures? 
If we bypass error handler here, we should not `return`

soyacz (commented):
1. this feature deserves a documentation.
2. how long ingest takes? won't it timeout the request and abruptly stop ingestion

argus/backend/service/replay_service.py
CodeLieutenant: > can this happen? don't we catch all the failures? If we bypass error handler here, we should not `return`

If I dont catch it per record, it will fail with everything, now the question is, do we want partials or only full?

CodeLieutenant:
> 1. this feature deserves a documentation.

The whole replay, or just backend, or cli, or all three?

> 2. how long ingest takes? won't it timeout the request and abruptly stop ingestion

with my run I pasted here, it took around 300ms, didn't measure it the proper way, but <1s, probably won't timeout

soyacz:
> > 1. this feature deserves a documentation.
> 
> The whole replay, or just backend, or cli, or all three?

Best would be all three - how clients must be configured to have these logs, and later how to apply these logs to argus (including options we have for cli), what happens if there's no release/group/test etc.

> > 2. how long ingest takes? won't it timeout the request and abruptly stop ingestion
> 
> with my run I pasted here, it took around 300ms, didn't measure it the proper way, but <1s, probably won't timeout

Most of the times it won't be long, but for heavy tests, with many error events it may take longer than 1s. @k0machi will this continue after request times out?
We can tackle this issue in the followup if it is a problem (maybe after migration to FastAPI).

argus/backend/service/replay_service.py
soyacz: ah, my mistake, I misunderstood the flow here. 
We want partials too. Sometimes we might just fail one request.
One more thing to check for you: when we post generic results table, what is returned when it finds validation error (value fall outside of limits)? IIRC it may return error, but I'm not sure.

CodeLieutenant:
> > > 1. this feature deserves a documentation.
> > 
> > 
> > The whole replay, or just backend, or cli, or all three?
> 
> Best would be all three - how clients must be configured to have these logs, and later how to apply these logs to argus (including options we have for cli), what happens if there's no release/group/test etc.
> 
> > > 2. how long ingest takes? won't it timeout the request and abruptly stop ingestion
> > 
> > 
> > with my run I pasted here, it took around 300ms, didn't measure it the proper way, but <1s, probably won't timeout
> 
> Most of the times it won't be long, but for heavy tests, with many error events it may take longer than 1s. @k0machi will this continue after request times out? We can tackle this issue in the followup if it is a problem (maybe after migration to FastAPI).

We can mitigate this with background processing, upload a file, parse all json lines, and insert them to a table, later pull and do it one by one.

CodeLieutenant:
@soyacz @k0machi ping

argus/backend/service/replay_service.py
CodeLieutenant: for validation errors, there is exception thrown IIRC in all endpoints, and we are keeping it that way so that we have one generic exception handler

soyacz (approved):
LGTM. Clean end-to-end implementation. The proxy-via-test_client design is the right call — no controller logic reimplemented, real auth/idempotency/decorators run as for a live request, and the dispatcher-table rewrite is well justified.

Verified:
- Ordering invariants (submit_run first, terminal set_status last, heartbeat collapse) match the plan.
- Magic-byte archive detection (not Content-Type) + skip-list (email/CI-trigger/finalize) are correct.
- create_missing_tests pre-step / default-path diagnosis cleanly avoids silently inserting runs with empty test_id.
- CLI streams via io.Pipe (flat memory), pre-validates JSONL, normalizes all input shapes to tar.zst.
- Solid test coverage (36 backend + Go pack/extract/controller/hierarchy tests).

Phase 2 (set_status auto-finalize) tracked as follow-up. 👍

soyacz:
@CodeLieutenant fix conflicts
@k0machi please take a look if we're good to go

CodeLieutenant:
@soyacz @k0machi Conflicts resolved, PR ready

PR #1011, opened 2026-05-27
feat(api): add email to public user JSON response

soyacz:
## Summary

- Adds the `email` field to `User.to_json()` so the public `/api/v1/users` endpoint includes user emails without requiring admin access.
- Previously, email was only available through the admin endpoint (`/admin/api/v1/users`), forcing consumers to request elevated privileges for basic contact info.


fruch (approved):
LGTM

k0machi (approved):

PR #1012, opened 2026-05-28
feat(jenkins): parse test_metadata from job config XML (primary) and build description (fallback)

fruch:
## Summary

- Adds `parse_test_metadata_from_config_xml()` utility that extracts SCT pipeline metadata (tier, test_type, duration_class, supported_backends) from the `<testMetadata>` XML element in Jenkins job config.xml
- Adds `parse_test_metadata_from_description()` as a legacy fallback that parses from Jenkins build descriptions (HTML and markdown formats)
- Adds `get_job_test_metadata()` method that fetches job config.xml and extracts metadata from the XML element
- Updates `get_build_test_metadata()` to try config.xml first, then fall back to description parsing
- Includes unit tests for both parsers with real-world examples

## Context

SCT PR [scylladb/scylla-cluster-tests#14818](https://github.com/scylladb/scylla-cluster-tests/pull/14818) adds `test_metadata` to test-case YAML files. At job creation time, `create_test_release_jobs.py` writes this metadata as a structured `<testMetadata>` XML element in the Jenkins job config.xml:

```xml
<testMetadata>
  <tier>tier1</tier>
  <test_type>longevity</test_type>
  <duration_class>short</duration_class>
  <supported_backends>
    <backend>aws</backend>
    <backend>gce</backend>
  </supported_backends>
</testMetadata>
```

This XML approach (agreed upon in PR review discussion with @soyacz) is preferred over the build description approach because:
1. It is set at job creation time and is immutable (not affected by build-time processes)
2. It is structured XML, not fragile string parsing
3. It cannot be accidentally modified by other Jenkins processes

The build description fallback handles legacy jobs that predate the XML element.

## Priority of metadata sources

1. **Job config.xml `<testMetadata>` element** — primary, immutable, set at job creation
2. **Build description parsing** — legacy fallback for older jobs

Note: The argus client in SCT does NOT send test_metadata at runtime via the API. Argus retrieves it independently by reading the Jenkins job config.

### DIFF
--- a/argus/backend/service/jenkins_service.py
+++ b/argus/backend/service/jenkins_service.py
@@ -13,6 +13,35 @@ from argus.backend.models.web import ArgusGroup, ArgusRelease, ArgusTest, UserOa
 LOGGER = logging.getLogger(__name__)
 GITHUB_REPO_RE = r"(?P<http>^https?:\/\/(www\.)?github\.com\/(?P<user>[\w\d\-]+)\/(?P<repo>[\w\d\-]+)(\.git)?$)|(?P<ssh>git@github\.com:(?P<ssh_user>[\w\d\-]+)\/(?P<ssh_repo>[\w\d\-]+)(\.git)?)"
 
+_METADATA_RE = re.compile(
+    r"<b>Metadata:</b>\s*"
+    r"tier=(?P<tier>[^\s|]+)\s*\|\s*"
+    r"type=(?P<test_type>[^\s|]+)\s*\|\s*"
+    r"duration=(?P<duration_class>[^\s|]+)\s*\|\s*"
+    r"backends=(?P<supported_backends>[^<]+)",
+)
+
+
+def parse_test_metadata_from_description(description: str | None) -> dict[str, Any] | None:
+    """Parse test_metadata fields from a Jenkins build description HTML.
+
+    Returns a dict with keys (tier, test_type, duration_class, supported_backends)
+    or None if the metadata block is not present.
+    """
+    if not description:
+        return None
+    match = _METADATA_RE.search(description)
+    if not match:
+        return None
+    backends_raw = match.group("supported_backends").strip()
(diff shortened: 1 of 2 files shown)


fruch:
This is the suggested "interface" for the test metadata, it's gonna be available on the job description

this PR is just ref to see where it should handled, it's not saving or doing anything with the metadata yet

argus/backend/service/jenkins_service.py
```diff
@@ -13,6 +13,35 @@
 LOGGER = logging.getLogger(__name__)
 GITHUB_REPO_RE = r"(?P<http>^https?:\/\/(www\.)?github\.com\/(?P<user>[\w\d\-]+)\/(?P<repo>[\w\d\-]+)(\.git)?$)|(?P<ssh>git@github\.com:(?P<ssh_user>[\w\d\-]+)\/(?P<ssh_repo>[\w\d\-]+)(\.git)?)"
 
+_METADATA_RE = re.compile(
+    r"<b>Metadata:</b>\s*"
+    r"tier=(?P<tier>[^\s|]+)\s*\|\s*"
+    r"type=(?P<test_type>[^\s|]+)\s*\|\s*"
```
soyacz, line 19: looks very fragile.
Why not using yaml/json/xml?

fruch: this PR is reference,

see
`parse_test_metadata_from_config_xml` below for code path that doesn't use regex and read it from the XML of the job metadata directly

soyacz: but we don't use `parse_test_metadata_from_config_xml` anywhere - only in tests.
And `parse_test_metadata_from_description` seems to be obsolete now

fruch: again, this PR is demo of how to retrieve them, it has both options shown.

where it should be used and how it would store it isn't part of what's here

k0machi:
Closing in favour of #1096

PR #1013, opened 2026-06-01
docs(plans): Add CLI release planner command plan

k0machi:
Add a full 7-section implementation plan for a new `argus planner`
subcommand in the Go CLI, covering list, get, create, update, delete and
copy operations plus a search command that exposes the backend test/group
lookup for UUID discovery.

The plan is phased PR-by-PR (types, read commands, discovery, then the
write paths), follows the cmd/discussions sub-package + services pattern,
and resolves release/user names client-side. The update phase targets the
upcoming diff-based PlanDiffPayload contract from
backend/release-planner-multi-edit-safety, and the test/group input
ergonomics are intentionally left open as Appendix proposals. Registered
under Client in MASTER.md and progress.json.

### DIFF
--- a/docs/plans/cli-planner-command.md
+++ b/docs/plans/cli-planner-command.md
@@ -61,12 +61,13 @@ The `ArgusReleasePlan` model (`argus/backend/models/plan.py:8`) has fields: `id`
   substring-replacing the source release name with the target release name; entities
   with no match are dropped unless present in `payload.replacements`.
 
-#### Upcoming update mechanism — diff-based (ref `backend/release-planner-multi-edit-safety`)
-The branch `backend/release-planner-multi-edit-safety` (commit
-`fix(service/planner): Diff-based plan updates`) replaces the full-object
+#### Update mechanism — diff-based (merged to master, `planner_service.py:62`)
+The diff-based plan update (commit `e1fa19db`, `fix(service/planner): Diff-based plan
+updates`) is **merged to master and present on this branch**. It replaces the full-object
 `TempPlanPayload` update with a **diff payload** for multi-editor safety
-(last-edit-wins per field, remove-wins for list concurrency). The new
-`PlanDiffPayload` shape is:
+(last-edit-wins per field, remove-wins for list concurrency). The `PlanDiffPayload`
+shape (`planner_service.py:62`, consumed by `update_plan` at `planner_service.py:178`)
+is:
 
 - `id` (required).
 - Scalar fields, **only sent if changed**: `name`, `description`, `owner`,
@@ -78,9 +79,9 @@ The branch `backend/release-planner-multi-edit-safety` (commit
   (list of entity ids).
 
 The server applies removes before adds, dedupes, and prunes `assignee_mapping`
-entries whose entity is no longer in `tests`+`groups`. **This plan targets the
-diff-based contract for `planner update` (Phase 5)** so the CLI lands aligned with the
-new backend rather than the legacy full-object replacement.
+entries whose entity is no longer in `tests`+`groups`. **`planner update` (Phase 5)
+targets this diff-based contract** so the CLI lands aligned with the backend rather than
+the legacy full-object replacement.
 
 ### Name-resolution endpoints (verified)
 - `GET /api/v1/releases` → enabled releases with `id`/`name` (`api.py:51`).
@@ -144,9 +145,9 @@ There is **no existing `planner`, `release`, or `view` command** — this is net
    `create --file` resolves it against the **target release's gridview**, warning about
    (and omitting) tests not present in the new release while still creating the plan.
    This is entirely client-side — no new backend endpoint.
-6. `update` speaks the diff-based `PlanDiffPayload` contract from
-   `backend/release-planner-multi-edit-safety`: it computes add/remove deltas for
-   lists/maps and only sends changed scalar fields.
+6. `update` speaks the diff-based `PlanDiffPayload` contract (merged to master,
+   `planner_service.py:62`): it computes add/remove deltas for lists/maps and only
+   sends changed scalar fields.
 7. `copy` runs the (existing) backend eligibility check, prints dropped entities as a
    warning, drops them by default, and honors an optional `--replacements` JSON file.
 8. Output respects the global `--text`/JSON modes and `--no-color`; every command
@@ -181,7 +182,7 @@ There is **no existing `planner`, `release`, or `view` command** — this is net
   - `PlanDiffRequest` (diff-based update: `id`, optional scalars, `tests_add`,
     `tests_remove`, `groups_add`, `groups_remove`, `participants_add`,
     `participants_remove`, `assignee_mapping_set`, `assignee_mapping_remove`) —
-    mirrors `PlanDiffPayload` from `backend/release-planner-multi-edit-safety`.
+    mirrors `PlanDiffPayload` (`planner_service.py:62`, on master).
   - `CopyPlanRequest` (`plan`, `keepParticipants`, `replacements`, `targetReleaseId`,
     `targetReleaseName`) and `CopyCheckResponse` (`status`, `targetRelease`,
     `originalRelease`, `missing.tests`, `missing.groups`).
@@ -341,8 +342,6 @@ exceptions are `--plan-id` and `--view-id`). Add to `PlannerService`:
 
 ### Phase 5 — Write: `update` (diff-based)
 **Importance: Important**
-**Depends on:** `backend/release-planner-multi-edit-safety` being merged (or the diff
-contract being final). Until then, mark as `blocked` and gate on the backend PR.
 
 - `PlannerService.UpdatePlan(ctx, PlanDiffRequest)` → POST `/planning/plan/update`
   with the diff payload; invalidate plan + plan-list caches.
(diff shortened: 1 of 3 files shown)


soyacz (commented):
caveman was here I see.
Can you add example how updating plan payload would look like? Will LLM be able to directly modify it (e.g. in file)?

k0machi:
> caveman was here I see. Can you add example how updating plan payload would look like? Will LLM be able to directly modify it (e.g. in file)?

Added.

docs/plans/cli-planner-command.md
```diff
@@ -0,0 +1,448 @@
...
+- **Command** (`cmd/planner/planner_test.go`): flag parsing/required-flag enforcement,
+  confirmation-prompt bypass with `--yes`, `--file`/stdin parsing, flag-over-file
+  precedence (pattern from `cmd/discussions/discussions_test.go`).
+
+### Integration tests
+- No new ScyllaDB-backed pytest suite required (CLI talks HTTP). If desired, a
+  `@pytest.mark.docker_required` smoke could exercise `planner_api.py` end-to-end, but
+  it is out of scope for this plan.
+
+### Manual testing (against staging)
```
fruch, line 783: I would recommend using LLM with this CLI to test:

Creation from scratch based on existing plan

Assignment of people to some jobs

Add/remove a job from plan

If LLM would be able to follow those set of actions is ready

soyacz (commented):
generally, I think we operate too much on UUID's.
I'd prefer making this cli and payloads working with names.
So instead release id we would provide a name of release
the same for groups and tests.
I understand this is more things to do on Argus side, but I'm afraid LLM with big number of UUID's will start to hallucinate (change slightly sometimes).
Also, reviewing what it wants to do is harder with UUID's.

As to workflow, I'd ppropose another flow: first we get the plan as json, This json is human (and machine) readable, meaning it will contain release/group/test names instead of uuid's.
Then LLM will be able to better tell if we cover certain features or not based on test names (and metadata) and easier manipulate.
We could also provide it with tests in SCT repo (from jenkinsfiles names), maybe even we could have metadata in SCT repo that could be visible for LLM to construct a new plan.
Having such plan modified by LLM we could asses visually if it's right (with uuid's won't be possible). 
Then such file could be a payload for plan create endpoint

@k0machi would this be feasible?

k0machi:
> generally, I think we operate too much on UUID's. I'd prefer making this cli and payloads working with names. So instead release id we would provide a name of release the same for groups and tests. I understand this is more things to do on Argus side, but I'm afraid LLM with big number of UUID's will start to hallucinate (change slightly sometimes). Also, reviewing what it wants to do is harder with UUID's.
> 
> As to workflow, I'd ppropose another flow: first we get the plan as json, This json is human (and machine) readable, meaning it will contain release/group/test names instead of uuid's. Then LLM will be able to better tell if we cover certain features or not based on test names (and metadata) and easier manipulate. We could also provide it with tests in SCT repo (from jenkinsfiles names), maybe even we could have metadata in SCT repo that could be visible for LLM to construct a new plan. Having such plan modified by LLM we could asses visually if it's right (with uuid's won't be possible). Then such file could be a payload for plan create endpoint
> 
> @k0machi would this be feasible?

It is feasible indeed, I was thinking about making cli operate on names instead of uuids and I should probably add this to the plan. As for json flow - that's basically what I have in the plan right now but with names instead of uuids. I will update the plan.

soyacz (commented):
won't it add complexity to support both and UUID's?

k0machi:
> won't it add complexity to support both and UUID's?

in the end the normal API accepts UUIDs, so just passing them through is the easiest thing to support. All resolution happens client-side and while it adds a bit of complexity it should be just tiered resolution:

1. Check if UUID.
2. Check if valid build_system_id
3. Check if group/name format is used
4. Check if it's just a test name

After that it resolves to a final uuid which gets submitted.

soyacz:
> > won't it add complexity to support both and UUID's?
> 
> in the end the normal API accepts UUIDs, so just passing them through is the easiest thing to support. All resolution happens client-side and while it adds a bit of complexity it should be just tiered resolution:
> 
> 1. Check if UUID.
> 2. Check if valid build_system_id
> 3. Check if group/name format is used
> 4. Check if it's just a test name
> 
> After that it resolves to a final uuid which gets submitted.

As we spoke, adding UUID support will add some complexity. Especially on the client side, where more options won't bring benefits (no one will use UUID's as we don't know them and LLM's can mistake them).
We can keep UUID's on the backend side (for Argus frontend), but CLI should not expose it.

docs/plans/cli-planner-command.md
```diff
@@ -0,0 +1,597 @@
...
+work for every release version. There are also no CLI primitives to discover the test
+and group UUIDs a plan must reference — that discovery is locked inside the web
+search box (`/api/v1/planning/search`).
+
+Measurable pain points:
+- **0** CLI commands exist for plan management today (`cli/cmd/` has `run`, `comment`,
+  `ssh`, `config`, etc., but no `planner`).
+- Creating a plan for N release versions requires N manual web sessions; there is no
+  way to template/script plan creation or to copy a plan across releases from CI.
+- Test/group UUIDs (required by the create/update payloads) are not retrievable from
```
soyacz, line 25: this is not precisely true. But indeed it's a pain point to work with UUID's on CLI level. In web view we don't care, all is done by frontend part and UUID's are not exposed in any way to the users)

docs/plans/cli-planner-command.md
```diff
@@ -0,0 +1,597 @@
...
+- `create_plan` reads `assignments` (entity_id→user_id) and converts to
+  `assignee_mapping` (`planner_service.py:119-153`). `view_id` optional → a view is
+  auto-created when absent.
+- Create/update/copy all enforce a uniqueness guard on `(name, target_version)` and
+  raise `PlannerServiceException` on collision.
+- `copy_plan` (`planner_service.py:296`) remaps tests/groups by `build_system_id`
+  substring-replacing the source release name with the target release name; entities
+  with no match are dropped unless present in `payload.replacements`.
+
+#### Upcoming update mechanism — diff-based (ref `backend/release-planner-multi-edit-safety`)
```
soyacz: let's treat it as existing mechanism in this doc as we will base on it

docs/plans/cli-planner-command.md
```diff
@@ -0,0 +1,597 @@
...
+  diffs directly without a full re-send.
+- Assignment edits map to `assignee_mapping_set` (entity→user) and
+  `assignee_mapping_remove` (entity ids). **Both sides accept names only (no UUIDs):**
+  the entity key resolves via `ResolveEntityID` (`build_system_id` / group-qualified /
+  unique bare name) and the user value via `ResolveUserID` before the map is built.
+- **`--add-group` always expands (Phase 2 `ExpandGroup`):** the group's enabled tests
+  are emitted as `tests_add`, no `groups_add` is sent, and any assignment on the group
+  fans out to each expanded test in `assignee_mapping_set`.
+- **DoD:**
+  - [ ] `argus planner update --plan-id <id> --name "new"` sends only the `name`
```
soyacz, line 615: can we use also names here?

k0machi: Probably not. The names are not very stable - which one are you going to use here, display name (which gets converted to a safe url name for internal 'name' field) but might contain spaces and such. I think using ID here is appropriate.

soyacz: but people don't know the plan id - we don't expose it anywhere.
LLM will have to do some steps to resolve the plan uuid anyway after getting request e.g. 'changing assignee to xxx for a test xxx in plan scylla-2026.2'

k0machi: We expose it actually - https://argus.scylladb.com/plan/74d6870a-44a0-11f1-a829-599c49cb3004 for example.

I do agree it's not very nice to use, however I don't like the UX of referring to a plan by its display name - in this case that would be `scylla-2026.2 - 2026.2.0 Release Plan`. I think we should introduce easy key of some sort to refer to a plan, like `$releaseName#$number`, e.g. `scylla-2026.2#1`

soyacz: I don't know how people work with these plans, but for me was very hard to get to the point I could see the plan id. I went to https://argus.scylladb.com/view/scylla-2026-2-2026-2-0-release-plan and didn't see it.
But I see the point. In that case we'll need an api&cli command to fetch plan id's so LLM can find the plan based on the user prompt (and if finds multiple ones, LLM will ask which to update).

k0machi: I can add plan keying into this plan and implement that as an extra field (New index)

soyacz: we don't have that many plans - we can just fetch all plan names and their id's and return it (maybe adding basic 'contain' filter, so we can fetch keys for all 2026.2 plans)

k0machi: Was thinking about that as well, we can just filter all plans and find the key there, so it'll be just an undexed field of `releaseName#planNumber`

soyacz (commented):
Right now, the plan is a bit tough to reason through because it focuses heavily on implementation details and module architecture too early.

At this stage, we should focus entirely on the CLI contract and the user experience. We want to fit our solution to the requirements, rather than forcing the requirements into our existing technical setup.

@k0machi Could you use an LLM to revamp this plan to focus strictly on the surface level? Specifically, we need:

- Proposed CLI commands and their exact parameters.
- Example outputs (like a mock --help flag for each command).

We can map out the backend gaps and internal module details later during implementation.

k0machi:
> Right now, the plan is a bit tough to reason through because it focuses heavily on implementation details and module architecture too early.
> 
> At this stage, we should focus entirely on the CLI contract and the user experience. We want to fit our solution to the requirements, rather than forcing the requirements into our existing technical setup.
> 
> @k0machi Could you use an LLM to revamp this plan to focus strictly on the surface level? Specifically, we need:
> 
> * Proposed CLI commands and their exact parameters.
> * Example outputs (like a mock --help flag for each command).
> 
> We can map out the backend gaps and internal module details later during implementation.

Added, also removed some fluff.

soyacz (commented):
Things to fix:
1. examples should show 2026.1 etc. not 6.2 (we no longer do OSS releases)
2. search command move out of planner (just make it general command)
3. improve search command docs what we can search with it

Otherwise, LGTM

PR #1014, opened 2026-06-05
refactor(client/tunnel): convert tunnel_* files into tunnel package module

CodeLieutenant:
## Summary

Addresses [Petr's review comment](https://github.com/scylladb/scylla-cluster-tests/pull/14728#discussion_r3309256014) on the SCT PR:

> "This is an issue for the client, but I think we should have been using module here, instead of using tunnel_* files. The prefix itself signals that these should be grouped into a module"

The four flat `tunnel_*.py` files and the `tunnel.py` re-export facade have been reorganised into a proper subpackage:

```
argus/client/tunnel/
├── __init__.py   ← public re-exports (was tunnel.py)
├── models.py     ← was tunnel_models.py
├── state.py      ← was tunnel_state.py
├── api.py        ← was tunnel_api.py
└── ssh.py        ← was tunnel_ssh.py
```

## This is not a breaking change

All public import paths are **identical** before and after this refactor.

Python resolves `tunnel/__init__.py` and `tunnel.py` identically, so every existing import continues to work without modification:

```python
from argus.client.tunnel import SSHTunnel, TunnelConfig, TunnelClientError
from argus.client.tunnel import resolve_tunnel_config, resolve_tunnel_config_with_reason
from argus.client.tunnel import delete_cached_tunnel_state
```

**Clients only need a version bump** — no code changes required.

The only paths that no longer exist are the four internal implementation modules (`argus.client.tunnel_api`, `argus.client.tunnel_state`, etc.), which were never exported from `argus/client/__init__.py` and were not part of any public API.


argus/client/tunnel/__init__.py
```diff
@@ -0,0 +1,13 @@
+from .api import resolve_tunnel_config, resolve_tunnel_config_with_reason
```
k0machi: What's the justification for relative imports?

CodeLieutenant: keeping backwards compatibility, nothing more

k0machi: I'd say replace them with absolute imports, I don't think this package will be used outside of argus client.

CodeLieutenant: Done

argus/client/tunnel/api.py
```diff
@@ -5,8 +5,8 @@
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 
-from argus.client.tunnel_models import DEFAULT_TUNNEL_TIMEOUT, TunnelClientError, TunnelConfig
-from argus.client.tunnel_state import (
+from .models import DEFAULT_TUNNEL_TIMEOUT, TunnelClientError, TunnelConfig
+from .state import (
```
k0machi: Ditto

argus/client/tunnel/ssh.py
```diff
@@ -9,15 +9,15 @@
 import tempfile
 import time
 
-from argus.client.tunnel_models import (
+from .models import (
```
k0machi: Ditto

argus/client/tunnel/ssh.py
```diff
@@ -9,15 +9,15 @@
...
     ALLOWED_HOST_KEY_TYPES,
     DEFAULT_RECONNECT_RETRIES,
     DEFAULT_TUNNEL_TIMEOUT,
     MAX_PORT_BIND_ATTEMPTS,
     TunnelClientError,
     TunnelConfig,
 )
-from argus.client.tunnel_state import get_tunnel_state_paths
+from .state import get_tunnel_state_paths
```
k0machi: Ditto

argus/client/tunnel/state.py
```diff
@@ -21,7 +21,7 @@
     PrivateFormat = None
     PublicFormat = None
 
-from argus.client.tunnel_models import TunnelClientError, TunnelConfig, TunnelStatePaths, parse_datetime
+from .models import TunnelClientError, TunnelConfig, TunnelStatePaths, parse_datetime
```
k0machi: Ditto

k0machi (changes requested):

soyacz (approved):
LGTM

argus/client/tests/test_tunnel.py
pehala, line 119: Unrelated to this PR, but very ugly.

argus/client/tests/test_tunnel.py
```diff
@@ -6,9 +6,9 @@
 import pytest
 
 from argus.client.base import ArgusClient
-from argus.client import tunnel_api
-from argus.client import tunnel_ssh
-from argus.client import tunnel_state
+from argus.client.tunnel import api as tunnel_api
+from argus.client.tunnel import ssh as tunnel_ssh
+from argus.client.tunnel import state as tunnel_state
```
pehala, line 11: nit: this is a weird structure to use for testing, usually you import directly what you test from each of them.

I suspect this is because you use a lot of protected names (i.e. starting with _) which shouldnt be protected if you are testing them.

pehala (approved):

argus/client/tests/test_tunnel.py
CodeLieutenant: This is actually for better naming in the tests, as in `tunnel.api` would be really ugly to see `tunnel.tunnel_api` but later in the tests using just `api` makes no sense, so it's just importing and changing names.

pehala: Why do you need to use modules in test? Why not import what you need from the module instead?

You wouldnt use `tunnel.api` you would use `api`

CodeLieutenant: Yes, using just api is confusing, so using tunnel_api is better in my opinion

k0machi (approved):

PR #1016, opened 2026-06-08
docs(plans): add Event Summarization implementation plan

CodeLieutenant:
Full plan for LLM-based summarization of SCT ERROR/CRITICAL events via a new argusAI worker (queue -> batched OpenAI call -> stored summary), with best-effort write-once semantics and UI/CLI display falling back to the original message.

## Open questions — Event Summarization

1. **What to summarize (granularity).** Working default is **per-event** (simplest UI/CLI
   mapping, mirrors embeddings, and what the batching design assumes). Two unresolved levers:
   - **Long-events-only vs. everything** — `EVENT_SUMMARIZATION_MIN_LENGTH` (`-1` = all). Needs a
     team call + the cost eval.
   - **Per-run digest** — a single consolidated summary per run is the most aggressive token cut
     for Zeus, but it's a different data model + UI and can't satisfy the per-event display
     requirement alone; candidate for a separate additive phase, not a replacement.
2. **`SctEventSummary` keying** (the "reference the original event" decision). Recommended:
   partition `(run_id, severity)` + clustering `ts` (cheap partition-read attach in `get_events`)
   **plus** an `event_id` column as the explicit cross-reference. **Decision needed:** is keying
   by that shape + `event_id`-as-column enough, or do we need `event_id` independently queryable
   (→ add a secondary index)?
3. **Keyspace** — `SctEventSummary` in `argus_tablets` (with the embedding tables) or the default
   keyspace.
4. `OPENAI_SUMMARY_MODEL` default (which model is "good enough")

## Evaluation: token usage & cost (TBD — must be measured)

These numbers are **not yet known** and gate the config defaults in this change.
Treat `OPENAI_SUMMARY_MODEL`, `EVENT_SUMMARIZATION_BATCH_SIZE`, and
`EVENT_SUMMARIZATION_MIN_LENGTH` defaults as placeholders until measured.

**Data source:** every OpenAI response returns a `usage` object (`prompt_tokens`,
`completion_tokens`, cached tokens) — log it per call to sum and price.

**Open question — does batching actually cost less?** Expectation: batching only saves the
fixed per-call overhead (system prompt + schema) paid once vs. N times; event-text and summary
tokens are identical either way, and prompt caching already discounts the repeat in one-by-one
mode. So the cost win is likely modest (biggest on small events, ~zero on large ones) while the
**latency / rate-limit win is the decisive one.** Confirm with real numbers.

Refs: https://scylladb.atlassian.net/browse/ARGUS-146

### DIFF
--- a/docs/plans/event-summarization.md
+++ b/docs/plans/event-summarization.md
@@ -1,0 +1,365 @@
+---
+status: draft # draft | approved | in_progress | blocked | complete
+domain: infrastructure
+created: 2026-06-04
+last_updated: 2026-07-03
+owner: CodeLieutenant
+---
+
+# Event Summarization — Design
+
+Refs: [ARGUS-146](https://scylladb.atlassian.net/browse/ARGUS-146) ·
+review discussion in [PR #1016](https://github.com/scylladb/argus/pull/1016)
+
+## 1. Problem Statement
+
+SCT ERROR/CRITICAL events are long and noisy: a single event message routinely runs hundreds
+to thousands of characters (the frontend already hard-truncates at 600 chars —
+`frontend/TestRun/SCT/SctEvent.svelte:30`, `MESSAGE_CUTOFF = 600`).
+
+The primary consumer this hurts is **Zeus**, the AI investigation service. Zeus loads a run's
+events into its conversation context at the start of an investigation, and those events are
+then **re-read on every subsequent step** — and most runs are investigated by Zeus multiple
+times. Every re-read pays the full token cost of the raw event bodies again. Long,
+low-signal-to-noise inputs also degrade LLM reasoning quality (hallucination risk).
+
+Summarizing each unique event **once**, with a cheap model, and storing the result turns a
+per-read cost into a one-time cost: every later read (by Zeus, and optionally by humans)
+consumes the short summary instead of the raw wall of text.
+
+For **humans** the value is contested (see the PR discussion): a summary is easier to scan,
+but it is preprocessed information — a triager verifying an LLM's conclusions wants the raw
+message, and a misleading summary is worse than a long one. The design therefore treats
+human-facing display as **opt-in, configurable per user**: the original message stays the
+default, the summary is one click away, and each user can flip which one they see first.
+
+## 2. Current State
+
+**Event pipeline (backend):**
+
+- `argus/backend/plugins/sct/testrun.py:79` — `SCTEvent` model (`sct_event`). PK = partition
+  `(run_id, severity)` + clustering `ts` (ASC). Fields incl. `event_id`, `event_type`,
+  `message` (required), `duplicate_id`.
+- `argus/backend/plugins/sct/testrun.py:108` — `SCTUnprocessedEvent`
+  (`sct_unprocessed_events`): the processing queue drained by the argusAI worker.
+- `argus/backend/plugins/sct/service.py:478` — `SCTService.submit_event` saves the event and
+  (line 512) enqueues an `SCTUnprocessedEvent` for ERROR/CRITICAL severities.
+- `argus/backend/plugins/sct/service.py:528` — `SCTService.get_events` delegates to
+  `SCTTestRun.get_events_limited` (`testrun.py:372`), which runs `SELECT * FROM sct_event ...`
+  and returns raw row dicts — **any new column on `sct_event` is returned automatically**.
+
+**The argusAI worker** (`argusAI/`, same repo — deployed as a separate systemd service):
+
+- `argusAI/event_similarity_processor_v2.py` — standalone worker process. Infinite loop:
+  fetch up to 100 queue rows (`_process_batch`), and per event: read message → sanitize
+  (`MessageSanitizer`) → generate embedding (local ONNX model, fast) → **duplicate check**
+  via ANN search (`_mark_event_is_duplicate`, line 126; short-circuit at line 284-285: a
+  duplicate gets `duplicate_id` set on `sct_event` and is dequeued, skipping all remaining
+  steps) → store embedding → delete queue row.
+- DB access is **raw CQL** via `argusAI/utils/scylla_connection.py` (`ScyllaConnection`,
+  prepared-statement cache). The worker never calls `cqlengine.connection.setup()` — CQLEngine
+  model methods (`.save()`, `.objects`) would raise in this process. Config comes from
+  `Config.load_yaml_config()` (plain dict, no Flask app context).
+- Deployed via `argusAI/deployment/argusai_event_similarity_processor.service`.
+
+**Duplicate handling downstream:** the frontend groups duplicates under the canonical event
+(`SctEvents.svelte`, `duplicate_id`); the CLI aggregates them too
+(`cli/cmd/run_nemeses_events.go:125`, `deduplicateEvents`). Both display surfaces already
+show the **canonical** event as the representative.
+
+**Frontend:** `frontend/TestRun/SCT/SctEvent.svelte` renders a single event; it already has a
+truncate/expand toggle (`fullMessage` `$state`, line 159; button at line 366).
+
+**CLI (Go):** `argus run events` — `runEventsCmd` in `cli/cmd/run_nemeses_events.go:176`;
+the `SCTEvent` struct is `cli/internal/models/runs.go:198`; output goes through
+`out.Write(models.SCTEventsResponse{...})` with a response cache.
+
+**Zeus:** external service, reached through the authenticated proxy
+`/api/v1/zeus/<path>` (`argus/backend/controller/api.py:628`); it reads run events through
+the Argus API.
+
+**Dependencies:** no OpenAI/LLM dependency exists anywhere in the repo today; `openai` is not
+in `pyproject.toml`.
+
+## 3. Goals
+
+1. Every **unique** (non-duplicate) ERROR/CRITICAL event gets a stored, model-generated
+   summary, best-effort, within the worker's normal processing lag. Duplicates are never
+   summarized — the canonical event's summary covers them (this is also where the cost
+   saving comes from: duplicates are the bulk of event volume).
+2. The summary is **stored once and reused by all investigations** — every Zeus read of the
+   run, and every human view, across the event's lifetime.
+3. Serving is zero-cost: `get_events` payloads carry a `summary` field with no extra query.
+4. **Fail silently:** an event without a summary (not yet processed, summarization failed,
+   feature disabled, worker down) renders exactly as today, everywhere — `summary` is `null`
+   and consumers fall back to `message`. No errors, no blank states.
+5. Human display is **configurable per user**: which body renders first (original vs.
+   summary) is a user preference; the other is always one click/flag away. Default:
+   **original first** (summaries are opt-in for humans).
+6. **Embeddings are never slowed, blocked, or altered by summarization.** The embedding path
+   stays byte-for-byte as-is and dequeues on its own schedule; an OpenAI outage has zero
+   effect on it.
+7. Metrics are defined **up front** (token usage, cost, latency, coverage) so the pilot can
+   be priced and the model choice validated with real numbers.
+
+## 4. Non-Goals (decisions from the PR review)
+
+These were in earlier drafts and are **deliberately dropped**:
+
+- **No separate summary table.** The summary is a column on `sct_event`. This removes the
+  keying and keyspace questions entirely, and removes the attach-query from serving.
+- **No second queue, worker, or systemd unit.** The existing similarity processor is the
+  single consumer of the single existing queue; this also eliminates the shared-queue race
+  that a second consumer would introduce. No new enqueue in `submit_event`.
+- **No batching.** One model call per event. There are several minutes of headroom before
+  auto-investigation picks up a run, so per-event latency is fine; batching's
+  chunking/char-budget/index-mapping complexity buys little (event tokens cost the same
+  either way; prompt caching already discounts the repeated system prompt).
+- **No truncation.** Never trim event input before summarizing — the tail of a stack trace
+  is exactly the detail a triager needs. Large events are sent as-is.
+- **No sanitization of summarizer input.** `MessageSanitizer` strips node names, IPs, and
+  paths — right for embedding similarity, wrong for summaries, where those are the details
+  worth keeping (see the worked example in the PR discussion, which preserves node name and
+  event id). The summarizer receives the raw message.
+- **No min-length gate.** Every unique ERROR/CRITICAL event is summarized, full stop.
+- **No `status=failed` markers, no retries.** On failure: log, leave `summary` null, move
+  on. Null already means "fall back to the original".
+- **No live model swap.** Changing the model is a config edit + worker restart.
+- **No per-run digest.** Different data model and UI; out of scope.
+- **No summarization before embedding.** Embeddings continue to consume the sanitized raw
+  message unchanged.
+
+## 5. Design
+
+### 5.1 Data model
+
+Add one nullable column to `SCTEvent` (`argus/backend/plugins/sct/testrun.py:79`):
+
+```python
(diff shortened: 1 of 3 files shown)


soyacz:
A few decisions to simplify this before implementation:

**Summarize every event** — per-event, no min-length param. Every ERROR/CRITICAL event gets summarized, full stop.

**Store `summary` directly on `SCTEvent`, drop the separate `SctEventSummary` table.** Add one nullable `summary` (Text) column to `sct_event`. `get_events_limited` already returns full rows, so the summary rides along for free — no second query, no N+1, and **Open Questions 2 and 3 (keying + keyspace) disappear**. Summary write is a raw-CQL upsert: `UPDATE sct_event SET summary=? WHERE run_id=? AND severity=? AND ts=?`. Removes most of Phase 3. (No `status=failed` marker needed — serving falls back to the original on null.)

**Reuse the similarity processor — one queue, one worker.** Don't add a second queue (`SctUnsummarizedEvent`), second worker, or second systemd unit, and don't add a second enqueue in `submit_event`. This also **eliminates the shared-queue race risk** — single consumer. When the processor picks up an event, it dispatches **two independent tasks: embedding (as today) and summarization, running in parallel.** From that point they don't depend on each other.

**Embeddings must never be slowed or blocked by summarization.** The embedding path stays byte-for-byte as-is and dequeues on its own schedule. Summarization runs as a parallel background task (bounded concurrency to protect OpenAI rate limits); on completion it does the `UPDATE` above, on failure it just logs and the event keeps `summary=null` (UI/CLI fall back to original). OpenAI outage/latency has zero effect on embeddings. Duplicates already short-circuit in the processor, so they're never summarized — correct and saves cost.

**No batching — one model call per event.** We have several minutes of headroom before autoinvestigation picks up a run, so per-event latency is fine; favor simplicity. No chunking/char-budget/index-mapping.

**No live model-swap.** Swapping the model = worker restart.

**No truncation.** Never trim event input before summarizing — that's exactly the detail a triager needs. Send large events as-is.

Start with `gpt-5-mini`; finalize the model on the cost eval.

soyacz:
@CodeLieutenant ^^

docs/plans/event-summarization.md
```diff
@@ -0,0 +1,405 @@
...
+---
+
+# Event Summarization
+
+## 1. Problem Statement
+
+SCT ERROR/CRITICAL events are long and noisy: a single event message routinely runs
+hundreds to thousands of characters (the frontend already hard-truncates at 600 chars —
+`frontend/TestRun/SCT/SctEvent.svelte:30`, `MESSAGE_CUTOFF = 600`). Raw event text is the
+primary input to our downstream AI investigation flow ("Zeus") and to humans triaging runs.
```
pehala: IMO the current events are fine for humans. I can see why it would be bad for zeus but the question is wont it introduce another level of uncertainty? E.g. if the summarization is on point in 90% cases and the Zeus is correct in 80% cases from these summarizations, total correct ratio decreases to 72% (0.9 * 0.8)

docs/plans/event-summarization.md
```diff
@@ -0,0 +1,405 @@
...
+
+This causes three measurable pains:
+
+1. **Token cost & latency in Zeus** — feeding full raw events inflates prompt token counts,
```
pehala: You will have to input the same amount of tokens into the summarization LLM and then the output will be again inputted to Zeus. I would be interested in analysis of pricing for this and if it actually improves actual costs

docs/plans/event-summarization.md
pehala: Dont forget that the 90% of accuracy would affect humans as well, which until now "enjoyed" 100% accuracy of what they see, because it wasn't preprocessed. For this fact alone I fear it might be detriment to human investigators.

Doesnt need to completely wrong, but misleading is enough to completely mitigate gains for humans. If we want to go forward with this I would like this to be opt-in for humans at start and as I personally would like to view raw events

docs/plans/event-summarization.md
```diff
@@ -0,0 +1,405 @@
...
+   which directly increases cost and investigation time per run.
+2. **Hallucination risk** — long, low-signal-to-noise event bodies degrade the quality of
+   downstream LLM reasoning.
+3. **Human readability** — triagers scroll through walls of stack traces to find the point.
```
pehala: IMO: not that big of deal. It is actually helpful to see where the error originated and we dont have 100 of different errors per run and same errors are already grouped togethere

docs/plans/event-summarization.md
```diff
@@ -0,0 +1,405 @@
...
+
+We want a concise, model-generated summary attached to each ERROR/CRITICAL event, stored for
+reuse, surfaced in the UI/CLI, with the full original always one click/flag away. When no
+summary exists, the system must silently fall back to the original message — no errors, no
+blank states.
```
pehala: Reuse is insides of one testrun or across tests?

If across tests, why dont we invest in better messages in SCT instead? Wouldnt that address problem at source?

pehala (commented):
Based on the design, I dont see this helpful for humans and I worry it will be detrimental more than  helpful for them.

For AI usage the overall idea makes sense, but I would like to some pricing calculations that it is worth it and expected token saved. Also I would liek to see PoC (can be manual) of the summaries vs non-summarises for Zeus so we could evaluate improvement in response quality. AI response quality is still enigma and I would like to make sure we have confidence in the idea first

docs/plans/event-summarization.md
soyacz: Your calculations are wrong. minimizing context makes LLM's work better: shorter events can yield better results. Example shrinking a very long (~3057 tokens) with 'haiku 4.5 model' and simple prompt (possibly to improve) to ~163 tokens:
`Summarize this error event in a way we don't lose information and use minimum tokens. Don't add anything more from your side.`:
```
  (DatabaseLogEvent Severity.ERROR) period_type=one-time event_id=5e953e08-c9e8-4ff5-962a-65ea9c0fbed2: type=OVERSIZED_ALLOCATION regex=seastar_memory - oversized allocation: line_number=12421413 node=longevity-many-clients-4h-2026-2-db-node-d651a7eb-4
  2026-06-25T10:53:21.042 longevity-many-clients-4h-2026-2-db-node-d651a7eb-4 !WARNING | scylla[9482]  [shard 22:strm] seastar_memory - oversized allocation: 135168 bytes. This is non-fatal, but could lead to latency and/or fragmentation issues. Please report: at 0x31f03eb 0x31f00a3 0x50ccadb 0x17ff5d3 /opt/scylladb/libreloc/libcryptopp.so.8+0x1de3ff /opt/scylladb/libreloc/libstdc++.so.6+0x151717 /opt/scylladb/libreloc/libstdc++.so.6+0x158f77 /opt/scylladb/libreloc/libstdc++.so.6+0x14a237 0x4d8ae73 0x4d8acef 0x27a1cd7 0x27a1bb3 0x279fccf 0x4d93993 0x4d93937 0x4d938eb 0x4d99ae3 0x4c327ab 0x180b987 0x180ba6b 0x180a223 0x51136bb 0x50d42b3 /opt/scylladb/libreloc/libc.so.6+0x8f773 /opt/scylladb/libreloc/libc.so.6+0xfabdb
  Backtrace:
[...]
```
to:
```
**OVERSIZED_ALLOCATION** 135KB on node longevity-many-clients-4h-2026-2-db-node-d651a7eb-4, shard 22 (non-fatal warning).

**Root cause**: JSON serialization of task manager response → std::string::reserve in libcryptopp via libstdc++.

**Call chain**: seastar::json::formatter → task_manager_json::task_identity list → basic_stringbuf::overflow → string allocation.

**Trigger**: HTTP request handler resuming coroutine during task status JSON formatting.

**Event ID**: 5e953e08-c9e8-4ff5-962a-65ea9c0fbed2 | **Timestamp**: 2026-06-25T10:53:21.042 UTC
```
Such thing is way easier to read for humans too.
Besides we plan to have an option to look at original one.
LLM's are best at summarizing (language transformation)

docs/plans/event-summarization.md
soyacz: for summarization we can use cheap model (gpt-5-mini costs ~20x less than Opus 4.8).
Also, events are loaded to agents context each time they look at it. Sometimes loaded multiple times during one investigation.

docs/plans/event-summarization.md
soyacz: see the example above. I want still keep possibility to look into original event (with easy to access button)

docs/plans/event-summarization.md
soyacz: Reuse means "summarize once and save for all investigations"
SCT overhead is not that big, often we just copy into the event - so, we can't do much.

docs/plans/event-summarization.md
```diff
@@ -0,0 +1,405 @@
...
+  loads + sanitizes each event message (raw CQL), groups eligible events into **chunks**, and
+  makes **one model call per chunk** (see "Batching" below). For each event in the chunk that
+  came back with a summary → `INSERT` a row (`status=ok`, model name); for events the model
+  omitted, or on a whole-chunk failure → `INSERT` a `status=failed` row (+ short `error`); then
+  `DELETE` all the chunk's `sct_unsummarized_events` rows. The PK is the event's identity, so a
+  re-attempt (e.g. process crash between INSERT and DELETE — same window the embedding worker
+  has) simply **upserts/overwrites** the prior row; effectively idempotent. We never
+  *deliberately* retry; dropping some is acceptable. Throughput-maximizing, not exhaustive.
+- **Model must be easily replaceable in production (A/B testing).** Wrap the LLM call behind a
+  thin `Summarizer` abstraction (small module under `argusAI/utils/`) so the concrete model —
```
fruch: We need to define it's metrics as we are doing for Zeus up front, so we can assess the pricing / usage of it

fruch (commented):
Also we'll need to assess the cost of it up front.

And even when implementing, doing it in phases and piloting on handful of jobs first

docs/plans/event-summarization.md
pehala: Could we make this configurable? I want it exactly opposite, original error first and one easy button click summarization

docs/plans/event-summarization.md
pehala: > Reuse means "summarize once and save for all investigations"

For all investigations inside one job or across the jobs as well?

soyacz:
One clarification for the cost concerns:
We only need to summarize **deduplicated** events, not all the error/critical events. @CodeLieutenant, please update the plan accordingly.
Most runs investigations now use Zeus (often multiple times), which loads events into the conversation context at the start, these events are then re-read on every subsequent Zeus step, summarizing them upfront will give us a major cost reduction on those repetitive reads.
and obviously we'll reduce costs on cheaper model.

docs/plans/event-summarization.md
soyacz: all

docs/plans/event-summarization.md
pehala: I disagree, IMO the oversized allocation event you posted is as easy to read as the summarization you posted and I have certainty that LLM didnt remove something important. As a final chain in the verification, I want to have as much raw information as possible to be able to verify what LLM wrote.

They is a class of errors which are horrible to investigate: Stress command failures (QUORUM, data validation, manager failures), where this might be useful because in SCT we dont provide enough context inside the error message. However, this would be better solved on SCT sides, unless we want to put all of the logs inside the summarization LLM context, which might defeat the point

`**Trigger**: HTTP request handler resuming coroutine during task status JSON formatting.` -> I dont know how accurate this is, I wouldn't trust it much (I dont have data about accurancy from developers so it is just an opinion)
`
**Call chain**: seastar::json::formatter → task_manager_json::task_identity list → basic_stringbuf::overflow → string allocation. `-> I think developers would like it more if we supplied the full stacktrace instead. I dont think this summarization is enough to start fixing

TLDR: I am fine with this being used for LLM, but I think we should make it configurable for humans (i.e. what is shown first) as I dont see much value there for human investigations and I would be happy proven wrong once it is deployed.

docs/plans/event-summarization.md
soyacz: sure, we can do it the opposite way, or make it configurable per user

fruch: People are lazy, if summaries would be good, they would look only on that.

docs/plans/event-summarization.md
fruch: This idea isn't holding anyone from improving SCT failure reports, where it's possible.

Just careful with putting actual data, or it's gonna end up like some of Gemini reports of failure

soyacz:
One more thing to include in this spec: add evaluation tests: take 20-50 events, summarize with opus 4.8 (or gpt-5.5) and then use cheaper model and evaluate if response is the same (or similar)

docs/plans/event-summarization.md
pehala: > sure, we can do it the opposite way, or make it configurable per user

I like per user the best, it will fit all usages and each of us can configure it how he likes it

docs/plans/event-summarization.md
```diff
@@ -0,0 +1,365 @@
...
+| Rate limits at high event volume | Medium | Medium | Bounded executor concurrency; duplicates (bulk of volume) never reach OpenAI. |
+| Lost summaries on worker crash/restart (no retry) | Medium | Low | Accepted under best-effort: fallback is the original message. A backfill pass over `summary IS NULL` events can be added later if coverage proves too low. |
+| Raw event content (node names, IPs, paths) sent to OpenAI | Low | Medium | Deliberate (§4 — sanitization would strip the useful details); events contain test-infrastructure data, not customer data; flag to security if that assumption ever changes. |
+| Secret leakage (`OPENAI_API_KEY` in YAML) | Low | High | Existing `ZEUS_TOKEN`/`*_SECRET` conventions; placeholder only in the example file. |
+| Stale CLI cache hides fresh summaries | High | Low | Accepted: TTL-bounded; summaries are not time-critical for CLI users. |
+
+## 11. Open Questions
+
+1. **Final model** — `gpt-5-mini` is the working default; the evaluation test (§7) decides.
+2. **Per-user preference storage** — localStorage (per-browser, zero backend work, matches
```
soyacz: yes, localStorage

docs/plans/event-summarization.md
```diff
@@ -0,0 +1,365 @@
...
+   existing patterns) vs. a server-side `User` column (cross-device). Proposal: start with
+   localStorage; revisit only if users ask.
+3. **Zeus payload shape** — is "prefer `summary` when non-null" enough on the Zeus side, or
```
soyacz: cli default should be summary and giving an option to see full message. Full messages should be available when '--raw' is provided (and Zeus should be instructed to limit fetched events by time/limit)

docs/plans/event-summarization.md
```diff
@@ -0,0 +1,365 @@
...
+   should Argus offer a summaries-only events view to guarantee the token saving?
+4. **Record the producing model?** Dropped with the separate table (no `model` column on
```
soyacz: no need for that

soyacz (approved):
Good design. I'm ok with it - just make LLM to address according to my comments in open questions

docs/plans/event-summarization.md
CodeLieutenant: Updated §11.2 to state this as a decision (localStorage), not a proposal — closing this out.

docs/plans/event-summarization.md
CodeLieutenant: Updated §5.6 — CLI now defaults to summary-first output with a `--raw` flag for the original message. Added a note to §5.7 that Zeus should bound its event fetches by time range/limit per investigation step (independent of the CLI default). §11.3 now records this as decided.

docs/plans/event-summarization.md
CodeLieutenant: Agreed — updated §11.4 to state this as decided (no `summary_model` column).

k0machi (approved):

PR #1018, opened 2026-06-10
feature(jenkins): billing_project choices from Jenkins in SCT wizard

soyacz:
## What

Adds `billing_project` to the SCT run/rebuild wizard as a dropdown whose options come directly from the Jenkins job, keeping Argus choices in sync with Jenkins.

## Changes

- **`jenkins_service.py`**: new `_extract_choice_parameters()` parses `ChoiceParameterDefinition` choices from the job config XML (handles both nested `ArrayList` and flat `<choices><string>` forms). `retrieve_job_parameters()` attaches `choices` to the returned parameters; `Parameter` TypedDict gains a `choices` field.
- **`SCTParameterWizard.svelte`**: new `billing_project` `SelectParam` in the Environment section, populated from `rawParams[...].choices`. Auto-hidden when the job does not define the parameter (existing `internalName in params` gate).
- **tests**: unit coverage for choice extraction (nested / flat / none).

## Notes

- Choices are fetched live from Jenkins — no hardcoded list, no drift.
- Backward compatible: non-choice params keep rendering as before.

closes ARGUS-104

### DIFF
--- a/argus/backend/service/jenkins_service.py
+++ b/argus/backend/service/jenkins_service.py
@@ -1,6 +1,6 @@
 import re
 import requests
-from typing import Any, TypedDict
+from typing import Any, NotRequired, TypedDict
 from uuid import UUID
 import xml.etree.ElementTree as ET
 import jenkins
@@ -19,6 +19,7 @@ class Parameter(TypedDict):
     name: str
     description: str
     value: Any
+    choices: NotRequired[list[str]]
 
 
 class JenkinsServiceError(Exception):
@@ -51,6 +52,22 @@ class JenkinsService:
                                         username=current_app.config["JENKINS_USER"],
                                         password=current_app.config["JENKINS_API_TOKEN"])
 
+    @staticmethod
+    def _extract_choice_parameters(config: ET.Element) -> dict[str, list[str]]:
+        """Map each ChoiceParameterDefinition name to its list of choices.
+
+        Handles both the nested ArrayList form
+        (<choices class="..."><a><string>..) and the flat form
+        (<choices><string>..).
+        """
+        choices: dict[str, list[str]] = {}
+        for define in config.iterfind(".//hudson.model.ChoiceParameterDefinition"):
+            name = define.findtext("name")
+            if not name:
+                continue
+            choices[name] = [s.text for s in define.findall("choices//string") if s.text is not None]
+        return choices
+
     def retrieve_job_parameters(self, build_id: str, build_number: int | None) -> list[Parameter]:
         job_info = self._jenkins.get_job_info(name=build_id)
         if not build_number:
@@ -73,6 +90,7 @@ class JenkinsService:
             }
         else:
             descriptions = {}
+        choices = self._extract_choice_parameters(config)
         params = next((a for a in build_info["actions"] if a.get(
             "_class", "#NONE") == "hudson.model.ParametersAction"), None)
         if params:
@@ -84,6 +102,11 @@ class JenkinsService:
                 "value", "")} for param in default_params if param["name"] != self.RESERVED_PARAMETER_NAME]
         for idx, param in enumerate(params):
             params[idx]["description"] = descriptions.get(param["name"], "")
+            if param["name"] in choices:
+                param_choices = choices[param["name"]]
+                params[idx]["choices"] = param_choices
+                if param_choices and param.get("value") not in param_choices:
+                    params[idx]["value"] = param_choices[0]
 
         return params
 
--- a/frontend/TestRun/Jenkins/ParameterEditor.svelte
+++ b/frontend/TestRun/Jenkins/ParameterEditor.svelte
@@ -76,7 +76,13 @@
                 {#if param.value || paramTab == 2}
                     <div class="mb-1">
                         <label class="form-label fw-bold" for={param.name}>{param.name}</label>
-                         {#if (param._class || "").includes("TextParameter")}
+                         {#if param.choices?.length}
+                            <select class="form-select" id={param.name} bind:value={buildParams[param.name]}>
+                                {#each param.choices as choice}
+                                    <option value={choice}>{choice}</option>
+                                {/each}
+                            </select>
+                        {:else if (param._class || "").includes("TextParameter")}
                             <textarea class="form-control" id={param.name} bind:value={buildParams[param.name]} rows="4"></textarea>
                         {:else}
                             <input class="form-control" id={param.name} type="text" bind:value={buildParams[param.name]}>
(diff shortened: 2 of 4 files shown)


frontend/TestRun/Jenkins/ParameterEditor.svelte
```diff
@@ -76,7 +76,13 @@
...
                     <div class="mb-1">
                         <label class="form-label fw-bold" for={param.name}>{param.name}</label>
-                         {#if (param._class || "").includes("TextParameter")}
+                         {#if param.choices?.length}
+                            <select class="form-select" id={param.name} bind:value={buildParams[param.name]}>
+                                {#each param.choices as choice}
+                                    <option value={choice}>{choice}</option>
+                                {/each}
+                            </select>
+                        {:else if (param._class || "").includes("TextParameter")}
```
k0machi, line 85: Why is this duplicating functionality of selectparam? Can I see a demo?

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
```diff
@@ -564,6 +577,16 @@
         }
     });
 
+    onMount(() => {
+        // Ensure billing_project holds a valid choice so it renders as a dropdown
+        // (SelectParam falls back to a disabled text input when the value is not
+        // among the choices). Done in onMount to avoid mutating state during render.
```
k0machi: That doesn't sound right

argus/backend/service/jenkins_service.py
```diff
@@ -19,6 +19,7 @@ class Parameter(TypedDict):
     name: str
     description: str
     value: Any
+    choices: list[str]
```
k0machi: This should be tagged as optional

frontend/TestRun/Jenkins/ParameterEditor.svelte
soyacz: yes, deployed to staging.
Maybe it is not needed (but I had problems when developing, we have 2 places with it: wizard and 'all parameters' tab

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
soyacz: ah, could be partially wrong. 'onMount' was needed, otherwise I got error about incorrect state mutation

k0machi: You should be able to use `onShow` function in the paramater definition, also check XML as it should provide a value instead (even if empty), I don't like hardcoding parameter logic into onMount at all.

k0machi: And looking into xml itself, the `<value>` tag is populated with selected option for this param so that should be used as a selector

frontend/TestRun/Jenkins/ParameterEditor.svelte
k0machi: Right, I guess this bit needs to be looked at later, this is the generic all params section.

argus/backend/service/jenkins_service.py
soyacz: Done — `choices` is now `NotRequired[list[str]]` (only set when the job defines them).

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
soyacz: Reworked. The backend now supplies a valid `value` for choice params (first choice when unset/invalid), so the wizard needs no client-side defaulting at all — dropped both the `onMount` block and the `onShow` attempt. The `onShow` route actually threw `state_unsafe_mutation` since `condition: () => true` makes it mutate state during the initial render; the region params dodge it only because their condition starts false. Letting the backend provide the value (your XML `<value>` point) is the clean fix.

frontend/TestRun/Jenkins/ParameterEditor.svelte
soyacz: Left as-is for now — this is the generic "All parameters" tab, which renders raw params rather than going through `SelectParam`. Agreed it can be unified later.

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
```diff
@@ -186,6 +186,19 @@
                     values: ["spot", "on_demand", "spot_fleet"],
                     labels: ["Spot", "On Demand", "Spot Fleet"]
                 },
+                billingProject: {
+                    name: "Billing Project",
+                    description: "Billing project to charge the run to. Choices come from the Jenkins job.",
+                    type: SelectParam,
+                    internalName: "billing_project",
+                    // Shown only when the job defines billing_project (gated by internalName in params)
+                    // Value is provided by the backend (first choice when unset), so no onShow default needed.
```
k0machi: Pointless comment

argus/backend/service/jenkins_service.py
```diff
@@ -84,6 +102,13 @@ def retrieve_job_parameters(self, build_id: str, build_number: int | None) -> li
                 "value", "")} for param in default_params if param["name"] != self.RESERVED_PARAMETER_NAME]
         for idx, param in enumerate(params):
             params[idx]["description"] = descriptions.get(param["name"], "")
+            if param["name"] in choices:
+                param_choices = choices[param["name"]]
+                params[idx]["choices"] = param_choices
+                # Ensure the value is a valid choice so the frontend renders a
+                # populated dropdown instead of falling back to a text input.
```
k0machi: Pointless comment

k0machi (approved):
LGTM but minor style fixes.

frontend/TestRun/Jenkins/SCTParameterWizard.svelte
soyacz: removed

argus/backend/service/jenkins_service.py
soyacz: removed

cezarmoise:
@soyacz This should be reverted.
Right now, if I try to rebuild something that doesn't have a billing project, it will auto add the first one in the list.

For example, https://argus.scylladb.com/tests/scylla-cluster-tests/3338f43b-ce50-4ba9-86f6-17215a735f5c, will have the billing project `Pay.com reproduction`, and it's hidden on the second page of the wizard.

soyacz:
> @soyacz This should be reverted. Right now, if I try to rebuild something that doesn't have a billing project, it will auto add the first one in the list.
> 
> For example, https://argus.scylladb.com/tests/scylla-cluster-tests/3338f43b-ce50-4ba9-86f6-17215a735f5c, will have the billing project `Pay.com reproduction`, and it's hidden on the second page of the wizard.

fixing it here: https://github.com/scylladb/argus/pull/1019

soyacz:
@cezarmoise btw. did it really picked billing project? I don't see it was selected in the runs

cezarmoise:
> @cezarmoise btw. did it really picked billing project? I don't see it was selected in the runs

Yes. Added reproduction here https://scylladb.atlassian.net/browse/ARGUS-165.

PR #1021, opened 2026-06-16
improvement(backend/planner): Introduce plan keys

k0machi:
This commit introduces a new mechanism to fetch a plan from backend,
using a human-readable 'key' in the form of 'release#number' to easily fetch
a specific plan by name. Other operations like copy, update, delete also
support fetching by key.

No tests were added as that part is handled by #982

Task: ARGUS-147

### DIFF
--- a/argus/backend/models/plan.py
+++ b/argus/backend/models/plan.py
@@ -22,6 +22,7 @@ class ArgusReleasePlan(Model):
     creation_time = columns.DateTime(default=lambda: datetime.datetime.now(tz=datetime.UTC))
     last_updated = columns.DateTime(default=lambda: datetime.datetime.now(tz=datetime.UTC))
     ends_at = columns.DateTime()
+    key = columns.Text()
 
     def __eq__(self, other):
         if isinstance(other, ArgusReleasePlan):
--- a/argus/backend/service/planner_service.py
+++ b/argus/backend/service/planner_service.py
@@ -139,6 +139,23 @@ class PlanningService:
     def version(self):
         return "v1"
 
+    def _generate_plan_key(self, release_id: UUID | str) -> str:
+        release: ArgusRelease = ArgusRelease.get(id=release_id)
+        candidate = f"{release.name}#1"
+        release_plans = list(ArgusReleasePlan.filter(release_id=release.id).allow_filtering().all())
+        if len(release_plans) == 0:
+            return candidate
+        existing_keys = [int(p.key.split("#")[1]) for p in release_plans]
+        previous_number = max(existing_keys)
+
+        return f"{release.name}#{previous_number+1}"
+
+    def _resolve_plan(self, ref: str | UUID) -> ArgusReleasePlan:
+        try:
+            return ArgusReleasePlan.get(id=UUID(str(ref)))
+        except (ValueError, ArgusReleasePlan.DoesNotExist):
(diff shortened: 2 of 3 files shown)


soyacz (commented):
## Review — plan keys (ARGUS-147)

Mechanism is sound and well-factored: `key` column + `_generate_plan_key` (stamp on create/copy) + `_resolve_plan` (UUID-or-key, replacing every bare `get(id=...)`). The `existing.id != plan.id` change in `update_plan` is correct and necessary now that `plan_request.id` may be a key. Migration assumed run before use.

### Should fix

- **Race → duplicate keys.** `_generate_plan_key` does read-max-then-`+1` with no lock and `key` has no unique constraint. Two concurrent `create_plan`/`copy_plan` on the same release both produce `#N+1`; duplicates persist silently and a later `_resolve_plan(key).get()` then raises `MultipleObjectsReturned`. At minimum document the collision window; ideally accept it explicitly or derive the number differently.

- **`release.name` containing `#` breaks parsing.** `int(p.key.split("#")[1])` takes the *second* segment, not the last. A name like `5.2#rc` → key `5.2#rc#1` → `split("#")[1] == "rc"` → `int()` crash, breaking all subsequent creates for that release. Use `rsplit("#", 1)[-1]`.

- **`_generate_plan_key` still crashes on any un-keyed plan.** Even with the migration run, it skips plans whose release is `DoesNotExist`, leaving `key=None`; `int(p.key.split(...))` then throws. Cheap guard: `[... for p in release_plans if p.key]`.

- **`_resolve_plan` key path raises uncaught.** Only the UUID branch handles `DoesNotExist`. `filter(key=...).get()` can raise `DoesNotExist`/`MultipleObjectsReturned` straight to the endpoint. Wrap or let callers handle consistently.

### Consider

- **`allow_filtering()` full scans.** `_resolve_plan`'s key path runs on every update/delete/copy/complete/owner-change as a cluster-wide scan. Add `index=True` to `key` for a fast lookup (and it would also make a uniqueness check feasible).

### Tests

Deferred to #982 per description — fine. Flag the duplicate-key and `#`-in-name paths there explicitly; they're the failure modes most likely to bite.

k0machi:
> ### Should fix
> * **Race → duplicate keys.**

Not a problem in our use case.

> * **`release.name` containing `#` breaks parsing.** `int(p.key.split("#")[1])` takes the _second_ segment, not the 

Release name cannot have `#` as that is a part of a URL fragment.

> * **`_generate_plan_key` still crashes on any un-keyed plan.** Even with the migration run, it skips plans whose release is `DoesNotExist`, leaving `key=None`; `int(p.key.split(...))` then throws. Cheap guard: `[... for p in release_plans if p.key]`.

You cannot create a plan for a release that doesn't exist.

> * **`_resolve_plan` key path raises uncaught.** Only the UUID branch handles `DoesNotExist`. `filter(key=...).get()` can raise `DoesNotExist`/`MultipleObjectsReturned` straight to the endpoint. Wrap or let callers handle consistently.

It can return both of those exceptions in this case in either way, both do `.get`.

> ### Consider
> * **`allow_filtering()` full scans.** `_resolve_plan`'s key path runs on every update/delete/copy/complete/owner-change as a cluster-wide scan. Add `index=True` to `key` for a fast lookup (and it would also make a uniqueness check feasible).

Index is more expensive in this case. Also no, uniqueness is not enforced.

soyacz (approved):
LGTM

PR #1022, opened 2026-06-17
improvement(cli/internal): Add models and route definitions for Plans

k0machi:
This commit adds new routes and models into CLI application that
describe envelopes and responses for endpoints corresponding to plan
management inside argus. It also adds some basic tests for model tabular
related methods.

Task: ARGUS-147

### DIFF
--- a/cli/internal/models/planner.go
+++ b/cli/internal/models/planner.go
@@ -1,0 +1,262 @@
+package models
+
+import "strconv"
+
+// ---------------------------------------------------------------------------
+// ReleasePlan – mirrors argus.backend.models.plan.ArgusReleasePlan
+// ---------------------------------------------------------------------------
+
+// ReleasePlan is a single release plan as returned by the planner endpoints.
+//
+// All UUID columns are serialised as strings by ArgusJSONEncoder; datetime
+// columns are ISO-8601 strings. AssigneeMapping maps an entity UUID (test or
+// group) to a user UUID.
+//
+// Key is the human-friendly "releaseName#planNumber" handle (e.g.
+// "scylla-2026.2#3") added by the backend plan-keying change; plan operations
+// accept it in place of a UUID.
+type ReleasePlan struct {
+	ID              string            `json:"id"`
+	Key             string            `json:"key"`
+	Name            string            `json:"name"`
+	Description     string            `json:"description"`
+	Owner           string            `json:"owner"`
+	Participants    []string          `json:"participants"`
+	TargetVersion   string            `json:"target_version"`
+	AssigneeMapping map[string]string `json:"assignee_mapping"`
+	ReleaseID       string            `json:"release_id"`
+	Tests           []string          `json:"tests"`
+	Groups          []string          `json:"groups"`
+	ViewID          string            `json:"view_id"`
+	CreatedFrom     string            `json:"created_from"`
+	Completed       bool              `json:"completed"`
+	CreationTime    string            `json:"creation_time"`
+	LastUpdated     string            `json:"last_updated"`
+	EndsAt          string            `json:"ends_at"`
+}
+
+// ReleasePlanList is the response payload for the plans-for-release endpoint.
+type ReleasePlanList = []ReleasePlan
+
+// Headers implements output.Tabular for ReleasePlan.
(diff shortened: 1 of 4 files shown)


soyacz (commented):
## Review — models + routes scaffolding for Plans

Cross-checked every route path and struct field against `planner_api.py`, `planner_service.py`, and `models/plan.py`. Routes all match. Field mapping accurate except one bug. This PR is type/route definitions only (no command wiring), so the issues below are latent — they bite once commands are wired.

### 🔴 Bug — `CopyPlanRequest.Plan` typed as `ReleasePlan` will break the copy endpoint
- Backend `copy_plan` does `TempPlanPayload(**payload["plan"])` (`planner_api.py:137`). `TempPlanPayload` is a frozen dataclass with **no `key` field**.
- `ReleasePlan.Key` has no `omitempty` → always serialized → server raises `TypeError: __init__() got an unexpected keyword argument 'key'`.
- Fix: a dedicated copy-plan struct without `Key` (mirror `TempPlanPayload`), or have the backend tolerate extra keys.

### 🟡 `CreatePlanRequest.Assignments` nil → `null` → server crash
- No `omitempty`; a nil Go map marshals as `"assignments":null`. Backend does `plan_request.assignments.items()` (`planner_service.py:162`) → `AttributeError` on `None`.
- `TestCreatePlanRequest_OmitsOptionalEmptyFields` constructs without `Assignments`, proving it serializes `null`. The command layer must always send `{}`. Same nil→null risk for `participants`/`tests`/`groups` (less critical — assigned directly to columns).

### 🟡 Response models describe the inner payload only
- All endpoints wrap as `{status, response: ...}`; search nests further (`response.hits`). `ReleasePlanList`, `SearchHitList`, `Release`, users all model the **inner `response`**, not the envelope. Fine if the API client unwraps centrally — worth a comment so consumers don't double-wrap.

### 🟢 Minor
- `Releases` comment says `all=1`, but backend uses truthy presence (`request.args.get("all", False)`), so `all=0` also returns everything. Usage guidance, not wrong.
- No model for the `resolve_entities` response though the route exists — presumably a later PR.

### Strengths
- Doc comments cite each Python counterpart + serialization notes; verified accurate, low comment-rot risk.
- Test coverage targets the right surface: `omitempty` behavior, `[]` vs `null`, row/header parity, string-vs-object dual modeling (`GridEntity` vs `SearchHit`/`SearchRef`).
- Pointer-for-optional diff pattern (distinguishing unset from zero) is correct and idiomatic.

**Verdict:** Solid scaffolding. Fix the `CopyPlanRequest.Plan`/`key` mismatch before the copy command lands — the one real correctness bug. Handle assignments-null in the command layer.

k0machi:
> 🔴 Bug — CopyPlanRequest.Plan typed as ReleasePlan will break the copy endpoint

Part of #1021

cli/internal/models/planner.go
```diff
@@ -0,0 +1,262 @@
...
+// ReleasePlan – mirrors argus.backend.models.plan.ArgusReleasePlan
+// ---------------------------------------------------------------------------
+
+// ReleasePlan is a single release plan as returned by the planner endpoints.
+//
+// All UUID columns are serialised as strings by ArgusJSONEncoder; datetime
+// columns are ISO-8601 strings. AssigneeMapping maps an entity UUID (test or
+// group) to a user UUID.
+//
+// Key is the human-friendly "releaseName#planNumber" handle (e.g.
```
soyacz, line 15: this is the only human-friendly thing? how about testsgroups/releases?

k0machi: those fields already exist and their resolution logic will be implemented in the next pr

soyacz (approved):
LGTM
@CodeLieutenant i see tests are failing `"TypeError: ArgusGenericClient.__init__() missing 1 required positional argument: 'log_dir'` - how did it pass when it was introduced?

k0machi:
> LGTM @CodeLieutenant i see tests are failing `"TypeError: ArgusGenericClient.__init__() missing 1 required positional argument: 'log_dir'` - how did it pass when it was introduced?

It was using a cached version of argus client from master instead of building the branch, so it was not visible there.

PR #1024, opened 2026-06-17
fix(pytest-argus-reporter): support log_dir/use_tunnel, require py>=3.12

CodeLieutenant:
argus-alm 0.15.17 made log_dir a required ArgusGenericClient argument (the request replay log feature) and also bumped its own minimum Python to 3.12. The reporter still constructed the client without log_dir, and advertised support for Python 3.10/3.11 where argus-alm>=0.15.17 cannot be installed -- so those runners silently fell back to an old client without log_dir, yielding "unexpected keyword argument 'log_dir'".

- Forward the new client knobs: --argus-log-dir (env ARGUS_LOG_DIR, default ".") -> log_dir --argus-use-tunnel / --argus-no-use-tunnel -> use_tunnel (defaults to None so the existing ARGUS_USE_TUNNEL env fallback is preserved)
- Raise requires-python to >=3.12 to match argus-alm; trim the nox matrix and classifiers to 3.12-3.14.
- nox tests now install the in-repo argus client (-e ..) so the reporter is verified against this repo's client API, not a published/cached wheel.

### DIFF
--- a/pytest-argus-reporter/noxfile.py
+++ b/pytest-argus-reporter/noxfile.py
@@ -3,11 +3,14 @@ import nox
 nox.options.default_venv_backend = "uv"
 
 
-@nox.session(python=["3.10", "3.11", "3.12", "3.13", "3.14"])
+@nox.session(python=["3.12", "3.13", "3.14"])
 def tests(session):
     """Run the test suite with coverage."""
 
-    session.install(".[dev]")
+    # Install the in-repo argus client (editable) so the reporter is tested
+    # against this repo's argus-alm rather than a cached/published wheel that
+    # may lag behind the client API the reporter targets.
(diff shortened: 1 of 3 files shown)


k0machi (approved):

soyacz:
2025.1 in SCT uses python 3.10, I don't remember dtest branches where we post results to Argus.
Why can't we maintain compatibility with 3.10?
@CodeLieutenant

fruch:
> 2025.1 in SCT uses python 3.10, I don't remember dtest branches where we post results to Argus.
> Why can't we maintain compatibility with 3.10?
> @CodeLieutenant 

The title of this is a bit confusing, cause there's a package named py, didn't even look at this, things it's about a specific package

Please refer to the actual reason we dropped the support for...

PR #1025, opened 2026-06-18
fix(api): remove 60s cache headers from structural release endpoints

pehala:
## Problem

When editing releases, groups, or tests in the admin UI, some structural API responses could remain stale in the browser for up to 60 seconds because these endpoints explicitly returned `Cache-Control: max-age=60`:

- `/releases`
- `/groups`
- `/tests`
- `/release/<id>/details`
- `/group/<id>/details`
- `/test/<id>/details`

These endpoints serve small, cheap queries, so the browser cache adds little value and can delay the UI from reflecting recent edits.

### `argus/backend/controller/api.py`
- Removed `response.cache_control.max_age = 60` from the six structural list/detail endpoints above.
- Returned the JSON responses directly where the cache header was no longer needed.

### DIFF
--- a/argus/backend/controller/api.py
+++ b/argus/backend/controller/api.py
@@ -59,7 +59,6 @@ def releases():
         "response": [dict(d.items()) for d in all_releases if d.enabled or force_all]
     })
 
-    response.cache_control.max_age = 60
     return response
 
 
@@ -304,12 +303,10 @@ def argus_groups():
     groups = service.get_groups(UUID(release_id))
     result_groups = [dict(g.items()) for g in groups if g.enabled or force_all]
 
-    response = jsonify({
+    return jsonify({
         "status": "ok",
         "response": result_groups
     })
-    response.cache_control.max_age = 60
-    return response
 
 
 @bp.route("/tests", methods=["GET"])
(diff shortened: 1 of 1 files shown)


argus/backend/controller/api.py
```diff
@@ -304,12 +303,10 @@ def argus_groups():
     groups = service.get_groups(UUID(release_id))
     result_groups = [dict(g.items()) for g in groups if g.enabled or force_all]
 
-    response = jsonify({
+    return jsonify({
```
k0machi: jsonify is technically not needed, can just return dict object.

k0machi (changes requested):
What happened in this PR? I see 5 commits but final change set is just cache control lines.

pehala:
> What happened in this PR? I see 5 commits but final change set is just cache control lines.

Good catch, it undergone a bunch of rebases and it seems I messed them up

pehala:
> What happened in this PR? I see 5 commits but final change set is just cache control lines.

I fixed both commits (They got mangled as it was developed in parallel to the first cache PR and I migrated changes from this to there), description and name, it should all make sense now

argus/backend/controller/api.py
pehala: Should I remove it from all cases here?

k0machi: Yes

k0machi:
@pehala double check the PR and I'll merge it - there's `jsonify` comment still there.

pehala:
> @pehala double check the PR and I'll merge it - there's `jsonify` comment still there.

Finally fixed, I also changed rest of the endpoints in the file (remove the jsonify) even if they didnt have the cache control header, is that fine?

k0machi:
> > @pehala double check the PR and I'll merge it - there's `jsonify` comment still there.
> 
> Finally fixed, I also changed rest of the endpoints in the file (remove the jsonify) even if they didnt have the cache control header, is that fine?

It is fine, I have merged the changes into #1059, so now this PR can be closed.

PR #1026, opened 2026-06-18
docs(plans): release dashboard / views unification plan

pehala:
## Summary

Adds a full implementation plan (`docs/plans/dashboard-view-unification.md`) for unifying the **release dashboard** and **Views** onto one shared stats collector, one snapshot cache, and one invalidation path.

This PR is **plan/docs only** — no code changes. It registers the plan in `MASTER.md` and `progress.json`.

## Why

The release dashboard and Views render the same test grid but run on divergent backends: release stats are cached/indexed, view stats recompute on every request and multiply work per widget. The release dashboard is effectively a constrained special case of a View (fixed test set + fixed widget layout), so the two can share one path.

## Phasing

- **Phase A — implementable now (internal, no UI/URL change).** Generic scoped snapshot store, shared collector extraction, view stats caching with version-scoped per-view invalidation, view versions/images via union of the release index tables, `resolve/tests` dedupe, invalidation instrumentation, and cache architecture docs + diagrams. Phase A is self-contained and delivers the view performance win on its own.
- **Phase B — for later.** Model the release dashboard as a lazy system view, widgetize ReleaseActivity, route `/dashboard/<name>` through the unified view path, migrate release snapshots to the generic store, and remove the bespoke release path. Depends on Phase A.

## Notes

- Plan follows the repo's 7-section format per `docs/plans/INSTRUCTIONS.md`.
- Includes a Performance Model section (per-test caching tradeoffs, hybrid invalidation, broad-view aggregate ceiling) and explicit "Needs Investigation" items.
- Draft — open for review of the approach before any implementation begins.

### DIFF
--- a/docs/plans/MASTER.md
+++ b/docs/plans/MASTER.md
@@ -17,7 +17,9 @@ Central tracking for all Argus implementation plans. Each plan is listed under i
 
 ## Backend
 
-_No plans in this domain yet._
+| Plan                                | Status  | File                                                           | Owner |
+| ----------------------------------- | ------- | -------------------------------------------------------------- | ----- |
+| Release Dashboard / Views Unification | `draft` | [dashboard-view-unification.md](dashboard-view-unification.md) | null  |
 
 ## Frontend
 
--- a/docs/plans/dashboard-view-unification.md
+++ b/docs/plans/dashboard-view-unification.md
@@ -1,0 +1,341 @@
+---
+status: draft
+domain: backend
+created: 2026-06-18
+last_updated: 2026-06-18
+owner: null
+---
+
+# Release Dashboard / Views Unification
+
+## 1. Problem Statement
+
+The release dashboard and the Views dashboard render the same test-grid UI but run on
+divergent backend paths. Release stats are cached and indexed; view stats are not.
+
+Measured/observed pain (single test-dashboard widget, page `/view/scylla-master-master-duty`):
+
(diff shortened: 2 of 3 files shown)


soyacz:
@pehala we're changing our SDLC to base each solution on the `intent.md` file (tell LLM to prepare it based on this PR). Before that, please create Jira for this effort (required) and we'll plan it for upcoming sprints.

pehala:
> @pehala we're changing our SDLC to base each solution on the `intent.md` file (tell LLM to prepare it based on this PR). Before that, please create Jira for this effort (required) and we'll plan it for upcoming sprints.

I will change the Pr for the intent, however, the goal here was also to get approvals for this effort, if we we can agree that this is more or less direction we want to take, I will create issues and assign it to myself (I think I can deliver this one to completion albeit slowly)

If there is no agreement yet, we can discuss it over the intent.md PR and then create issues once this is clearer

soyacz:
> > @pehala we're changing our SDLC to base each solution on the `intent.md` file (tell LLM to prepare it based on this PR). Before that, please create Jira for this effort (required) and we'll plan it for upcoming sprints.
> 
> I will change the Pr for the intent, however, the goal here was also to get approvals for this effort, if we we can agree that this is more or less direction we want to take, I will create issues and assign it to myself (I think I can deliver this one to completion albeit slowly)
> 
> If there is no agreement yet, we can discuss it over the intent.md PR and then create issues once this is clearer

yes, let's have a discussion over intent.md PR (still requires Jira, can be just a placeholder)

PR #1029, opened 2026-06-22
fix(client): support Python 3.10+ and normalize Z-suffixed timestamps

CodeLieutenant:
The argus client and pytest-argus-reporter are consumed by environments running Python 3.10/3.11 (e.g. SCT), but the tunnel client used 3.11+ only constructs that broke at import time:

- datetime.UTC -> timezone.utc (UTC alias is 3.11+)
- typing.NotRequired -> required base + total=False TypedDict split (3.11+)

Also fix tunnel timestamp parsing: Python 3.10's datetime.fromisoformat rejects the trailing 'Z' UTC designator that the tunnel API emits, so normalize 'Z' to '+00:00' before parsing. Covered by new regression tests.

Lower requires-python to >=3.10 for argus-alm and pytest-argus-reporter so the client installs on 3.10, and extend the plugin nox matrix to 3.10/3.11. The backend keeps targeting 3.12+ and is unchanged.

Closes: https://scylladb.atlassian.net/browse/ARGUS-174

### DIFF
--- a/argus/client/session.py
+++ b/argus/client/session.py
@@ -4,7 +4,7 @@ import os
 import threading
 import time
 import weakref
-from datetime import UTC, datetime
+from datetime import datetime, timezone
 
 import requests
 from requests.adapters import HTTPAdapter
@@ -226,7 +226,7 @@ class TunneledSession(requests.Session):
             self._tunnel = tunnel
             self._tunnel_config = config
             self._tunnel_port = local_port
-            self._tunnel_established_at = datetime.now(UTC).isoformat()
+            self._tunnel_established_at = datetime.now(timezone.utc).isoformat()
             self._tunnel_warning_emitted = False
             self._tunnel_disabled_until = 0.0
             LOGGER.info(
--- a/argus/client/tests/test_tunnel.py
+++ b/argus/client/tests/test_tunnel.py
@@ -1,4 +1,4 @@
-from datetime import UTC, datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from io import StringIO
 from unittest.mock import Mock
 import os
@@ -11,6 +11,7 @@ from argus.client.tunnel import ssh as tunnel_ssh
 from argus.client.tunnel import state as tunnel_state
 from argus.client.session import TunneledSession
 from argus.client.tunnel import TunnelConfig
+from argus.client.tunnel.models import parse_datetime
 
 
 def _write_text(path: str, text: str) -> None:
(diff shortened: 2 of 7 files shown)


k0machi:
I don't like adding hacks for older versions of python, can we confirm which SCT branches still run 3.10? I see py314 in its pyproject and >3.10 <3.15 constraint.

Update: Ugh, 2025.x branches all do use 3.10. Can you separate out this logic to a different module and use some sort of version checking importing substitute functions instead of how it is right now?

fruch:
> I don't like adding hacks for older versions of python, can we confirm which SCT branches still run 3.10? I see py314 in its pyproject and >3.10 <3.15 constraint.
> 
> Update: Ugh, 2025.x branches all do use 3.10. Can you separate out this logic to a different module and use some sort of version checking importing substitute functions instead of how it is right now?

I don't see no hacks here, it's basic support for older versions compatibility.
no reason to complicate it further.

just to keep in mind the client version support need to be more rigid then Arugs server code.

fruch (approved):

fruch:
> > I don't like adding hacks for older versions of python, can we confirm which SCT branches still run 3.10? I see py314 in its pyproject and >3.10 <3.15 constraint.
> > Update: Ugh, 2025.x branches all do use 3.10. Can you separate out this logic to a different module and use some sort of version checking importing substitute functions instead of how it is right now?
> 
> I don't see no hacks here, it's basic support for older versions compatibility. no reason to complicate it further.
> 
> just to keep in mind the client version support need to be more rigid then Arugs server code.

you should have seen python2 and python3 support in the same code base, back then... those would have multiple packages to handle the diff (i.e. six)

soyacz (approved):
LGTM — clean, well-scoped fix. Verified the shipped surface (`argus.client*` + `argus.common*` + top-level `argus`) has no leftover 3.11+ constructs, and `argus.backend` is excluded from the wheel via `packages.find`, so `requires-python = ">=3.10"` stays honest. `parse_datetime` Z-normalization handles microseconds/explicit-offset/naive correctly on 3.10, and the new test assertions check genuinely-correct expected values.

One non-blocking gap: **the `tunnel` module this PR makes 3.10-safe is never imported or tested on 3.10/3.11 in CI.** `test.yml` pins `uv run pytest` (which collects the new `argus/client/tests/test_tunnel.py` and imports `tunnel/models.py`) to Python 3.12; the only 3.10/3.11 leg is `pytest-argus-reporter`'s nox, which runs in its own dir and doesn't collect the tunnel tests. So re-adding `from datetime import UTC` (or `NotRequired`) to `argus/client/tunnel/` would keep all CI green yet break again at import on SCT's 3.10 — the exact regression this PR fixes, unguarded against recurrence.

Suggest a 3.10 leg for the root client suite — either a `python-version` matrix in `test.yml`, or a client nox session mirroring the reporter's.

PR #1030, opened 2026-06-26
cli/planner commands management

k0machi:
- **feature(cli/search): top-level argus search for plan discovery**
  Add the top-level 'argus search' command that queries the planning search
  endpoint for tests/groups/releases and prints their build_system_id as the
  canonical reference for planner commands. PlannerService.Search resolves an
  optional --release name to its id, passes the free-text/facet query through
  verbatim, and filters out the synthetic 'Add all...' special row. Corrects
  the plan's query-language note: facet values must not be quoted or contain
  spaces.
  
  Implements Phase 3 of docs/plans/cli-planner-command.md.
  

- **feature(cli/planner): create and delete commands with template round-trip**
  Add `argus planner create` and `argus planner delete`, plus `get --template`.
  
  create accepts a --file/-f JSON spec (and stdin) and repeatable flags
  (--test, --group, --assign entity=user, --participant) with flags overriding
  or augmenting file fields. All references are by name or build_system_id and
  resolved to UUIDs client-side; groups always expand to their enabled tests and
  group assignments fan out to each test, so the backend only ever receives test
  UUIDs. Missing tests are warned about and omitted; ambiguous refs abort.
  
  get --template emits a release-independent, name-based plan spec (group/test
  names, usernames) that round-trips back through create. delete addresses a plan
  by UUID or key, toggles view deletion via --delete-view, and confirms unless
  --yes is given.
  

- **feature(cli/planner): resolve names in get/list output with --raw escape hatch**
  Render get/list with human-readable resolved data by default (release,
  owner, qualified test/group names, target version) instead of raw UUIDs.
  Text output shows a curated column set (Key, Name, Description, Release,
  Target Version, Owner, Last Updated) while JSON keeps full detail.
  
  Resolution is lossless: unresolved references fall back to their raw
  UUIDs. Add a --raw flag to both commands to restore the previous
  unresolved output; on get it is mutually exclusive with --template.
  

- **feature(cli/planner): diff-based update command**
  Add 'argus planner update' which edits a plan by sending only what changed.
  Changes come from a --file diff and/or flags (scalars override the file,
  collections augment it). Every reference is by name or build_system_id and
  resolved client-side to UUIDs.
  
  The diff builder resolves owner to a UUID, emits tests_add/tests_remove
  (missing tests warn-and-skip, ambiguous aborts), expands --add-group to its
  enabled tests instead of groups_add, maps --remove-group to groups_remove,
  resolves participants, and builds assignee_mapping_set/remove with group
  assignments fanned out to each enabled test. Unchanged scalars are never sent.
  

Fixes ARGUS-153

### DIFF
--- a/cli/cmd/planner/create.go
+++ b/cli/cmd/planner/create.go
@@ -1,0 +1,180 @@
+package planner
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/scylladb/argus/cli/internal/cmdctx"
+	"github.com/scylladb/argus/cli/internal/logging"
+	"github.com/scylladb/argus/cli/internal/models"
+	"github.com/scylladb/argus/cli/internal/services"
+	"github.com/spf13/cobra"
+)
+
+// registerCreate adds the "create" sub-command to parent.
+func registerCreate(parent *cobra.Command) {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create a release plan",
+		Long: `Create a release plan from a JSON file and/or flags.
+
+The plan can be described by a --file (the same schema 'get --template' emits)
+and/or built up with flags; flags override matching fields from the file. Every
+reference is by name (or, for tests, build_system_id) and resolved to a UUID
+automatically — raw UUIDs are not accepted.
+
+  # From an edited template (release-independent round-trip):
+  argus planner create --file plan.json
+
+  # From flags:
+  argus planner create --release scylla-2026.2 --name "2026.2 Longevity" \
+    --owner alice --test scylla-2026.2/tier1/longevity-100gb \
+    --group tier2 --assign tier1/longevity-200gb=bob
+
+Groups (from a file or --group) are always expanded to their enabled tests; an
+assignment on a group fans out to each of those tests. Tests that do not exist
+in the target release are reported and omitted, and the plan is created anyway.`,
+		RunE: runCreate,
+	}
+
+	cmd.Flags().StringP("file", "f", "", "Plan spec JSON file (\"-\" for stdin)")
+	cmd.Flags().StringP("release", "r", "", "Release name")
+	cmd.Flags().String("name", "", "Plan name")
+	cmd.Flags().String("description", "", "Plan description")
+	cmd.Flags().String("owner", "", "Owner username")
+	cmd.Flags().StringArray("participant", nil, "Participant username (repeatable)")
+	cmd.Flags().String("target-version", "", "Target version")
+	cmd.Flags().String("view-id", "", "Existing view UUID to attach (optional)")
+	cmd.Flags().StringArray("test", nil, "Test by build_system_id or group/test (repeatable)")
+	cmd.Flags().StringArray("group", nil, "Group name to expand into its enabled tests (repeatable)")
+	cmd.Flags().StringArray("assign", nil, "Assignment as entity=username (repeatable)")
+
+	parent.AddCommand(cmd)
+}
+
+// runCreate is the RunE handler for "planner create".
+func runCreate(cmd *cobra.Command, _ []string) error {
+	cmd.SilenceUsage = true
+	ctx := cmd.Context()
+	client := cmdctx.APIClientFrom(ctx)
+	out := cmdctx.OutputterFrom(ctx)
+	c := cmdctx.CacheFrom(ctx)
+	log := logging.For(cmdctx.LoggerFrom(ctx), "planner-create")
+
+	tmpl, err := loadTemplate(cmd)
+	if err != nil {
+		return err
+	}
+	if err := overlayFlags(cmd, &tmpl); err != nil {
+		return err
+	}
+	groups, _ := cmd.Flags().GetStringArray("group")
+	viewID, _ := cmd.Flags().GetString("view-id")
+
+	svc := services.NewPlannerService(client, c)
+
+	req, warnings, err := svc.BuildCreateRequest(ctx, tmpl, groups)
+	if err != nil {
+		log.Error().Err(err).Msg("failed to build create request")
+		return err
+	}
+	req.ViewID = viewID
+	for _, w := range warnings {
+		log.Warn().Msg(w)
+	}
+
+	plan, err := svc.CreatePlan(ctx, req)
+	if err != nil {
+		log.Error().Err(err).Msg("failed to create plan")
+		return err
+	}
+
+	log.Info().Str("plan_id", plan.ID).Str("key", plan.Key).Msg("plan created successfully")
+	return out.Write(models.NewKVTabular(plan))
+}
+
+// loadTemplate reads the --file plan spec (path or stdin) into a PlanTemplate.
+// Returns a zero template when --file is not set.
+func loadTemplate(cmd *cobra.Command) (models.PlanTemplate, error) {
+	var tmpl models.PlanTemplate
+	path, _ := cmd.Flags().GetString("file")
+	if path == "" {
+		return tmpl, nil
+	}
+
+	var raw []byte
+	var err error
+	if path == "-" {
+		raw, err = io.ReadAll(bufio.NewReader(os.Stdin))
+	} else {
+		raw, err = os.ReadFile(path)
+	}
+	if err != nil {
+		return tmpl, fmt.Errorf("reading plan file: %w", err)
+	}
+	if err := json.Unmarshal(raw, &tmpl); err != nil {
+		return tmpl, fmt.Errorf("parsing plan file: %w", err)
+	}
+	return tmpl, nil
+}
+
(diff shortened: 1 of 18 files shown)


soyacz:
A note: we talked today and decided to simplify some things and improve error handling. @k0machi let me know when it's ready for review after addressing

k0machi:
@soyacz ready for review.

soyacz:
@k0machi I still see 'participants' and 'tests'. I thought we agreed to drop these and put '$owner' for not assigned tests in `assignments' part. Also `groups` I think should have assignments (or actually it should be dropped and moved also to `assignments` part)

k0machi:
> @k0machi I still see 'participants' and 'tests'. I thought we agreed to drop these and put '$owner' for not assigned tests in `assignments' part. Also `groups`I think should have assignments (or actually it should be dropped and moved also to`assignments` part)

Which request/response did you see this in? This is a bug.

soyacz:
> > @k0machi I still see 'participants' and 'tests'. I thought we agreed to drop these and put '$owner' for not assigned tests in `assignments' part. Also `groups`I think should have assignments (or actually it should be dropped and moved also to`assignments` part)
> 
> Which request/response did you see this in? This is a bug.

./argus planner get --plan-id scylla-2026.2#1

k0machi:
> > > @k0machi I still see 'participants' and 'tests'. I thought we agreed to drop these and put '$owner' for not assigned tests in `assignments' part. Also `groups`I think should have assignments (or actually it should be dropped and moved also to`assignments` part)
> > 
> > 
> > Which request/response did you see this in? This is a bug.
> 
> ./argus planner get --plan-id scylla-2026.2#1

For now, also add `--template` to get the new format. I'll invert the flag into `--raw` instead and make the template default

Update: Added.

soyacz:
## Review: PR #1030 — cli/planner commands management

### Findings

**1. `cli/internal/services/planner.go:594` — ambiguous-group assignment silently dropped instead of aborting (correctness).**
`resolveAssignmentTargets` discards `ExpandGroup`'s error and always rewraps as `ErrEntityNotFound`:
```go
ids, gerr := s.ExpandGroup(ctx, releaseID, entityRef)
if gerr != nil {
    return nil, fmt.Errorf("%q is not a test or group in release: %w", entityRef, ErrEntityNotFound)
}
```
But `ExpandGroup→findGroup` returns `ErrAmbiguousEntity` when a group name/pretty-name matches 2+ groups. In `BuildCreateRequest` (planner.go:529) the `errors.Is(err, ErrEntityNotFound)` branch then **warn-and-omits** it. Scenario: two groups share a pretty-name; `--assign "Tier 1"=bob` → that group's tests are silently left out of the created plan and mislabeled `"is not in release"`, instead of aborting with a disambiguation list. Breaks the documented "ambiguous aborts" invariant that holds for ambiguous *tests* and group-removal. Untested (the ambiguous-create test exercises the bare-test path, not the group path). Fix: propagate `gerr` when it wraps `ErrAmbiguousEntity`.

**2. `cli/internal/services/planner.go:1025` — dead loop in `participantDeltas` (cleanup).**
```go
for _, t := range addTests.items {
    if user, ok := final[t]; ok && user != ownerID && !current[user] { addSet.add(user) }
}
for user := range desired { if !current[user] { addSet.add(user) } }
```
Every user added by the first loop is in `final`, hence already in `desired`, hence re-added by the second loop (and `addSet` dedups). First loop does nothing. Cost: maintainer reads 4 lines that look load-bearing but aren't. Delete it.

Minor (not surfaced as a finding): in `BuildCreateRequest`, conflicting assignments resolving to the same test UUID (`{"tier1":"bob", "tier1/t":"alice"}`) pick the assignee by map-iteration order — nondeterministic, but that's user-contradictory input.

redundancies:
`BuildResolvedPlans` (plural): zero production callers — only a test. `list` uses `BuildPlanSummaries`, `get`/`update` use `BuildResolvedPlan` (singular). Dead.

`planner.go:L779-790: delete: BuildResolvedPlans — no prod caller, every command uses BuildResolvedPlan or BuildPlanSummaries. Nothing replaces it (drop its test at planner_test.go:L819-835 too).`

`planner.go:L1025-1029: delete: first addTests loop in participantDeltas. Every user it adds is in final→desired, re-added by the desired loop below; addSet dedups. Nothing replaces it.`

k0machi:
@soyacz addressed the review and also tested the command outputs and flow - everything should now report and support the same format.

soyacz:
## Review: PR #1030 — cli/planner commands management

Adds `argus planner create/update/delete/overview/search` plus resolved/raw/template output modes to an existing bare-bones planner CLI, built on a new client-side name→UUID resolver (gridview-backed, with sentinel errors distinguishing safe-to-warn misses from must-abort ambiguity) and a diff-based update that only sends changed fields. Solid design overall; the bugs below are concentrated in the diff/consistency logic of `BuildUpdateRequest` and in edge cases of the shared resolver, not in the top-level architecture.

- **`cli/internal/services/planner.go:1070-1085` (`planLeftEmpty`) — the "plan must keep ≥1 test" guard doesn't see tests removed via group cascade.** `argus planner update --remove-group X` on a plan whose only tests all belong to group X leaves `diff.GroupsRemove` holding just the group UUID; `planLeftEmpty` only tracks `addTests`/`removeTests` deltas, so the client-side guard passes and the backend can be asked to leave the plan with zero tests, defeating the safety check that exists everywhere else in this function.

- **`cli/internal/services/planner.go:878-916` (`GroupsAdd`/`GroupsRemove` in `BuildUpdateRequest`) — hard-abort on any resolution error, unlike every sibling loop in the same function.** `TestsAdd`/`TestsRemove` warn-and-skip on `ErrEntityNotFound` (line ~865-900), but a typo'd `--add-group`/`--remove-group` name aborts the whole update instead. A user fixing five things in one `update` call loses all of them because one group name was misspelled, when the equivalent typo in `--remove-test` would just log a warning and proceed.

- **`cli/internal/services/planner.go:1020-1065` (`participantDeltas`) — renaming the owner via `--owner` spuriously adds the outgoing owner as a participant.** `final` is built from the plan's pre-existing `AssigneeMapping` (untouched entries carried forward) and then compared against the *new* `ownerID`. If alice is the current owner and has an existing explicit test assignment untouched by this update, `--owner bob` causes `alice != bob` to hold for that entry, so alice gets emitted in `ParticipantsAdd` even though she was never meant to become a participant — she's just the person losing ownership.

- **`cli/internal/services/planner.go:522-555` (`BuildCreateRequest` assignments loop) — overlapping group/test assignment entries pollute `Participants` with a user who loses the assignment race.** `participants.add(userID)` runs unconditionally for every `tmpl.Assignments` entry before the final `assignments[id]` map is settled. If a group is assigned to bob and one of its member tests is separately assigned to alice in the same template, both bob and alice land in `Participants`, but map iteration order nondeterministically decides which one actually ends up in `Assignments` for that test — the loser is listed as a participant on a plan they have no assignment on.

- **`cli/internal/services/planner.go:389-400` (`resolveTest`) — an ambiguous or errored group prefix is silently swallowed and misreported as "test not found."** The group-qualified branch (`if g, err := findGroup(grid, groupPart); err == nil { ... }`) only proceeds on success; any error, including `ErrAmbiguousEntity` for a genuinely ambiguous group name, falls through to the bare-name match against the *original unsplit* ref (e.g. `"tier1/mytest"`), which can never match. A caller then sees `no test "tier1/mytest" in this release` instead of the real "ambiguous group, disambiguate" condition — and through warn-and-skip callers (`resolveAssignmentTargets`, `resolveTestForDiff`) this silently drops a test from a plan rather than aborting for disambiguation.

- **`cli/cmd/planner/delete.go:60-67` — the plan summary is written to stdout unconditionally, even under `--yes`, before the deletion status object.** A fully non-interactive `delete --plan-id X --yes` writes two separate JSON documents to stdout in sequence (the summary, then `{"status":"deleted",...}`), breaking any script or CI step doing a single `json.Unmarshal` of the command's stdout.

- **`cli/cmd/planner/delete.go:90-101` (`confirm`) — closed/empty stdin is indistinguishable from an explicit "no."** `scanner.Scan()` returning `false` with no error (clean EOF, e.g. stdin redirected from `/dev/null` in a non-TTY context without `--yes`) takes the same path as a typed "n": it logs "deletion aborted by user" and returns `status: aborted`, misreporting an automation gap (forgot to pass `--yes`) as a deliberate user decision.

soyacz:
> @soyacz addressed the review and also tested the command outputs and flow - everything should now report and support the same format.

the outputs look good. See if the above AI review found something worth addressing.
When addressed, I think we can merge and release for broader testing (on new release maybe) and showcase tomorrow.

k0machi:
@soyacz addressed the relevant points and also added fixes from the demo.

soyacz (approved):

soyacz:
@k0machi let's move to the 'execution' part - first 'labeling' as we discussed

PR #1032, opened 2026-06-29
fix(cli/prometheus): print only address, drop progress chatter

soyacz:
## What

`argus prometheus start` printed step-by-step progress lines to stdout:

```
Looking for Prometheus data archive...
Extracting Prometheus TSDB data to /home/zeus/.cache/argus-cli/prometheus/<run-id>...
Starting Prometheus container...
Prometheus running at http://localhost:32772
Container: argus-prom-1300a473
```

## Why

Only the address is actionable. The cache-dir path in particular misled driving LLM agents into treating `.cache/argus-cli` as a working directory and hallucinating writes there.

## Change

- Drop all progress chatter; keep only `Prometheus running at http://localhost:PORT`.
- Failures still surface as errors (unchanged).
- The already-running branch now prints the same address line instead of a table exposing the internal container name.

`log.Info/Debug` (stderr, log-level gated) untouched; `list`/`stop` status lines untouched.

### DIFF
--- a/cli/cmd/prometheus.go
+++ b/cli/cmd/prometheus.go
@@ -107,14 +107,9 @@ var promStartCmd = &cobra.Command{
 		// Check if container already exists for this run
 		containerName := promContainerName(runID)
 		if containerExists(ctx, containerName) {
-			out := OutputterFrom(ctx)
 			port, _ := getContainerPort(ctx, containerName)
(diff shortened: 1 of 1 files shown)


soyacz:
@fruch this caused hallucination in perf tests investigation and caused failures

fruch (approved):

PR #1033, opened 2026-07-01
feature(cli/results): add cross-test generic results catalog and search

CodeLieutenant:
Adds argus results catalog and argus results query CLI commands backed by new /api/results/catalog and /api/results/search endpoints, enabling cross-test reads over generic result tables without a test_id up front. Uses a scatter-gather scan+fan-out over generic_result_metadata_v1's (test_id, name) pairs, mirroring the existing pytest result_filter pattern, and reuses the ALLOW FILTERING range-query path already used by _get_tables_data.

Closes https://scylladb.atlassian.net/browse/ARGUS-176

### DIFF
--- a/argus/backend/controller/results_api.py
+++ b/argus/backend/controller/results_api.py
@@ -1,0 +1,53 @@
+from datetime import datetime, timezone
+
+from flask import Blueprint, jsonify, request
+
+from argus.backend.error_handlers import handle_api_exception
+from argus.backend.service.results_service import ResultsService
+from argus.backend.service.user import api_login_required
+
+bp = Blueprint("results_api", __name__, url_prefix="/api/results")
+bp.register_error_handler(Exception, handle_api_exception)
+
+
+@bp.route("/catalog", methods=["GET"])
+@api_login_required
+def results_catalog():
+    """Cross-test catalog of every generic result table name, with per-name
+    test count and column metadata."""
+    service = ResultsService()
+    catalog = service.get_generic_result_catalog()
+
+    return jsonify({
+        "status": "ok",
+        "response": catalog,
+    })
+
+
+@bp.route("/search", methods=["GET"])
+@api_login_required
+def results_search():
+    """Cross-test flattened cell dump over generic result tables, filtered by
+    table name substring, status, and a sut_timestamp range."""
+    name = request.args.get("name")
+    statuses = request.args.getlist("status[]")
+    before = request.args.get("before")
+    after = request.args.get("after")
+    limit = int(request.args.get("limit", 500))
+
+    before_dt = datetime.fromtimestamp(int(before), tz=timezone.utc) if before else None
+    after_dt = datetime.fromtimestamp(int(after), tz=timezone.utc) if after else None
+
+    service = ResultsService()
+    result = service.search_generic_results(
+        name=name,
+        statuses=statuses or None,
+        before=before_dt,
+        after=after_dt,
+        limit=limit,
+    )
+
+    return jsonify({
+        "status": "ok",
+        "response": result,
+    })
--- a/argus/backend/service/results_service.py
+++ b/argus/backend/service/results_service.py
@@ -13,8 +13,13 @@ from argus.backend.db import ScyllaCluster
 from argus.backend.models.result import ArgusGenericResultMetadata, ArgusGenericResultData, ArgusBestResultData, ColumnMetadata, ArgusGraphView
 from argus.backend.plugins.sct.udt import PackageVersion
 from argus.backend.service.testrun import TestRunService
+from argus.backend.util.common import chunk, matches_substring
 
 LOGGER = logging.getLogger(__name__)
+
+# Number of (test_id, name) partitions to fan out to concurrently in
+# ResultsService.search_generic_results via execute_async.
+GENERIC_RESULTS_FANOUT_CHUNK_SIZE = 100
 
 type RunId = str
 type ReleasesMap = dict[str, list[RunId]]
@@ -428,12 +433,18 @@ class ResultsService:
         tables_meta = self.cluster.session.execute(query=query, parameters=(test_id,))
         return [ArgusGenericResultMetadata(**table) for table in tables_meta]
 
-    def _get_tables_data(self, test_id: UUID, table_name: str, ignored_runs: list[RunId],
-                         start_date: datetime | None = None, end_date: datetime | None = None) -> list[ArgusGenericResultData]:
+    @staticmethod
+    def _build_generic_result_data_query(test_id: UUID, table_name: str, start_date: datetime | None = None,
+                                          end_date: datetime | None = None) -> tuple[str, list]:
+        """Build the "cell dump" query/parameters for a single (test_id, name) partition.
+
+        Shared by [_get_tables_data] (single-test callers) and
+        [ResultsService.search_generic_results] (cross-test scatter-gather), so the
+        sut_timestamp range/ALLOW FILTERING logic lives in exactly one place.
+        """
         query_fields = ["run_id", "column", "row", "value", "status", "sut_timestamp"]
         raw_query = (f"SELECT {','.join(query_fields)}"
                      f" FROM generic_result_data_v1 WHERE test_id = ? AND name = ?")
-
         parameters = [test_id, table_name]
 
         if start_date:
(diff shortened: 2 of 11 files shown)


argus/backend/tests/results_service/test_cross_test_results.py
```diff
@@ -0,0 +1,188 @@
...
+def test_catalog_groups_same_named_table_across_tests(two_tests_same_table_name):
+    _, _, _, table_name, other_table_name = two_tests_same_table_name
+    service = ResultsService()
+    catalog = service.get_generic_result_catalog()
+
+    by_name = {entry["name"]: entry for entry in catalog}
+    assert by_name[table_name]["test_count"] == 2
+    column_names = {col["name"] for col in by_name[table_name]["columns"]}
+    assert column_names == {"p99", "p999"}
+    assert by_name[other_table_name]["test_count"] == 1
```
pehala, line 82: Can you assert the full on dict instead? I find it much more readable.

argus/backend/tests/results_service/test_cross_test_results.py
```diff
@@ -0,0 +1,188 @@
...
+    ).save()
+
+    ArgusGenericResultData(
+        test_id=fake_test.id, name=table_name, run_id=ignored_run_id, column="col1", row="row1",
+        sut_timestamp=start_time, value=1.0, status="UNSET",
+    ).save()
+    ArgusGenericResultData(
+        test_id=fake_test.id, name=table_name, run_id=kept_run_id, column="col1", row="row1",
+        sut_timestamp=start_time, value=2.0, status="UNSET",
+    ).save()
```
pehala, line 156: This should be in a setup

argus/backend/tests/results_service/test_cross_test_results.py
```diff
@@ -0,0 +1,188 @@
...
+
+    ArgusGenericResultData(
+        test_id=fake_test.id, name=table_name, run_id=uuid4(), column="col1", row="row1",
+        sut_timestamp=None, value=1.0, status="UNSET",
+    ).save()
+    ArgusGenericResultData(
+        test_id=fake_test.id, name=table_name, run_id=uuid4(), column="col1", row="row1",
+        sut_timestamp=datetime.now(UTC), value=2.0, status="UNSET",
+    ).save()
```
pehala, line 183: this should bein a setup

argus/backend/tests/results_service/test_cross_test_results.py
```diff
@@ -0,0 +1,188 @@
...
+        sut_timestamp=now - timedelta(days=5), value=10.0, status="UNSET",
+    ).save()
+    ArgusGenericResultData(
+        test_id=test_id_b, name=table_name, run_id=uuid4(), column="p99", row="run1",
+        sut_timestamp=now - timedelta(days=1), value=20.0, status="ERROR",
+    ).save()
+    ArgusGenericResultData(
+        test_id=test_id_c, name=other_table_name, run_id=uuid4(), column="ops", row="run1",
+        sut_timestamp=now - timedelta(days=10), value=1000.0, status="UNSET",
+    ).save()
```
pehala, line 68: nit: I would split it into multiple fixtures, per item, so you can actually request it individually in the tests and not having to do `_, _, _, table_name, other_table_name = two_tests_same_table_name`. You can also setup_cluster with autouse which will make sure they all created everytime

argus/backend/controller/results_api.py
```diff
@@ -0,0 +1,53 @@
...
+
+@bp.route("/catalog", methods=["GET"])
+@api_login_required
+def results_catalog():
+    """Cross-test catalog of every generic result table name, with per-name
+    test count and column metadata."""
+    service = ResultsService()
+    catalog = service.get_generic_result_catalog()
+
+    return jsonify({
```
pehala, line 21: @k0machi On one of my PRs you mentioned that jsonify is not needed. Does it apply here as well? COuld you explain when it is needed and when not? I think I still dont fully understand it

argus/backend/service/results_service.py
```diff
@@ -445,10 +456,104 @@ def _get_tables_data(self, test_id: UUID, table_name: str, ignored_runs: list[Ru
...
+        return raw_query, parameters
+
+    def _get_tables_data(self, test_id: UUID, table_name: str, ignored_runs: list[RunId],
+                         start_date: datetime | None = None, end_date: datetime | None = None) -> list[ArgusGenericResultData]:
+        raw_query, parameters = self._build_generic_result_data_query(test_id, table_name, start_date, end_date)
         query = self.cluster.prepare(raw_query)
         data = self.cluster.session.execute(query=query, parameters=tuple(parameters))
         return [ArgusGenericResultData(**cell) for cell in data if cell["run_id"] not in ignored_runs]
 
+    def get_generic_result_catalog(self) -> list[dict]:
```
pehala, line 468: So this method exists to only discovery which tables/fields to search in?

argus/backend/service/results_service.py
```diff
@@ -445,10 +456,104 @@ def _get_tables_data(self, test_id: UUID, table_name: str, ignored_runs: list[Ru
...
+                                before: datetime | None = None, after: datetime | None = None,
+                                limit: int = 500) -> dict:
+        """Cross-test "cell dump" search over generic_result_data_v1.
+
+        Mirrors the scatter-gather pattern used by PytestViewService.result_filter:
+        enumerate partitions with a scan, then fan out chunked async queries per
+        partition, then merge/sort/limit app-side. generic_result_data_v1 has a
+        composite partition key (test_id, name), so the fan-out keys on
+        (test_id, name) pairs rather than on test_id alone.
+        """
```
pehala, line 511: AFAIK This pattern is constant source of performance problems, would it be possible to use different paradigm, like cache tables?

argus/backend/controller/results_api.py
k0machi: Yes, it is not needed here as well. `jsonify` takes the dict and returns a Response object, this is useful when you want to alter that response's headers for example or do anything else with that response before returning it from the controller. If you are not touching the Response, there is no point to call jsonify, as just returning dict object from a flask view automatically does this.

argus/backend/service/results_service.py
k0machi: Agreed, this pattern should be avoided now.

argus/backend/controller/results_api.py
soyacz: @CodeLieutenant adjust and add this info to AGENTS.md

argus/backend/service/results_service.py
```diff
@@ -445,10 +456,104 @@ def _get_tables_data(self, test_id: UUID, table_name: str, ignored_runs: list[Ru
...
         data = self.cluster.session.execute(query=query, parameters=tuple(parameters))
         return [ArgusGenericResultData(**cell) for cell in data if cell["run_id"] not in ignored_runs]
 
+    def get_generic_result_catalog(self) -> list[dict]:
+        """Cross-test catalog of every generic result table name.
+
+        Scans the small generic_result_metadata_v1 table (partitioned by test_id
+        only) in full and groups rows by table name, since there is no other way
+        to discover "which tables exist across all tests" short of a name-keyed
+        index table (out of scope for v1 per the design doc).
```
soyacz, line 474: this kind of comments are useless, this also requires adjusting AGENTS.md

argus/backend/controller/results_api.py
soyacz, line 1: Good idea to create this file. Let's move all results api (`argus/backend/controller/api.py`) here (you can rename endpoints and adjust frontend to use new one). Don't change client-api endpoints.

argus/backend/controller/results_api.py
```diff
@@ -0,0 +1,53 @@
...
+        "status": "ok",
+        "response": catalog,
+    })
+
+
+@bp.route("/search", methods=["GET"])
+@api_login_required
+def results_search():
```
soyacz, line 29: I don't think it should be called 'search' - when I think of search I usually don't think of fetching data, but just telling where this data is.
I propose to rename it to `get_results` (in comparison to existing `test_results` that should be renamed to `get_test_results`)

argus/backend/service/results_service.py
```diff
@@ -445,10 +456,104 @@ def _get_tables_data(self, test_id: UUID, table_name: str, ignored_runs: list[Ru
...
+                        "value": cell["value"],
+                        "status": cell["status"],
+                        "sut_timestamp": cell["sut_timestamp"],
+                    })
+
+        if statuses:
+            cells = [cell for cell in cells if cell["status"] in statuses]
+
+        cells.sort(key=lambda cell: cell["sut_timestamp"] or datetime.min, reverse=True)
+        return {"total": len(cells), "cells": cells[:limit]}
```
soyacz, line 555: this response looks very bad. Not humans, nor LLM will be able to easily reason from them.

Need to group by table, -> subgroups `test_id` -> subgroups `run_id`. Stretch goal: along `test_id` show canonical name (like build_id). Test names often show feature (like CDC, LWT, topology changes) or test type (like performance, artifact) .
Propose a json format that is both token efficient and clear.

argus/backend/service/results_service.py
```diff
@@ -445,10 +456,104 @@ def _get_tables_data(self, test_id: UUID, table_name: str, ignored_runs: list[Ru
...
+                        continue
+                    cells.append({
+                        "test_id": str(test_id),
+                        "table": table_name,
+                        "run_id": str(cell["run_id"]),
+                        "column": cell["column"],
+                        "row": cell["row"],
+                        "value": cell["value"],
+                        "status": cell["status"],
+                        "sut_timestamp": cell["sut_timestamp"],
```
soyacz, line 548: besdes sut_timestamp. there's a `WRITETIME` - which is the time people think when filtering by time. sut_timestamp is datetime (usually just a date) of the Scylla build.

argus/backend/service/results_service.py
```diff
@@ -445,10 +456,104 @@ def _get_tables_data(self, test_id: UUID, table_name: str, ignored_runs: list[Ru
...
+        return [
+            {
+                "name": name,
+                "test_count": len(entry["test_ids"]),
+                "columns": list(entry["columns"].values()),
+            }
+            for name, entry in sorted(catalog.items())
+        ]
+
+    def search_generic_results(self, name: str | None = None, statuses: list[str] | None = None,
```
soyacz, line 501: I think we could also filter by columns (list), often when investigating we just need one metric (like p99, duration etc.).

argus/backend/service/results_service.py
soyacz: I don't see how we could use cache tables. Do you have something on your mind?

argus/backend/service/results_service.py
```diff
@@ -445,10 +456,104 @@ def _get_tables_data(self, test_id: UUID, table_name: str, ignored_runs: list[Ru
...
+                                before: datetime | None = None, after: datetime | None = None,
+                                limit: int = 500) -> dict:
+        """Cross-test "cell dump" search over generic_result_data_v1.
+
+        Mirrors the scatter-gather pattern used by PytestViewService.result_filter:
```
soyacz, line 506: let's focus on what this does and not how it does the things in 'docstrings'

cli/cmd/results.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		log.Info().Int("total", result.Total).Int("cells", len(result.Cells)).Msg("results query fetched successfully")
+		return out.Write(result)
+	},
+}
+
+func init() {
+	resultsQueryCmd.Flags().String("name", "", "Filter table names by substring")
+	resultsQueryCmd.Flags().Int("limit", 500, "Maximum number of cells to return")
```
soyacz, line 160: limit number of tables, not cells

soyacz (commented):
@CodeLieutenant We need to pause on this implementation and align on the design and API agreement first. Could you provide a brief specification detailing:
Expected endpoints, parameters, and example responses.
A high-level design addressing the performance implications we're seeing (specifically the full scans and fetching items one by one).

Additionally, this PR highlights a need to update our AGENTS.md. Our method docstrings should define clear contracts rather than just explaining the internal logic ("why and how").

Let's take this back to the drawing board so we can align on the architecture before rewriting the code.

PR #1034, opened 2026-07-01
overhaul(client/replay-log): synchronous lock-based writer, crash-safe pre-flight records

CodeLieutenant:
Replace the background-thread + queue writer with synchronous, lock-serialized writes so a failing write can never grow unbounded memory and close() can no longer drop in-flight requests. record() now writes a "pending" entry before the HTTP call and a "final" entry after, so a crash mid-call still leaves a replayable record. Switches from fdatasync to write+flush only, and unifies status-code handling around a strict 2xx check.

Also fixes several failure paths the old writer-thread absorbed silently but that were left unguarded on the synchronous path: unguarded open()/close() I/O errors, a pre-flight write whose serialization failure could block the real HTTP call, and an atexit registration that leaked a strong reference (and, if fixed naively, could unregister other instances' cleanup).

Closes https://scylladb.atlassian.net/browse/ARGUS-182


argus/client/replay_log.py
```diff
@@ -138,21 +143,32 @@ def __init__(
...
         self._test_type: str = test_type or "unknown"
-        self._queue: queue.Queue = queue.Queue()
-        self._closed = threading.Event()
-        self._enqueue_lock = threading.Lock()
-        self._writer_thread = threading.Thread(
-            target=self._writer_loop, daemon=True, name="argus-replay-log-writer",
-        )
-        self._writer_thread.start()
-        atexit.register(self.close)
+        self._cv = threading.Condition()
```
pehala: Why not use lock? It is much simpler

soyacz:
**Success gate: PR description says "strict 2xx" but code accepts 3xx.**

The description says this "unifies status-code handling around a strict 2xx check", but `replay_log.py:80` uses `response.ok` (status < 400). The docstring (`status < 400`) and `test_replay_record_uses_response_ok_gate` (302 → `success=True`) both confirm 3xx is deliberately accepted — the opposite of "strict 2xx". Old code was `199 < status_code < 300`.

Not a functional bug (code/docstring/test are internally consistent, and `requests` auto-follows redirects so a raw 3xx rarely reaches this gate), but the description contradicts the discriminator it describes. Either fix the wording, or — if strict 2xx was actually intended — switch to `199 < response.status_code < 300` and update the two `response.ok` tests.

soyacz (approved):
LGTM. Solid overhaul — synchronous lock-serialized writer, careful in-flight drain on close(), and the pending/final phase split are well-designed and thoroughly tested. One non-blocking nit left inline re: "strict 2xx" wording vs `response.ok`.

k0machi (approved):

argus/client/replay_log.py
```diff
@@ -215,35 +237,69 @@ def record(
...
-        ``rec.success`` based on the response. If the block raises, ``success``
-        stays ``False`` and ``error`` captures the exception message. The
-        record is enqueued on exit either way.
+        The record is written to disk immediately, before the caller's HTTP
+        call runs, so a crash/SIGKILL mid-call still leaves a (pending)
+        record to replay. The caller runs the HTTP call inside the ``with``
+        block and sets ``rec.success`` based on the response; if the block
+        raises, ``success`` stays ``False`` and ``error`` captures the
+        exception. The record is written again with the final outcome on
+        exit either way.
```
soyacz: I think we can drop requirement for 'crash/SIGKILL'. This should simplify the code.
Basically, if we have this kind of interruption the last pending Argus calls are our least problem.
Just saving after Argus request with all the data should be sufficient

soyacz: @k0machi @CodeLieutenant  WDYT?

pehala:
I am still not onboard with this design, I believe it can be made even simpler with much less complexity. I already discussed it with @CodeLieutenant and we seemed to reach consensus on how this could be done.

With that said, this is still an improvement of sorts, it just isnt in its final form, so feel free to merge

soyacz:
> I am still not onboard with this design, I believe it can be made even simpler with much less complexity. I already discussed it with @CodeLieutenant and we seemed to reach consensus on how this could be done.
> 
> With that said, this is still an improvement of sorts, it just isnt in its final form, so feel free to merge

No, better to apply all the simplifications here, no need to release anything we know we change in a day

argus/client/replay_log.py
CodeLieutenant: Removed

soyacz:
@CodeLieutenant still need to change commit msg to reflect what was done, and some minor things:
## Finding
he title/body say *"record() now writes a 'pending' entry before the HTTP call and a 'final' entry after, so a crash mid-call still leaves a replayable record."* No such pre-flight write exists anywhere in the diff — `record(...)` is called exactly once per `post()` (`argus/client/base.py:185-219`), never before `session.post()`.

Given no hard-kill of the calling thread and a process death (crash/OOM/SIGKILL) losing one in-flight request is acceptable — the actual gap this would cover doesn't matter for how this is used. So this isn't a functional bug to fix, it's a **stale/inaccurate commit message**: reword it to describe what actually shipped (synchronous bounded-memory writer, correct atexit lifecycle, no more dropped-on-close records) and drop the crash-mid-call claim, since it isn't implemented and isn't needed.

### Other findings

- **`_evaluate_response` loosens the status gate, description says the opposite** (`argus/client/base.py:21-37`). Old: strict `199 < status_code < 300`. New: `response.ok`, true for any status `< 400` (3xx now passes the gate before the JSON-body check). PR body says this "unifies status-code handling around a strict 2xx check" — it's the reverse. Low real-world impact, but worth either restoring the strict 2xx check or fixing the wording so the description matches the code.

- **Dead attribute**: `ReplayLogOnlyResponse.ok = True` (`replay_log.py:211`) isn't read anywhere in this diff — the `replay_log_only` branch calls `record(success=False)` directly, bypassing `_evaluate_response`. Harmless; confirm no external caller depends on `.ok` here before treating it as intentional.

k0machi (approved):

PR #1036, opened 2026-07-06
backend/planner labels

k0machi:
- **feature(service/planner): add per-entity options with labels to release plans**
  Introduce an options object on ArgusReleasePlan stored as serialized JSON
  in a text column, keyed by test or group UUID. Initially carries a labels
  array per entity. Threaded through create/update (diff-based set/remove with
  pruning) and copy (remap onto copied entities).
  

- **feature(cli/planner): wire per-entity labels into planner commands**
  Extend the assignments map value to an object carrying an assignee plus
  per-entity options (labels), while still accepting the bare-string assignee
  form for backward compatibility. Labels flow through get/create/update: they
  are back-resolved in templates, emitted as options on create, and applied as
  merge deltas on update via new --label/--unlabel flags, computing the backend
  options_set/options_remove diff. Labels may attach to unassigned ($owner)
  entities. --assign behaviour is unchanged.
  

Task: ARGUS-183

### DIFF
--- a/argus/backend/service/planner_service.py
+++ b/argus/backend/service/planner_service.py
@@ -22,6 +22,15 @@ from argus.backend.util.common import chunk
 LOGGER = logging.getLogger(__name__)
 
 
+class EntityOptions(TypedDict, total=False):
+    """Per-entity options attached to a plan, keyed by test/group UUID."""
+    labels: list[str]
+
+
+# Mapping of entity UUID (test or group) -> its options
+PlanOptions = dict[str, EntityOptions]
+
+
 @dataclass(frozen=True, init=True, repr=True, kw_only=True)
 class CreatePlanPayload:
     name: str
@@ -33,7 +42,7 @@ class CreatePlanPayload:
     tests: list[str]
     groups: list[str]
     assignments: dict[str, str]
-    options: dict[str, Any] = field(default_factory=dict)
+    options: PlanOptions = field(default_factory=dict)
     view_id: Optional[str] = None
     created_from: Optional[str] = None
 
@@ -57,7 +66,7 @@ class TempPlanPayload:
     last_updated: str
     ends_at: str
     created_from: Optional[str]
-    options: dict[str, Any] = None
+    options: PlanOptions = None
     view_id: Optional[str] = None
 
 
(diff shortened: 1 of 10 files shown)


argus/backend/service/planner_service.py
```diff
@@ -33,6 +33,7 @@ class CreatePlanPayload:
     tests: list[str]
     groups: list[str]
     assignments: dict[str, str]
+    options: dict[str, Any] = field(default_factory=dict)
```
soyacz: **Type the option bag: `dict[str, Any]` → `dict[str, EntityOptions]`**

The inner value has a known shape and mirrors the Go `EntityOptions`. A `TypedDict` documents it at zero runtime cost and keeps the opaque passthrough (`**payload`, `json.dumps`, `current_options[key] = value` all still work). Already the convention here (`views.py`, `tunnel_service.py`) and already imported.

```python
class EntityOptions(TypedDict):
    labels: list[str]
```

- `CreatePlanPayload.options: dict[str, EntityOptions]`
- `PlanDiffPayload.options_set: dict[str, EntityOptions]`

Check-time only (no runtime validation); no methods needed since diffing lives in the CLI.

soyacz (commented):
**`TempPlanPayload.options` is unused** — `copy_plan` reads `original_plan.options` from the DB, never the payload. Drop the field or comment why it exists.

**Backend coverage gap** — no test covers adding a test via `tests_add` *and* setting its options in the same update. That's the case where pruning (`all_entity_ids` computed after the add) must retain the new entry. Code is correct; a test would lock it.

**No runtime validation on `options`** — `create_plan`/`update_plan` `json.dumps` arbitrary payload dicts straight in. Low risk (authenticated, internal), but there's no cap on size/shape. Fine to accept, noting it.

**Malformed-options handling is asymmetric** — `BuildTemplate` errors, `BuildResolvedPlan` silently swallows. Both defensible (round-trippable vs display-only), just make the intent explicit.

**Nits**
- `parseAssignments` (create) vs `parseAssignStrings` (update) are near-twins — a one-line note on why they diverge would help.
- Labels show only in JSON output, not the `ResolvedPlan` text table.
- Deploy note: new `options` column needs a schema sync; existing rows are `None` and all read paths are null-safe, so no data migration.

Overall: clean, well-tested, well-documented; nothing blocking. The EntityOptions typing and the dead-field cleanup are the only two I'd nudge before merge.

k0machi:
> TempPlanPayload.options is unused — copy_plan reads original_plan.options from the DB, never the payload. Drop the field or comment why it exists.

Much like other unused fields on that dict, it exists because the plan gets sent wholesale to backend to that endpoint.

> Backend coverage gap — no test covers adding a test via tests_add and setting its options in the same update. That's the case where pruning (all_entity_ids computed after the add) must retain the new entry. Code is correct; a test would lock it.

I added some basic tests to cover for now until I merge the big test PR which covers other half.

soyacz (approved):
LGTM

PR #1038, opened 2026-07-09
cli/jenkins service commands

k0machi:
- **feature(cli/test): add test execute and params commands**
  Add a 'test' command group wiring the Jenkins test-execution endpoints into
  the Go CLI: 'test params' fetches a job's parameter set (values map by
  default, --full for details) and 'test execute' triggers a build with the
  merge cascade defaults < --file < --param, plus --dry-run and --wait.
  

- **feature(cli/test): resolve tests by --release and --test**
  Let `test params` and `test execute` address a test by release name plus a
  test reference, resolved to a build_system_id via the planner gridview, as an
  alternative to the existing --build-id. The two addressing modes are mutually
  exclusive; --release and --test are required together.
  
  Resolution reuses the planner's name-resolution machinery (and its disk cache)
  through a new PlannerService.ResolveTestBuildID, which returns the matched
  test's build_system_id and rejects group references since a build targets a
  single job.
  

- **feature(cli/test): execute plan tests by label**
  Add a third addressing mode to `test execute`: --plan-id with one or
  more --label flags fans out and triggers a Jenkins build for every
  enabled test in the plan carrying the given label(s). Labels match as a
  union by default; --match-all switches to intersection. Matching is
  case-insensitive, labeled groups expand to their member tests, and
  results are deduplicated by build_system_id.
  
  Each build is triggered with its own fetched defaults plus any
  repeatable --param overrides; --file and --build-number are rejected in
  plan mode. --wait waits for all builds in parallel under a single
  batch-wide --wait-timeout, and per-test failures are recorded per row
  without aborting the batch.
  

- **feature(cli/test): add check-queue command**
  Add `test check-queue` to report the current Jenkins queue/build status
  of one or more queue items. Items are supplied with repeatable --item
  flags and/or a --file (path or "-" for stdin) carrying the JSON emitted
  by `test execute`, accepting either {"queueItem": 123} or
  {"queueItem": [123, 456]}.
  
  Items are merged, de-duplicated, and sorted; each is queried once via the
  existing queue_info endpoint and rendered as one row (JSON array or text
  table) reporting started/pending/error. A per-item failure is recorded
  in its row without aborting the batch. This is a one-shot snapshot with
  no waiting.
  

- **feature(service/testrun): add run lookup by build id and number**
  Add TestRunService.get_run_by_build_number plus HTML and JSON routes to
  resolve a run from its build_system_id and Jenkins build number. This
  yields a stable, clickable Argus link that resolves at click-time even
  before the run reports its id back to Argus.
  

- **feature(cli/test): show Argus run link on --wait**
  Print the Argus run URL for started builds when using --wait, for both
  single execute and plan fan-out. Poll interval is bumped from 2s to 3s.
  

- **feature(service/jenkins): guess next build number on trigger**
  Add JenkinsService.next_build_number and let the build endpoint return it
  on demand (includeBuildNumber), so callers can derive a stable Argus run
  link right after triggering, before the build leaves the queue. Resolving
  the number is best-effort and never fails the trigger.
  

- **feature(cli/test): show guessed Argus run link on non-wait execute**
  A single 'test execute' without --wait now prints the queue item plus the
  guessed build number and Argus run URL, obtained by opting into the
  backend's next-build-number guess. The link resolves at click-time, so a
  rare wrong guess just 404s briefly. Fan-out output is unchanged.

### DIFF
--- a/argus/backend/controller/api.py
+++ b/argus/backend/controller/api.py
@@ -7,7 +7,8 @@ from flask import (
     current_app,
     g,
     redirect,
-    request, Response
+    request, Response,
+    url_for,
 )
 from flask.json import jsonify
 from argus.backend.error_handlers import handle_api_exception
@@ -44,6 +45,24 @@ def app_version():
         "status": "ok",
         "response": {
             "commit_id": argus_version
+        }
+    })
+
+
+@bp.route("/test/<path:build_id>/<int:build_number>")
+@api_login_required
+def get_run_by_build(build_id: str, build_number: int):
+    # JSON sibling of main.get_run_by_build: resolve a run from its
+    # build_system_id + Jenkins build number and return its id and Argus URL.
+    run = TestRunService().get_run_by_build_number(build_id, build_number)
+    if not run:
+        raise Exception(f"Run not found for {build_id} #{build_number}")
+    return jsonify({
+        "status": "ok",
+        "response": {
+            "run_id": str(run.id),
+            "plugin_name": run._plugin_name,
+            "url": url_for("main.get_run_by_plugin", plugin_name=run._plugin_name, run_id=run.id, _external=True),
         }
     })
 
--- a/argus/backend/controller/main.py
+++ b/argus/backend/controller/main.py
@@ -58,6 +58,21 @@ def get_run_by_plugin(plugin_name: str, run_id: UUID | str, tab: str):
     return render_template("run_view_by_plugin.html.j2", run=run, tab=tab)
 
 
+@bp.route("/test/<path:build_id>/<int:build_number>", defaults={"tab": "details"})
+@bp.route("/test/<path:build_id>/<int:build_number>/<string:tab>")
+@login_required
+def get_run_by_build(build_id: str, build_number: int, tab: str):
+    # Resolve a run from its build_system_id + Jenkins build number. This gives
+    # a stable, clickable link that can be produced the moment a build starts —
+    # even during the brief window before its run_id exists — and resolves once
+    # the run has been reported to Argus.
+    run = TestRunService().get_run_by_build_number(build_id, build_number)
+    if not run:
+        flash(f"Run {build_id} #{build_number} not found.", "error")
+        return redirect(url_for("main.error", type=404))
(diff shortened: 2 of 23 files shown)


soyacz:
### Issues / risks (all minor)

1. **TOCTOU on the guessed build number** (`jenkins_service.py`, `next_build_number`): it reads Jenkins' `nextBuildNumber` *after* the trigger call returns. If another build of the same job is queued/started concurrently in that window, the guess can point at the wrong build. This is inherent to the "best-effort guess" design and is documented as such — just confirm the CLI always presents it as a guess (not authoritative) until it resolves via `get_run_by_build_number`, which appears to be the case.

2. **Non-deterministic pick on duplicate build numbers** (`testrun.py:94-95`, `get_run_by_build_number`): `runs[0] if runs else None` — under `ALLOW FILTERING` with no clustering-key match, row order isn't guaranteed. If two runs ever share the same `(build_id, build_number)` (e.g. a duplicate submission), the resolved run is arbitrary. Edge case, likely fine in practice.

3. **No URL-escaping of `build_system_id` in `argusRunURL`** (`cli/cmd/test/execute.go:337-341`): the build ID is concatenated raw into the URL path. Given job names in this system are always simple identifiers (`scylla-2026.2/longevity/longevity-100gb`), this is low risk today but would break if a job name ever contains characters needing escaping.

4. **Redundant double timeout** (`waitAllBuilds`): passes `timeout` explicitly into `WaitForBuild` even though the parent context already has `context.WithTimeout(ctx, timeout)` applied — harmless belt-and-suspenders, not a bug.

Nothing here blocks merging — findings 1–2 are pre-existing/accepted tradeoffs given the docstrings, and 3–4 are nits.


@k0machi tomorrow I'll try looking at it. If I won't make it, you can merge (I'll create followup task if I find something worth changing)

k0machi:
> TOCTOU on the guessed build number 

Expected, I know about this and the functionality is for a single job start, we don't really expect 10 people to fire off same staging job in same folder.

> Non-deterministic pick on duplicate build numbers

Kinda okay, since it will return all 3 in the view anyway.

> No URL-escaping of build_system_id

Not needed, those parts are already used in jenkins and escaped there.

PR #1040, opened 2026-07-13
feature(deploy): rotate and compress argus logs weekly

soyacz:
## Summary
- uwsgi previously logged only to journald, which was retaining prod logs for just 1 day
- `uwsgi.ini` now logs to `/var/log/argus/argus.log`
- Added `docs/config/argus.logrotate` (daily rotation, keep 7, compressed) alongside the existing `argus.service`/`argus.nginx.conf` deploy docs
- README updated with the steps to create the log directory and install the logrotate config on a new machine

## Test plan
- [ ] Deploy to a test host, confirm uwsgi writes to `/var/log/argus/argus.log`
- [ ] Run `logrotate -d /etc/logrotate.d/argus` to verify the config parses and rotation/compression rules apply as expected

### DIFF
--- a/docs/config/argus.logrotate
+++ b/docs/config/argus.logrotate
@@ -1,0 +1,9 @@
+/var/log/argus/argus.log {
+    daily
+    rotate 7
+    compress
+    delaycompress
+    missingok
+    notifempty
+    copytruncate
+}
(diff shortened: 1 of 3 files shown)


CodeLieutenant (approved):

docs/config/argus.logrotate
```diff
@@ -0,0 +1,9 @@
+/var/log/argus/argus.log {
+    daily
+    rotate 7
+    compress
```
CodeLieutenant, line 4: Maybe we can use something from zeus, there we do logrotate with zstd compression

soyacz: I'll try with that and see if we need something more later

PR #1041, opened 2026-07-14
fix(backend/replay): reconstruct events, logs and terminal state on ingest

CodeLieutenant:
Replaying a run's request-log left the run visibly incomplete: events were empty, collected logs were missing, and generic (dtest/pytest) runs never reached a terminal status. Make ingest faithfully rebuild a run, and stay idempotent when the same archive is uploaded more than once.

Logs:
- Add a default-on S3 log back-fill: after replaying, list each run's S3 prefix and submit links for archives that reached S3 but whose logs/submit was never recorded (loader/monitor/sct-runner bundles). Re-uses the app's read-only S3 creds and the link-only log model (no upload/hosting) and is idempotent. Exposed on the ingest endpoint (?backfill_logs=false to disable) and the CLI (--backfill-logs).

Terminal state:
- Stop skipping /testrun/$type/$id/finalize on replay. It has no external side effect and is the only source of terminal status + scylla_version + end_time for the generic and driver-matrix plugins, so skipping it left replayed runs stuck unfinished. Run it in the terminal ordering group, after every middle record.
- Preserve the run's real finish time: finish_run now honours an optional end_time in the finalize body, and ingest injects it from the record ts (milliseconds -> seconds) so end_time reflects the original run, not the replay moment. Live finalize calls are unaffected.

Idempotency:
- finalize_nemesis is a no-op once a nemesis has left RUNNING, so re-uploading an archive no longer double-counts nemesis_stats or drifts end_time/duration.
- Tolerate a status-less generic finalize instead of raising.

Closes https://scylladb.atlassian.net/browse/ARGUS-186

### DIFF
--- a/argus/backend/service/replay_service.py
+++ b/argus/backend/service/replay_service.py
@@ -23,18 +23,33 @@ exactly as they would for a live client.
 
 Two side concerns remain server-side:
 
-* **Ordering** -- ``submit_run`` must run before anything referencing the run,
-  terminal ``set_status`` runs last so ``end_time`` is back-filled, and
-  heartbeats collapse to the most recent record (per run).
-* **Skip-list** -- endpoints with side effects outside Argus
+* **Ordering** -- ``submit_run`` must run before anything referencing the run;
+  terminal ``set_status`` and ``finalize`` run last; and heartbeats collapse to
+  the most recent record (per run). ``finalize`` is the only call that
+  back-fills ``end_time`` -- and, for the generic (pytest/dtest) and
+  driver-matrix plugins, the terminal ``status`` and ``scylla_version`` too --
+  so it must land after every middle record.
+* **Skip-list** -- only endpoints with side effects *outside* Argus
   (``/testrun/report/email``, ``/planning/plan/trigger``) are skipped so
-  replay doesn't send emails or trigger CI jobs.
+  replay doesn't send emails or trigger CI jobs. ``finalize`` is deliberately
+  *not* skipped: it only writes to the run row and is the single source of
+  terminal status/version for non-SCT plugins, so skipping it silently left
+  replayed dtest/pytest runs stuck without a final status.
 
 When ``create_missing_tests`` is set, ``submit_run`` records get a pre-step
 that ensures the ``ArgusRelease/Group/Test`` triple exists for the run's
 build_id, so that the controller's ``assign_categories`` can populate
 ``test_id`` on the row. This is opt-in because curated instances already have
 the hierarchy populated and don't want orphan auto-creates.
+
+When ``backfill_logs`` is set, a post-step lists each run's S3 log prefix
+(derived from the run-id in any recorded log link) and submits links for any
+archive that reached S3 but whose ``logs/submit`` was never recorded -- e.g.
+loader/monitor/sct-runner bundles collected at teardown. The ingest endpoint
+enables this by default (disable with ``?backfill_logs=false``); it re-uses
+the app's existing read-only S3 credentials and the link-only log model
+(nothing is uploaded or hosted), and is idempotent, so the extra S3 list per
+ingest is the only cost when there is nothing to back-fill.
 """
 import gzip
 import io
@@ -59,9 +74,6 @@ LOGGER = logging.getLogger(__name__)
 SKIP_ENDPOINTS: frozenset[str] = frozenset({
     "/testrun/report/email",       # would send actual mail
     "/planning/plan/trigger",      # would re-trigger CI jobs
-    # ``finalize`` is folded into terminal ``set_status`` by the new client
-    # flow; older logs that still emit it are no-ops.
-    "/testrun/$type/$id/finalize",
 })
 
 # Set of ``set_status`` body.new_status values that we treat as terminal.
@@ -74,6 +86,33 @@ TERMINAL_STATUSES = frozenset({
 # ``/testrun/$type/submit``); we prepend this when reconstructing the URL
 # for the Flask test client.
 CLIENT_ROUTE_PREFIX = "/api/v1/client"
+
+# The ``logs/submit`` endpoint template, referenced from both ordering and the
+# S3 log back-fill.
+LOGS_SUBMIT_ENDPOINT = "/testrun/$type/$id/logs/submit"
+
+# The ``finalize`` endpoint template. It writes the run's terminal state
+# (``end_time`` for every plugin; ``status`` and ``scylla_version`` for the
+# generic/driver-matrix plugins), so ordering runs it in the terminal group.
+FINALIZE_ENDPOINT = "/testrun/$type/$id/finalize"
+
+# SCT's ``S3Storage.bucket_name`` default (sdcm/utils/common.py) -- every SCT
+# log collector (``LogCollector.collect_logs``, ``collect_sct_logs``,
+# ``upload_system_table_to_s3``) uploads here unless a run overrides it, so
+# it's the right fallback when a run recorded no S3 links to derive its
+# bucket from.
+_DEFAULT_LOG_BACKFILL_BUCKET = "cloudius-jenkins-test"
+
+# Parse ``bucket`` and ``key`` out of a virtual-hosted S3 URL, e.g.
+# ``https://cloudius-jenkins-test.s3.amazonaws.com/<run_id>/<ts>/<name>``.
+# Mirrors ``TestRunService._match_s3_link`` so the URLs we synthesise line up
+# byte-for-byte with the ones the client records.
+_S3_LINK_RE = re.compile(
+    r"(https:\/\/)?(?P<bucket>[\w\-]+)\.s3(?P<region>\.[\w\-\d]*)?\.amazonaws\.com\/(?P<key>.+)"
+)
+
+# Suffixes stripped from an S3 key's basename to derive a human log name.
+_LOG_NAME_SUFFIXES = (".tar.zst", ".tar.gz", ".tar.zstd", ".tgz", ".tar", ".zip")
 
 
 @dataclass
@@ -91,6 +130,9 @@ class ReplaySummary:
     succeeded: int = 0
     failed: int = 0
     skipped_no_replay: int = 0
+    # Number of log links discovered in S3 and back-filled onto runs (i.e. logs
+    # that were uploaded but whose ``logs/submit`` call was never recorded).
+    backfilled_logs: int = 0
     errors: list[dict] = field(default_factory=list)
 
     def as_dict(self) -> dict:
@@ -100,6 +142,7 @@ class ReplaySummary:
             "succeeded": self.succeeded,
             "failed": self.failed,
             "skipped_no_replay": self.skipped_no_replay,
+            "backfilled_logs": self.backfilled_logs,
             "errors": self.errors,
         }
 
@@ -130,10 +173,15 @@ class ReplayService:
         app=None,
         auth_header: str | None = None,
         create_missing_tests: bool = False,
+        backfill_logs: bool = False,
+        s3_client=None,
     ) -> None:
         self._app = app
         self._auth_header = auth_header
         self._create_missing_tests = create_missing_tests
+        self._backfill_logs = backfill_logs
+        # Lazily built from app config on first use; injectable for tests.
+        self._s3 = s3_client
 
     # ------------------------------------------------------------------
     # Entry point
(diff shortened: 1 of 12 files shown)


argus/backend/service/replay_service.py
```diff
@@ -478,6 +539,144 @@ def _ensure_hierarchy_for_submit_run(record: dict) -> None:
...
+        so it never has to match a client-chosen name.
+        """
+        name = key.rsplit("/", 1)[-1]
+        for suffix in _LOG_NAME_SUFFIXES:
+            if name.endswith(suffix):
+                return name[: -len(suffix)]
+        return name
+
+    @classmethod
+    def _resolve_run_bucket(cls, recorded_links: set[str]) -> str | None:
```
soyacz, line 606: sct already has mechanism for logs upload to Argus - did you verify with it - we could reuse logic from there if possible.

argus/backend/service/replay_service.py
```diff
@@ -544,6 +743,16 @@ def _process_one(
         body = record.get("body")
         method = (record.get("method") or "POST").upper()
 
+        # Preserve the run's original finish time on replay. ``ts`` is the
```
soyacz, line 776: we usually not have this one, what then?

argus/backend/service/replay_service.py
```diff
@@ -308,14 +364,19 @@ def _apply_ordering(records: list[dict]) -> list[dict]:
                 submit_runs.append(r)
             elif endpoint == "/testrun/$type/$id/set_status" and ReplayService._is_terminal_status(r):
                 terminal_statuses.append(r)
+            elif endpoint == FINALIZE_ENDPOINT:
```
soyacz, line 375: we'll usually not have this is replay log - Argus is finalized after logs collection - so we can't have it in logs.

argus/backend/service/replay_service.py
CodeLieutenant: Set as default `now` timestamp, is that acceptable?

argus/backend/service/replay_service.py
CodeLieutenant: Problem is when you dont have argus you cannot link them, this is the extracted way to link those s3 files to argus

argus/backend/service/replay_service.py
CodeLieutenant: lets have here so that we dont crash if it exists

argus/backend/service/replay_service.py
soyacz: I'd rather to use the last seen timestamp. Recovering from the failure may take days and we'll have days-long tests

argus/backend/service/replay_service.py
soyacz: I meant that we have algorithm to get logs from s3 in SCT (without argus) and actually, you can send logs to argus (when you have argus run id).
I was thinking to reuse logic from there in discovering logs, so we don't miss anything (as things in SCT are usually more complex) - just verify if your solution is complete.

argus/backend/service/replay_service.py
CodeLieutenant: Ok thats good one

argus/backend/service/replay_service.py
CodeLieutenant: But how to reuse the solution from SCT when we use argus-cli to post this stuff to argus after argus either recovered or we just want data in argus. only way i see is that SCT appends the log files as special line in the replay log and then we treat it like that?

What do you think?

soyacz: I meant to see what SCT code does for finding logs, and apply the same logic here.
SCT won't append these logs as collect logs phase and pushing to argus is after we already collected replay log.

argus/backend/service/replay_service.py
CodeLieutenant: Done

argus/backend/service/replay_service.py
CodeLieutenant: Done

soyacz (approved):
LGTM

k0machi (approved):

k0machi:
@CodeLieutenant conflicts here.

CodeLieutenant:
> @CodeLieutenant conflicts here.

Resolved

soyacz:
AI review:

## What this PR does

**Problem** — Replaying a run's request-log left the run visibly incomplete: events were empty, collected logs were missing, and generic (dtest/pytest) runs never reached a terminal status. Re-uploading the same archive could also double-count nemesis stats.

**Approach** — Stop unconditionally skipping `/testrun/$type/$id/finalize` during replay (it was treated as folded into terminal `set_status`); dispatch it in the terminal ordering group instead, since it's the only source of terminal status/scylla_version/end_time for the generic and driver-matrix plugins. `finish_run` on the driver-matrix/generic/sct models now accepts a caller-supplied `end_time` (epoch seconds), which ingest injects from the finalize record's own `ts`, falling back to the highest `ts` seen anywhere in the run's replay log when the finalize record itself has none. A new default-on S3 log-backfill lists each run's S3 prefix after replay and submits `logs/submit` records for any log objects that reached S3 but were never recorded (e.g. loader/monitor/sct-runner bundles lost to a mid-run crash), resolving the bucket from a recorded link, then config, then a hardcoded SCT default. `finalize_nemesis` becomes a no-op once a nemesis has left `RUNNING`, and generic finalize tolerates a missing `status`, for idempotent re-ingest.

**Risk tier** — High (per review directive): touches the production ingest path for test-run data, reuses the app's AWS credentials for a new S3-listing capability, and underpins idempotency guarantees for re-ingest.

**Ticket** — ARGUS-186: the linked ticket describes and was closed by a *client-side* fix in a different repo (scylladb/scylla-cluster-tests#15418) that wires the real-time event pipeline to the replay-only client so events actually land in the replay log. This PR is the backend counterpart — making ingest faithfully replay what that client-side fix now records (events, logs, terminal state) — so it's a companion fix, not scope creep, though the PR itself closes an already-Done ticket from a different repo's PR.

## Verdict

Needs changes — the core finalize/end-time-ordering fix is sound and well-tested, but the new S3 log-backfill subsystem has an unguarded bucket parameter reachable from untrusted archive content. 1 critical, 4 major, 6 minor.

## Findings

### 🔴 Critical
- **`argus/backend/service/replay_service.py:606-622`** (`_resolve_run_bucket`) — The S3 bucket used for `list_objects_v2` (line 680) is regex-extracted from a **recorded log link inside the replay archive** — i.e., attacker-controlled input, since any authenticated user can POST an ingest archive (`@api_login_required` only, no ownership/role check — `argus/backend/controller/client_api.py`) containing a forged `logs/submit` record with `log_link: "https://<attacker-bucket>.s3.amazonaws.com/..."`. That string is handed straight to `s3.list_objects_v2(Bucket=bucket, ...)` using the app's own AWS credentials, with no allowlist check. The codebase already has the right pattern for this — `S3_ALLOWED_BUCKETS` in `argus/backend/service/testrun.py:258` gates `proxy_s3_file` — but it isn't applied here. The listed keys are then persisted as new `log_link`s and dispatched through `logs/submit`, so a crafted archive can make the server enumerate and then serve (via the existing `get_log` presign) objects from an arbitrary bucket the app's credentials can reach. Fix: validate the derived bucket against `S3_ALLOWED_BUCKETS` (or a dedicated backfill allowlist) before calling `list_objects_v2`, and skip when it doesn't match.

### 🟠 Major
- **`argus/backend/service/replay_service.py:588-589, 606-622, 645-670`** — `_s3_url` always synthesizes a non-region-qualified host (`{bucket}.s3.amazonaws.com`), while `_S3_LINK_RE` and `_resolve_run_bucket` explicitly parse-and-discard an optional region from *recorded* links (SCT log collectors run across many regions per this PR's own comment). For a run whose original client recorded a region-qualified link, the backfill's `url not in recorded_links` dedup check never matches (regioned vs non-regioned host strings differ), and the name-based dedup in `submit_logs` (derived name vs. client-chosen name) won't catch it either — so a duplicate log entry gets created on every re-ingest, defeating the PR's own idempotency claim. No test in the diff uses a region-qualified recorded link. Fix: preserve the recorded link's actual host when re-deriving the URL (store the host, not just the bucket), or normalize both sides before comparing; add a regioned-link test.
- **`argus/backend/service/replay_service.py:681-684` and the call site at `:205`** — The "S3 outage ≠ ingest failure" isolation (`except Exception` around `_list_s3_run_objects`) wraps only the listing call, not `_resolve_run_bucket` or the surrounding per-run body, and `ingest()`'s call to `_build_log_backfill_records(records)` (line 205) has no try/except at all. An exception anywhere else in the backfill path fails the *entire* ingest, the opposite of the stated intent. Compounding this: failures are only ever `LOGGER.exception`'d — no `summary.errors` entry, no counter, no metric (the app has Prometheus instrumentation in `argus/backend/metrics.py` that nothing here reports to). If S3 creds/policy/network break, every future ingest silently no-ops the backfill forever, and the response still reads `"status": "ok", "backfilled_logs": 0` — indistinguishable from "nothing to back-fill." Fix: widen the try/except to cover the whole per-run backfill body (or the call site), and add a `backfill_errors` counter to `ReplaySummary`/response so a systemic failure is visible to the caller, not just server logs.
- **`argus/backend/service/replay_service.py:578-584, 624-643`** — The boto3 S3 client is built with no explicit connect/read timeouts or bounded retries (botocore defaults ~60s), and `_list_s3_run_objects`'s paginated listing runs synchronously in the Flask request thread, once per distinct run touched by the archive, with no cap on run count or overall time budget. This is the first place this client-construction pattern (copied from `testrun.py`, previously only used for local presigning) makes an actual blocking network call. A degraded/unreachable S3 endpoint can stall a request worker for minutes, multiplied by however many runs are in the archive — on by default. Fix: pass explicit timeouts/retry config to the S3 client and/or cap backfill work per ingest call.
- **`argus/backend/plugins/driver_matrix_tests/model.py:454-456`, `argus/backend/plugins/generic/model.py:84-88`, `argus/backend/plugins/sct/testrun.py:111-113`** — No test in this PR exercises the real `datetime.utcfromtimestamp`/`datetime.fromtimestamp(_, UTC)` conversions against an actual persisted run; every `test_replay_service.py` assertion stops at "the right JSON was sent to a mock HTTP client." A wrong-unit bug in any one plugin's conversion (e.g. missing the `/1000`, or a `fromtimestamp`/`utcfromtimestamp` mixup) would pass every test here. Fix: add at least one integration-style test that calls `finish_run({"end_time": <epoch>})` against a real run and asserts the persisted `end_time`.

### 🟡 Minor / nits
- **`argus/backend/service/replay_service.py:657-670`** — Bucket resolution + `list_objects_v2` runs for *every* run touched by an archive, including generic/driver-matrix runs that never use S3 log links, adding an unconditional S3 round-trip to replays unrelated to SCT/S3 logs unless the caller opts out. Consider skipping backfill for run types that don't support S3 logs, or only attempting it when at least one `logs/submit` was seen for that run.
- **`argus/backend/service/replay_service.py:606-616`** — `_resolve_run_bucket` iterates a `set[str]` and returns the first regex match; Python's set iteration order for strings is hash-randomized per process, so the chosen bucket (and hence listed prefix) can vary run-to-run when a run recorded links from more than one bucket. Iterate a sorted/ordered sequence instead.
- **`argus/backend/service/replay_service.py:110-112`** — `_S3_LINK_RE` duplicates `TestRunService._match_s3_link` (`testrun.py:196-197`) with tiny drift (`[\w\-]+` vs `[\w\-]*`), per the code's own comment ("mirrors..."). Hoist into one shared function so the two can't silently desync.
- **`argus/backend/plugins/generic/model.py:86` vs `driver_matrix_tests/model.py:456` / `sct/testrun.py`** — The new epoch→`end_time` conversion is pasted into all three plugins with inconsistent UTC handling (`datetime.fromtimestamp(end_time, UTC)` vs. naive/deprecated `datetime.utcfromtimestamp(end_time)`). A shared helper on `PluginModelBase` would remove one bug surface instead of three.
- **`argus/backend/service/replay_service.py:104, 609, 622`** — New optional config key `REPLAY_LOG_BACKFILL_BUCKET` isn't documented in `argus_web.example.yaml`, unlike the sibling `AWS_CLIENT_ID`/`AWS_CLIENT_SECRET` keys it reuses. Add a commented entry, or drop the config rung if it's not meant to be used.
- **`argus/backend/service/replay_service.py:624-643`** — S3 pagination (`IsTruncated`/`ContinuationToken` loop) has zero test coverage; every mock hardcodes `IsTruncated: False`. Add a two-page test.

## Tests

The finalize-ordering and end_time-injection/fallback tests (`test_finalize_is_dispatched_and_runs_in_terminal_group`, `test_finalize_end_time_falls_back_to_last_seen_ts_when_own_ts_absent`, etc.) are well-targeted and assert the dispatch-layer behavior precisely, including the ms→s unit conversion. The `test_sct_api.py` nemesis-idempotency addition is a solid real (non-mocked) regression test — replays the same finalize twice and asserts stats/end_time are unchanged. The S3 backfill tests correctly verify the missing-vs-recorded diff and the disable/dry-run paths, but none use a region-qualified recorded link or exercise S3 pagination, and no test anywhere calls a plugin's `finish_run` against a real persisted run to check the actual datetime conversion — all replay-level assertions stop at "the right JSON reached a mock client." No assertions were rewritten to paper over new behavior; all test changes are additive.

## Notes

- The `finalize` un-skip also affects the `sirenada` plugin, whose `finish_run` unconditionally raises (`argus/backend/plugins/sirenada/model.py`, unchanged). No `ArgusSirenadaClient` caller ever invokes `finalize_run`, so this only bites a hand-packed archive (which the module docstring says is supported) — low likelihood, worth a one-line mention in the PR description if intentional.
- `finalize_nemesis`'s idempotency short-circuit (`sct/service.py:464`) returns `"updated"` for both a real update and a no-op skip — harmless today, but the caller can't distinguish the two from the return value.

<details>
<summary>Reviewer agents</summary>

- **correctness-and-tests** — Finalize-ordering/end_time-injection core is sound and well-tested at the dispatch layer, but flagged the region-URL mismatch that breaks backfill idempotency and the error-swallowing that can silently disable backfill forever; no test exercises the actual model-level datetime conversion.
- **security** — Critical: the S3 backfill derives its `Bucket=` parameter for `list_objects_v2` from untrusted archive content with no allowlist, turning the app's AWS credentials into an attacker-directed bucket-listing oracle — recommended mirroring the existing `S3_ALLOWED_BUCKETS` check used by `proxy_s3_file`.
- **pragmatism** — The S3 backfill subsystem is proportionate to a real, distinct problem (logs uploaded out-of-band to S3, unlike events); flagged a duplicated regex that admits mirroring existing code, and the epoch→end_time conversion pasted across three plugin models instead of one shared helper. Complexity rated appropriate overall.
- **production-readiness** — GO WITH CAUTION: design is sound and idempotent, but the failure-isolation boundary around S3 listing is drawn too narrowly, the new synchronous per-run S3 call has no timeout/backpressure, and backfill failures are invisible beyond a log line — recommended widening the try/except, adding client timeouts, and surfacing a failure counter before shipping.

</details>

PR #1044, opened 2026-07-16
feature(readme): make docs CLI-first

pehala:
## Summary
- rewrite the root README as a CLI-first landing page
- inline CLI installation and first-successful-command guidance in the root README
- move production deployment steps into a dedicated `docs/deployment.md`
- simplify `cli/README.md` by front-loading the most common authentication flows
- remove the redundant `docs/cli-install.md`

## Why
The previous root README was dominated by source deployment details and did not mention the CLI, even though the CLI is one of the most likely entry points for users. That made the front page useful mainly for self-hosting, while pushing common user tasks like installing the CLI, authenticating, and finding the right docs further away.

This change makes the top-level documentation better match what users usually want first:
- install the CLI
- run a first successful command
- find the full CLI guide
- find API, local-dev, and deployment docs quickly

It also reduces documentation drift by removing a separate CLI install page that had already diverged from the actual CLI behavior.

## Testing
- reviewed rendered markdown structure in the edited files
- verified links/references after removing `docs/cli-install.md`
- verified the PR diff against `upstream/master` only contains the intended docs changes


k0machi (approved):
LGTM

k0machi:
I would perhaps also look into addinng CI badges into root readme.

fruch:
> I would perhaps also look into addinng CI badges into root readme.

Who's the target audience for those ?

soyacz (approved):

PR #1048, opened 2026-07-20
feature(argusAI): event summarization evaluation harness

CodeLieutenant:
Here's the updated PR description covering both commits. Suggested title: feature(argusAI): event summarization — evaluation harness + worker integration, target scylladb/argus:master.

## Summary

Implements Event Summarization (design: #1016) end to end — both the offline **evaluation
harness** that picks the model and tunes the prompt, and the **production worker path** that
generates and stores summaries. SCT `ERROR`/`CRITICAL` events are long and noisy; summarizing
each unique event once (with a cheap model) turns a per-read cost into a one-time cost for
every later read by Zeus (and, optionally, humans).

Two commits:
1. Evaluation harness + shared `Summarizer`.
2. Worker integration: `sct_event.summary` + a best-effort `SummaryDispatcher`.

Nothing is hard-coded — model, API key, prompt, and concurrency are all configurable; the key
is read from config/env, never committed.

## Production path (worker)

- **`sct_event.summary`** — one nullable column (`flask sync-models` adds it automatically).
  Serving is zero-cost: `get_events` already does `SELECT *`, so the field travels in every
  payload. Null everywhere it isn't set, and every consumer falls back to `message`.
- **`SummaryDispatcher`** (`argusAI/utils/summary_dispatcher.py`) — summarizes the **raw**
  message of each **unique** event on a bounded `ThreadPoolExecutor` and writes it back with a
  raw-CQL `UPDATE sct_event SET summary = ? WHERE run_id = ? AND severity = ? AND ts = ?`.
  Dispatched from `event_similarity_processor_v2` immediately after the duplicate check
  resolves to *unique* (duplicates are never summarized — they're the bulk of volume).
- **Best-effort, isolated:** the embedding path is never blocked or altered; disabled unless
  `EVENT_SUMMARIZATION_ENABLED` + `OPENAI_API_KEY` are set; a failed/dropped summarization
  leaves `summary` null forever (no retries, no re-enqueue); backlog beyond a small in-flight
  cap is dropped with a log line rather than backpressuring the loop; the executor is drained
  on shutdown.

## Evaluation harness (`argusAI/eval/`)

Pulls real SCT events via the `argus` CLI, summarizes each across candidate models/prompts,
scores every summary against the **original** with a frontier LLM judge (coverage /
faithfulness / conciseness + dropped-fact and hallucination detection), and renders a
self-contained HTML report: a quality-vs-cost graph plus every event with all summaries
side-by-side. This is the evaluation gate the design requires before rollout.

Layered by responsibility: `events_source` (CLI extraction), `prompts`, `judge`, `pricing`,
`harness` (the sweep), `metrics` (aggregation), `report` (HTML), `results` (domain types),
`config` + `__main__` (CLI).

  Implements Event Summarization (design: #1016) end to end — both the offline **evaluation
  harness** that picks the model and tunes the prompt, and the **production worker path** that
  generates and stores summaries. SCT `ERROR`/`CRITICAL` events are long and noisy; summarizing
  each unique event once (with a cheap model) turns a per-read cost into a one-time cost for
  every later read by Zeus (and, optionally, humans).
  Nothing is hard-coded — model, API key, prompt, and concurrency are all configurable; the key
  is read from config/env, never committed.

## Validation

- Full sweep: 372 cells, **0 summarize/judge errors**.
- Judge parser hardened against malformed model JSON (null/string scores, trailing prose,
  code fences) — degrades to a recorded error, never aborts the sweep.
- Partial results are flushed on crash/Ctrl-C so paid API calls are never lost.
- `ruff` clean; behavior-preserving refactor verified against saved results.

## How to run

```bash
pip install -e '.[ai-eval]'
export OPENAI_API_KEY=sk-...
python -m argusAI.eval --config argusAI/eval/config.yaml --open
```

## Not in this PR (follow-ups)

- Web UI toggle + per-user "original vs. summary first" preference (design §5.5).
- CLI `--summary`/`--raw` flags (design §5.6).
- Zeus preferring `summary` when non-null (design §5.7).

Closes https://scylladb.atlassian.net/browse/ARGUS-146


argusAI/eval/config.example.yaml
```diff
@@ -0,0 +1,59 @@
...
+request_timeout: 90
+max_concurrency: 4           # bounded, to respect rate limits
+# argus_cli: /Users/you/bin/argus   # defaults to PATH, then ~/bin/argus
+no_cache: false
+output_dir: argusAI/eval/out
+
+# --- Pricing (USD per 1M tokens) — override the placeholders with real numbers -
+# Unknown models cost $0 and are flagged in the report.
+pricing:
+  gpt-5-nano:   { input_per_1m: 0.05, output_per_1m: 0.40, cached_per_1m: 0.005 }
```
soyacz: let's use recent models in example file

argusAI/eval/results.py
```diff
@@ -0,0 +1,70 @@
...
+    prompt_tokens: int = 0
+    completion_tokens: int = 0
+    cached_tokens: int = 0
+    latency_ms: float = 0.0
+    cost_usd: float = 0.0
+    judge: JudgeScore | None = None
+    error: str | None = None
+
+    @property
+    def compression_ratio(self) -> float:
```
soyacz, line 63: compression ratio is wrong. We could use tiktoken for counting tokens and not chars (this is what we pay for)

argusAI/utils/summarizer.py
```diff
@@ -0,0 +1,184 @@
...
+Rules:
+- Do NOT invent, infer, or add anything not present in the event. If something is unknown, omit it — never guess.
+- Do NOT editorialize or suggest fixes.
+- Drop only redundant boilerplate and repeated log noise. When in doubt, keep it.
+- Use the fewest tokens that lose no information a triager would need.
+
+Output plain text, no preamble.
+
+Length discipline (this is what keeps summarization worthwhile):
+- Never output more characters than you received. A summary that is longer than the original is a failure.
```
soyacz: LLM's are based on tokens, not characters and usually hallucinate with this character-based instructions. I think we should remove this line. This scenario can happen only when events are small - impact will be very low.

soyacz (commented):
another round - this time as me :)

argusAI/eval/results.py
CodeLieutenant: Done

argusAI/eval/config.example.yaml
CodeLieutenant: Using now 5.6-terra as the default

soyacz:
@CodeLieutenant is it ready for the review? will you add more things?

CodeLieutenant:
> @CodeLieutenant is it ready for the review? will you add more things? 

One more push

soyacz:
from AI:

**What it does:** Implements event summarization end to end (design #1016 / ARGUS-146): a nullable `sct_event.summary` column, a best-effort `SummaryDispatcher` hooked into the existing similarity worker after the unique-event check, a shared `Summarizer`, an offline eval harness (models × prompts × LLM judge → HTML report), plus CLI summary-first output with `--raw` and a Svelte "AI summary" toggle. The approach is sound: additive column, zero-cost serving via existing `SELECT *`, strict isolation of the LLM path from the embedding path, and an evaluation gate before rollout.

**Reviewed files:** 42 (full diff)
**Existing review comments noted:** 5 threads (char-based prompt instruction, token-based compression ratio, example model names, backlog cap, prompt provenance) — not re-reported.

## Issues (sorted by confidence)

1. **PR description** (confidence: 95/100) — The "Not in this PR (follow-ups)" section lists the Web UI toggle (§5.5) and CLI `--summary`/`--raw` (§5.6), but the diff includes both (`frontend/TestRun/SCT/SctEvent.svelte`, `cli/cmd/run_nemeses_events.go`).
   **Reproduction:** A reviewer trusting the description skips reviewing ~500 lines of shipped UI/CLI code.
   **Fix:** Update the PR body (the design doc's §12 status table is already correct).

2. **uv.lock** (confidence: 90/100) — `pyproject.toml` adds `openai` to the `ai` extra and a new `ai-eval` extra, but the committed `uv.lock` at the PR head contains no `openai`/`tiktoken` entries (verified via `git show HEAD:uv.lock`; the repo tracks the lockfile, and regenerating it locally produces a ~1000-line change).
   **Reproduction:** `uv sync --locked --extra ai` fails with "lockfile out of date"; plain `uv sync` (as CI runs it) silently re-resolves, so the pinned world drifts and `openai`'s version is unpinned.
   **Fix:** Run `uv lock` and commit the result.

3. **docs/plans/event-summarization.md:§4 vs argusAI/utils/summary_dispatcher.py:2493** (confidence: 85/100) — The design doc added in this PR states "**No min-length gate.** Every unique ERROR/CRITICAL event is summarized, full stop" as a review decision, but the code implements `EVENT_SUMMARIZATION_MIN_CHARS` and `argus_web.example.yaml` recommends `~800` as "the data-backed production value". §5.3's config table also omits `MIN_CHARS`, `MAX_BACKLOG`, and `PROMPT_VERSION`.
   **Reproduction:** An operator following §4 and one following the example yaml configure opposite behaviors; the doc can't be trusted as the decision record it claims to be.
   **Fix:** Update §4/§5.3 to reflect the gate (default 0 = off) and the two extra knobs, or drop the knob.

## Suggestions (low confidence)

- The reply "Using now 5.6-terra as the default" to soyacz's model-names comment is **not reflected at the current head** — `config.example.yaml`/`pricing.py` still list gpt-5-family only (may be the pending "one more push").
- `Summarizer`'s ImportError message says to install the `'ai-eval'` extra, but the production worker installs `'ai'` — mention both.
- Frontend: toggling one event's view silently rewrites the global `localStorage` default for all future events; consider separating "toggle this event" from "set my default".
- If embedding storage (step 4) fails after dispatch, reprocessing re-summarizes the event — idempotent write, but double API cost; a cheap guard would be checking `summary IS NOT NULL` isn't needed, just noting it's accepted.

## Strengths

- Best-effort isolation is genuinely engineered, not asserted: bounded semaphore backlog with drop-and-log, slot release on submit failure (with a test proving no leak), disabled/keyless/malformed-config modes all inert, executor drained on shutdown.
- Test diff is clean: no rewritten assertions, no skipped tests; the dispatcher suite covers exactly the design's §8 matrix, and the backend test exercises the worker's literal CQL upsert against the real schema.
- The broad `except Exception` blocks are the one place they're justified (LLM path must never break the embedding worker) and each carries a `noqa` with rationale.
- Shared `Summarizer`/`EventEmbedder` means the eval measures byte-for-byte what production will store; judge parser hardening (`raw_decode`, type coercion) degrades to recorded errors instead of aborting paid sweeps.
- Secrets stay in env/config per existing conventions; eval `.gitignore` keeps outputs and working configs out of git.

## Systemic notes

- Raw event text (test-infra log content) flows unsanitized into OpenAI prompts — deliberate and documented in §10, and the blast radius is one text column, but it is latent prompt-injection surface if event sources ever carry user-controlled text; worth revisiting when Zeus starts consuming summaries.
- The integration test (design §8, Docker ScyllaDB) is acknowledged as pending in §12 — fine as a follow-up, just don't let phase 3 close without it.

soyacz:
Merging this one. 
I have some ideas about improving the prompt - but we can do it in a followup.

PR #1049, opened 2026-07-22
feature(cli/users): add users list and get subcommands

k0machi:
Add an 'argus users' command with 'list' (all users: username, email, uuid)
and 'get' (single user by --username/--uuid/--email). Consolidate user
fetching and caching into a new UserService that owns getUsers, ListUsers,
GetUser and ResolveUserID; PlannerService now delegates to it via thin shims.

### DIFF
--- a/cli/cmd/users.go
+++ b/cli/cmd/users.go
@@ -1,0 +1,21 @@
+package cmd
+
+import (
+	"github.com/scylladb/argus/cli/cmd/users"
+	"github.com/spf13/cobra"
+)
+
+// usersCmd is the parent command for user lookups.
+var usersCmd = &cobra.Command{
+	Use:   "users",
+	Short: "Look up Argus users",
+	Long: `List Argus users or fetch a single user by username, UUID or email.
+
+  argus users list
+  argus users get --username alice`,
+}
+
+func init() {
+	users.Register(usersCmd)
+	rootCmd.AddCommand(usersCmd)
+}
--- a/cli/cmd/users/get.go
+++ b/cli/cmd/users/get.go
@@ -1,0 +1,61 @@
+package users
+
+import (
+	"github.com/scylladb/argus/cli/internal/cmdctx"
+	"github.com/scylladb/argus/cli/internal/logging"
+	"github.com/scylladb/argus/cli/internal/models"
+	"github.com/scylladb/argus/cli/internal/services"
+	"github.com/spf13/cobra"
+)
+
+// registerGet adds the "get" sub-command to parent.
+func registerGet(parent *cobra.Command) {
+	cmd := &cobra.Command{
+		Use:   "get",
+		Short: "Show a single user",
+		Long: `Show a single user addressed by exactly one of --username, --uuid
+or --email, e.g.:
+  argus users get --username alice
+  argus users get --uuid 7f3c1e90-...
+  argus users get --email alice@scylladb.com`,
+		RunE: runGet,
+	}
+
+	cmd.Flags().String("username", "", "Look up the user by username")
+	cmd.Flags().String("uuid", "", "Look up the user by UUID")
+	cmd.Flags().String("email", "", "Look up the user by email")
+	cmd.MarkFlagsMutuallyExclusive("username", "uuid", "email")
+	cmd.MarkFlagsOneRequired("username", "uuid", "email")
+
+	parent.AddCommand(cmd)
+}
+
+// runGet is the RunE handler for "users get".
+func runGet(cmd *cobra.Command, _ []string) error {
+	cmd.SilenceUsage = true
+	ctx := cmd.Context()
+	client := cmdctx.APIClientFrom(ctx)
+	out := cmdctx.OutputterFrom(ctx)
+	c := cmdctx.CacheFrom(ctx)
+	log := logging.For(cmdctx.LoggerFrom(ctx), "users-get")
+
+	var field, value string
+	for _, f := range []string{"username", "uuid", "email"} {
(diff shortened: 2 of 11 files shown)


soyacz:
get fetches picture_id - redundant
I would like also to get user by name e.g. `argus users get --name "john doe"` - but only if it's not requires backend changes (we could do it in a followup then)

fruch:
@soyacz do we have done mapping of which parts of the APIs are exposed in CLI ?

soyacz:
from AI:
Build passes, new tests pass, and the `getUsers` shim is still used (planner.go:770,883), so the delegation is justified. I have what I need.

### Findings

- **`get` and `list` present the same user inconsistently** (`get.go:59` vs `list.go:44`). `list` projects to `UserListEntry` with the id field relabeled `uuid`; `get` renders the raw `models.User` via `NewKVTabular`, so it surfaces `id` (not `uuid`), plus `full_name` and `picture_id`. Same command family shows the UUID under two different names and leaks `picture_id`, which is not user-facing info. Suggest `get` also project onto a consistent view (or extend `UserListEntry` to include `full_name`).

- **`GetUser` uuid branch linear-scans a map already keyed by UUID** (`users.go`, `match` func). `UsersMap` is keyed by `ID`, so the uuid case could be an O(1) key lookup. Negligible at this scale; the deliberate tradeoff is case-insensitive matching via `EqualFold`. Minor — worth a one-line comment noting why the scan, or just accept.

- **Duplicated match logic between `ResolveUserID` and `GetUser("username", …)`** (`users.go`). `ResolveUserID` is effectively `GetUser` by username returning only the ID with different error text. Not required to unify given the divergent messages/return, but flagging the overlap.

- **Nit (pre-existing, carried verbatim):** `ResolveUserID`'s doc says "0 or >1 matches error with candidate usernames," but the code only reports the *count* (`(%d matches)`), not the candidate names. Comment overstates the behavior; now duplicated in two files.

### Conventions / correctness
- Follows the established `cmd/<group>/{root,list,get}.go` + `Register(parent)` subpackage pattern (matches `planner`, `search`, `test`). ✅
- `MarkFlagsMutuallyExclusive` + `MarkFlagsOneRequired` correctly guarantee exactly one flag, so the first-non-empty loop in `runGet` is safe. ✅
- `getUsers` cache-`Set` error intentionally ignored — moved verbatim from prior code, consistent with the codebase. ✅

### Coverage gap (low priority)
No test asserts the memoization contract (single network fetch across repeated resolves, or that a `PlannerService` and its `UserService` share one fetch) — that shared-fetch behavior is the stated rationale for the refactor and is currently unverified. Cmd-layer (`list.go`/`get.go`) is untested but thin enough to accept.

Overall: clean, low-risk, mergeable. Only the `get`/`list` field-naming inconsistency is worth addressing before merge; the rest are optional polish.

soyacz:
> @soyacz do we have done mapping of which parts of the APIs are exposed in CLI ?

No, @k0machi can you send LLM to do this map along with missing parts? We could update this map each time we extend the cli

fruch:
> > @soyacz do we have done mapping of which parts of the APIs are exposed in CLI ?
> 
> No, @k0machi can you send LLM to do this map along with missing parts? We could update this map each time we extend the cli

Or the API, we have auto generate docs for the API ?

k0machi:
> get fetches picture_id - redundant I would like also to get user by name e.g. `argus users get --name "john doe"` - but only if it's not requires backend changes (we could do it in a followup then)

`get` should actually fetch `safe_user` filter in its entirety - including full_name and picture_id, so adding a full name search isn't a problem. Don't think removing the field for picture_id is necessary as you'll fetch it either way via that endpoint.

soyacz:
> > get fetches picture_id - redundant I would like also to get user by name e.g. `argus users get --name "john doe"` - but only if it's not requires backend changes (we could do it in a followup then)
> 
> 
> 
> `get` should actually fetch `safe_user` filter in its entirety - including full_name and picture_id, so adding a full name search isn't a problem. Don't think removing the field for picture_id is necessary as you'll fetch it either way via that endpoint.

This picture is is not used and it will increase token usage for llm. Just dropping it from the final response should not be a big deal

k0machi:
@soyacz Done, using `UserListEntry` now for both get and list. I've also added `search` subcommand that will search across all provided fields and full name.

soyacz (approved):
Looks good, we can merge this one.
Before merging, some of below could be fixed (not that important, but still worth):

### Findings (all minor)

- **PR description is stale.** Body describes only `list` and `get`; `search` and the whole diacritic-normalization feature (a separate commit, ~100 lines) are undocumented. Per review discipline, undocumented intent is a finding — update the PR body to state the search + accent-folding approach.
- **`GetUser` silently defaults unknown `field` to username** (`services/users.go`, `default:` in the `match` closure). Safe for current callers (mutually-exclusive validated flags), but as an exported method an invalid field is treated as `username` rather than erroring. Consider an explicit `default: return false` + error, or document that `field` is trusted.
- **Empty-flag edge:** `--username ""` passes cobra's `MarkFlagsOneRequired` (flag is *changed*), but `runGet`'s loop skips empty values, leaving `field=""` → falls through to a username match against `""`. Yields a confusing "no user with `\"\"`" rather than rejecting the empty value. Low priority.
- **Cosmetic:** the `json:"uuid"` tag renders the column/key header as `Uuid` (via `snakeToTitle`), not `UUID`. Matches the existing reflection behavior, so not new, but the `uuid` field is new here.
- **Micro-perf (negligible):** `SearchUsers` rebuilds the `transform.Chain` inside `normalizeLatin` for every field of every user. Fine at user-list scale; only worth noting if the map ever grows large.

PR #1052, opened 2026-07-24
cli/validate params for execute

k0machi:
- **feature(service/jenkins): validate Scylla version source for SCT builds**
  scylla-cluster-tests jobs require exactly one Scylla version source
  (scylla_version, scylla_repo, an image, a rolling-upgrade repo, or a BYO
  repo) to start correctly; setting none or several silently produces a
  doomed build. Add JenkinsService.validate_sct_version_source, mirroring
  the SCTParameterWizard front-end, and call it in the /jenkins/build
  controller when the resolved test's plugin is scylla-cluster-tests so
  CLI/UI/API callers get a clear DataValidationError instead of a broken
  run. The check also closes the UI gap where the raw-param tabs bypass the
  client-side wizard validation.
  

- **feature(cli/test): clear conflicting Scylla version sources before execute**
  Job defaults frequently seed several Scylla version-source parameters at
  once (e.g. a default scylla_version alongside a scylla_repo), which the
  backend now rejects as ambiguous for scylla-cluster-tests jobs. When the
  caller explicitly picks exactly one source via --file/--param, clear the
  sibling families' keys (emptying rather than deleting, so Jenkins cannot
  re-apply its own default) so only the chosen source is sent; ambiguous or
  absent selections are left for the backend validator.
  
  Add IsValidationError so a backend DataValidationError is recognised as
  terminal: in plan fan-out, abort immediately instead of triggering the
  remaining doomed builds.
  

- **feature(service/jenkins): add fromDefaults mode to job parameter fetch**
  retrieve_job_parameters seeds a job's parameters from its last build by
  default, which carries over whatever the previous run mutated. Add a
  fromDefaults flag that instead returns the job's configured default
  parameters (the initial set), independent of any build history — this
  also lets a fresh job's parameters be fetched without a prior build.
  Wire it through the /jenkins/params controller (default false, so the
  front-end's rebuild/clone behaviour is unchanged). Factor the default
  extraction into _default_params_from_job_info, reused by both modes.
  

- **feature(cli/test): seed execute/params from job defaults, add --rebuild**
  test execute and test params now seed parameters from the job's
  configured defaults by default, giving each trigger a clean initial set
  rather than inheriting the last build's values — important for --plan-id
  fan-out, where every job would otherwise carry over whatever its previous
  run changed. Add --rebuild to restore last-build seeding; --build-number
  still targets a specific build (and implies rebuild-style seeding). The
  mode is sent to the backend via the new fromDefaults param.

### DIFF
--- a/argus/backend/controller/testrun_api.py
+++ b/argus/backend/controller/testrun_api.py
@@ -436,6 +436,17 @@ def build_jenkins_job():
     payload = get_payload(request)
     service = JenkinsService()
 
+    # scylla-cluster-tests jobs require exactly one Scylla version source to be
+    # set, otherwise the build fails to start. Validate before triggering so the
+    # caller (CLI/UI/API) gets a clear error instead of a broken run. Skipped
+    # when the test/plugin can't be resolved (e.g. a brand-new job).
+    try:
+        test = ArgusTest.get(build_system_id=payload["buildId"])
+    except ArgusTest.DoesNotExist:
+        test = None
+    if test and test.plugin_name == "scylla-cluster-tests":
+        JenkinsService.validate_sct_version_source(payload["parameters"])
+
     result = service.build_job(build_id=payload["buildId"], params=payload["parameters"])
 
     response = {
--- a/argus/backend/controller/testrun_api.py
+++ b/argus/backend/controller/testrun_api.py
@@ -419,7 +419,11 @@ def get_jenkins_job_params():
     payload = get_payload(request)
     service = JenkinsService()
 
-    result = service.retrieve_job_parameters(build_id=payload["buildId"], build_number=payload["buildNumber"])
+    result = service.retrieve_job_parameters(
+        build_id=payload["buildId"],
+        build_number=payload["buildNumber"],
+        from_defaults=payload.get("fromDefaults", False),
+    )
 
     return {
         "status": "ok",
--- a/argus/backend/service/jenkins_service.py
+++ b/argus/backend/service/jenkins_service.py
@@ -84,18 +84,25 @@ class JenkinsService:
             choices[name] = [s.text for s in define.findall("choices//string") if s.text is not None]
         return choices
 
-    def retrieve_job_parameters(self, build_id: str, build_number: int | None) -> list[Parameter]:
+    def _default_params_from_job_info(self, job_info: dict) -> list[Parameter]:
+        """Build the parameter list from the job's configured default values
+        (its "initial" parameters), independent of any past build. Optional
+        choice parameters default to empty rather than their first option."""
+        default_params = next((prop for prop in job_info["property"] if prop.get(
+            "_class", "") == "hudson.model.ParametersDefinitionProperty"), {}).get("parameterDefinitions", {})
+        return [{"name": param["name"],
+                 "value": "" if param["name"] in self.OPTIONAL_CHOICE_PARAMETERS
+                 else param.get("defaultParameterValue", {}).get("value", "")}
+                for param in default_params if param["name"] != self.RESERVED_PARAMETER_NAME]
+
+    def retrieve_job_parameters(self, build_id: str, build_number: int | None,
+                                from_defaults: bool = False) -> list[Parameter]:
+        """Return a job's parameters. By default (from_defaults=True, used by the
+        CLI) the job's configured default values are returned so a fresh trigger
+        starts from a clean slate. When from_defaults is False the values are
+        seeded from a past build instead: the given build_number, or the last
+        build when build_number is falsy (the rebuild/clone behaviour)."""
         job_info = self._jenkins.get_job_info(name=build_id)
-        if not build_number:
-            next_build_number = job_info.get("nextBuildNumber")
-            if not next_build_number:
-                raise JenkinsServiceError("#noBuildsAvailable")
-            try:
-                build_info = self._jenkins.get_build_info(name=build_id, number=next_build_number - 1)
-            except jenkins.JenkinsException:
-                raise JenkinsServiceError("#noBuildsAvailable")
-        else:
-            build_info = self._jenkins.get_build_info(name=build_id, number=build_number)
         raw_config = self._jenkins.get_job_config(name=build_id)
         config = ET.fromstring(raw_config)
         parameter_defs = config.find("*//parameterDefinitions")
@@ -107,17 +114,26 @@ class JenkinsService:
         else:
             descriptions = {}
         choices = self._extract_choice_parameters(config)
-        params = next((a for a in build_info["actions"] if a.get(
-            "_class", "#NONE") == "hudson.model.ParametersAction"), None)
-        if params:
-            params = [param for param in params["parameters"] if param["name"] != self.RESERVED_PARAMETER_NAME]
+
+        if from_defaults:
+            params = self._default_params_from_job_info(job_info)
         else:
-            default_params = next((prop for prop in job_info["property"] if prop.get(
-                "_class", "") == "hudson.model.ParametersDefinitionProperty"), {}).get("parameterDefinitions", {})
-            params = [{"name": param["name"],
-                       "value": "" if param["name"] in self.OPTIONAL_CHOICE_PARAMETERS
-                       else param.get("defaultParameterValue", {}).get("value", "")}
-                      for param in default_params if param["name"] != self.RESERVED_PARAMETER_NAME]
+            if not build_number:
+                next_build_number = job_info.get("nextBuildNumber")
+                if not next_build_number:
+                    raise JenkinsServiceError("#noBuildsAvailable")
+                try:
+                    build_info = self._jenkins.get_build_info(name=build_id, number=next_build_number - 1)
+                except jenkins.JenkinsException:
+                    raise JenkinsServiceError("#noBuildsAvailable")
+            else:
+                build_info = self._jenkins.get_build_info(name=build_id, number=build_number)
+            params = next((a for a in build_info["actions"] if a.get(
+                "_class", "#NONE") == "hudson.model.ParametersAction"), None)
+            if params:
+                params = [param for param in params["parameters"] if param["name"] != self.RESERVED_PARAMETER_NAME]
+            else:
+                params = self._default_params_from_job_info(job_info)
         self._apply_choice_defaults(params, choices, descriptions)
 
         return params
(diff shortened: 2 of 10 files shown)


soyacz:
# PR #1052 — cli/validate params for execute

**Author:** k0machi · `cli/validate-params-for-execute` → `master` · +480/-46, 10 files
**Verdict:** Changes requested. One blocking regression (F1) plus an open semantic
question on the family model (F2). Risk tier **Medium-High** — touches the
build-trigger path shared by CLI, UI and the planner API.

## HLD (reconstructed)

**Problem.** scylla-cluster-tests jobs need exactly one Scylla version source
(`scylla_version` / `scylla_repo` / a cloud image / `new_scylla_repo` / a BYO repo).
Setting none or several produces a build that starts and dies. Separately,
`test execute` seeded parameters from the job's *last build*, so every trigger
inherited whatever the previous run mutated — bad for `--plan-id` fan-out.

**Approach.** Three layers:
1. Backend validator `JenkinsService.validate_sct_version_source`, called from the
   `/jenkins/build` controller when the resolved test's plugin is
   `scylla-cluster-tests`.
2. `retrieve_job_parameters(from_defaults=...)` to return the job's *configured*
   defaults instead of a build's parameters; wired through `/jenkins/params`
   (default `False`, so the UI is untouched).
3. CLI: seed from job defaults by default (`--rebuild` restores last-build
   seeding), pre-clear sibling version-source families when the caller explicitly
   picks exactly one, and treat a backend `DataValidationError` as terminal in
   plan fan-out.

Intent is well documented in the commit messages — no HLD finding. Tests pass
locally (`go test ./...` all ok; `pytest argus/backend/tests/jenkins_service` 19
passed).

---

## F1 — Blocking: bare `test execute` on an SCT job now fails where it used to work

Layers 1 and 3 interact badly in the *no-override* case. Chain:

`argus test execute --build-id <sct-job>` with no `--param`/`--file`
→ `seedFromDefaults` = true (`cli/cmd/test/params.go:100`)
→ backend returns the job's configured defaults, which per commit 2's own message
"frequently seed several Scylla version-source parameters at once"
→ `explicitParamKeys(nil, nil)` is empty → `families == 0`
→ `normalizeSCTVersionSource` returns early (`cli/cmd/test/execute.go:405`)
→ controller raises `DataValidationError` (`argus/backend/controller/testrun_api.py:452`).

Before this PR the same invocation seeded from the last build, which by
construction carried exactly one source, and succeeded. The CLI's normalization
only rescues the case where the user *already* named a source — i.e. exactly the
case that needed no rescuing.

**Compounding:** `--plan-id/--label` with no `--param` hits this on target #1, and
the new abort path returns before `out.Write(results)`:

```go
if services.IsValidationError(err) {
    return fmt.Errorf("aborting plan execution: %s failed validation: %w", ...)
}
```
(`cli/cmd/test/execute.go:273`)

So any target that was already queued earlier in the loop is lost from the output
— no queue items, no URLs, no per-test statuses. The abort itself is reasonable;
returning without writing `results` is not. Write the partial `results` (marking
the aborted rows) and *then* return the error.

**Also affected:** the documented round-trip
`test params > params.json` → `test execute --file params.json`. `test params` now
emits the job defaults, so the file can contain two non-empty sources, and feeding
it straight back is rejected.

Options: pick a deterministic family when none is explicit; or fail locally with
the found sources listed rather than after a round-trip; or keep last-build seeding
as the default and make `--from-defaults` opt-in. Whichever — the bare invocation
must not regress silently.

## F2 — Open question: is `rolling_upgrade` really exclusive with `scylla_version`?

`SCT_VERSION_SOURCE_FAMILIES` treats `new_scylla_repo` as a *mutually exclusive*
alternative to `scylla_version`. If SCT rolling-upgrade jobs legitimately carry a
base `scylla_version` (or `base_versions`) *plus* the upgrade target
`new_scylla_repo`, the validator rejects correct builds — which would invalidate
commit 1, not just polish it. I could not verify SCT's semantics from this repo;
please confirm.

Related gap either way: `sctFamilyCompanionKeys` lists `byo_scylla_branch` for
`scylla_byo` but nothing for `rolling_upgrade`, while the wizard shows
`base_versions` alongside `new_scylla_repo` (`SCTParameterWizard.svelte:423`).
Should `base_versions` be cleared when switching away from `rolling_upgrade`?

## F3 — The validator is *stricter* than the wizard it claims to mirror

The comment says the families mirror `SCTParameterWizard`. They don't, in two ways
that let the UI produce a payload the backend now rejects:

- `determineVersionSource` (`SCTParameterWizard.svelte:22-30`) has no
  `byo_scylla_repo` key. A BYO job (`byo_scylla_repo` set, `scylla_version` empty)
  falls through to the `scylla_version` default source, the wizard renders an empty
  `scylla_version` field, and `byo_scylla_repo` stays populated.
- No `onChange` handler clears `new_scylla_repo`, `byo_scylla_repo` or
  `byo_scylla_branch`. `scylla_version`'s handler (lines 262-268) clears only
  repo + the four images. And handlers fire on edit only, so untouched inherited
  values survive regardless.

Result: a rebuild of a BYO or rolling-upgrade job through the UI can now fail
validation where it previously went through. Tightening the raw-param tabs is
stated intent; this asymmetry looks unintended. Fix the wizard to clear all
sibling families (and add `byo_scylla_repo` to `determineVersionSource`).

## F4 — Family definitions triplicated with no consistency check

`SCT_VERSION_SOURCE_FAMILIES` (Python), `sctVersionSourceFamilies` (Go),
`keysToCheck` + per-field `onChange` clearing (Svelte). The Go map carries
`// Keep this in sync with argus/backend/service/jenkins_service.py` — nothing
enforces it, and F3 is what drift already looks like. Minimum: serve the families
from the backend (or an existing shared constants surface) so the wizard and the
CLI read one definition; failing that, a test asserting the Go and Python maps
match.

## F5 — `anyKeyExplicit` counts key *presence*, backend counts non-empty *value*

`explicitParamKeys` discards values (`cli/cmd/test/execute.go:437`), so an
explicitly-emptied sibling still counts as a selected family. Repro:

```
--param scylla_repo=http://x --param scylla_version=
```
→ explicit contains both keys → `families == 2` → normalization disabled entirely
→ an inherited non-empty `scylla_ami_id` from job defaults survives → backend
rejects. Under value-based counting `chosen = scylla_repo`, the AMI is cleared, and
the trigger works.

Fix: `anyKeyExplicit` should require a non-empty explicit value. (This does *not*
rescue a `params.json` holding two non-empty sources — that rejection is correct;
see F1 for the round-trip complaint.)

## F6 — Validation lives in the controller, so two trigger paths bypass it

`validate_sct_version_source` is called only from `build_jenkins_job`. Bypassed by:

- `argus/backend/service/planner_service.py:772` — `service.build_job(...)` for the
  backend-side plan trigger, building `final_params` from merged param sets.
- `argus/backend/controller/testrun_api.py:552` — `clone_build_job`, which calls
  `build_job` internally (`jenkins_service.py:350`).

If the invariant is "an SCT build must have exactly one source", it belongs in
`build_job` (resolving the test there, as the controller does), not in one caller.

## F7 — The new controller branch is untested; mutating it breaks nothing

`test_jenkins_build_returns_queue_item` is green only because
`ArgusTest.get(build_system_id="job/foo")` misses and `test` is `None`. Flip `==`
to `!=` on the `plugin_name` check and no test fails — the one line gating a
user-facing rejection has no coverage. The service-level tests for
`validate_sct_version_source` are good, but they don't cover the wiring.

Add: a test with a resolvable SCT test asserting the rejection, and one with a
non-SCT plugin asserting pass-through.

## F8 — Docstring contradicts the signature

```python
def retrieve_job_parameters(self, build_id, build_number, from_defaults: bool = False):
    """Return a job's parameters. By default (from_defaults=True, used by the
    CLI) the job's configured default values are returned ..."""
```
(`argus/backend/service/jenkins_service.py:96-101`)

The default is `False`. Reword to describe the API default (build-seeded) and note
that the CLI opts in.

---

## Minor

- `_default_params_from_job_info`: `.get("parameterDefinitions", {})` — an empty
  dict iterates to strings and `param["name"]` would raise `TypeError`. Moved
  verbatim, but `[]` is the correct default while you're in there.
- `ArgusTest.get(build_system_id=...)` can raise `MultipleObjectsReturned`; only
  `DoesNotExist` is caught. `validate_build_system_id` makes duplicates unlikely,
  not impossible.
- `JenkinsService()` is constructed (opening a Jenkins connection) before
  validation runs — move the validation above line 441.
- `"scylla-cluster-tests"` is a bare literal in the controller; there may be a
  plugin-name constant to use.

## What's good

- Tests are honest: the two fetch modes are made distinguishable by giving the fake
  job different defaults vs. last-build values, so the mode assertions can actually
  fail. No assertion was loosened to accommodate new behaviour — the single test
  edit (`from_defaults=False` in `assert_called_once_with`) tightens it.
- `normalizeSCTVersionSource`'s no-op cases are explicitly tested, including
  "absent sibling keys must stay absent".
- Emptying rather than deleting keys so Jenkins can't re-apply its own default is
  the right call, and the reasoning is in the comment.
- No weakened gates: nothing skipped, no suppressions, CI untouched. UI
  `/jenkins/params` callers (`JenkinsBuildModal`, `JenkinsCloneModal`) both omit
  `fromDefaults`, so the `False` default genuinely preserves their behaviour.

k0machi:
> F1 — Blocking: bare `test execute` on an SCT job now fails where it used to work

Why is this blocking? This is exactly what was asked for. Provide --rebuild flag if you do not want to specify parameters.

> F2 — Open question: is rolling_upgrade really exclusive with scylla_version?

Since when it isn't? Rolling upgrade works differently from standard tests...

> F6 — Validation lives in the controller, so two trigger paths bypass it

This is fine.

soyacz:
> > F1 — Blocking: bare `test execute` on an SCT job now fails where it used to work
> 
> Why is this blocking? This is exactly what was asked for. Provide --rebuild flag if you do not want to specify parameters.
> 
> > F2 — Open question: is rolling_upgrade really exclusive with scylla_version?
> 
> Since when it isn't? Rolling upgrade works differently from standard tests...
> 
> > F6 — Validation lives in the controller, so two trigger paths bypass it
> 
> This is fine.

fix things that are worth fixing.
Also, in commit msg would be nice to show params in examples - so we could easily see what's change about and how it will look. Can you do that?

k0machi:
There's nothing too important in that review, I'll fix the typos real quick (F8 and some others) and it's ready for merge

k0machi:
@soyacz ready

soyacz (approved):

PR #1053, opened 2026-07-29
feature(cli/view): add view management commands

k0machi:
Add a `view` subcommand mirroring `planner` with list, get, create,
update, and delete for managing Argus user views. References (items and
widget filters) use release-qualified build_system_id paths resolved to
UUIDs via a shared global resolver. `get` and `list` surface the plan
link as its human-friendly plan key and a web URL that opens the view.


soyacz:
AI review:

## What this PR does

**Problem** — Argus user views (dashboards) had no CLI management, unlike plans.
**Approach** — Adds a `view` subcommand (list/get/create/update/delete) mirroring `planner`'s structure; items and widget filters use release-qualified build_system_id paths resolved to UUIDs via a new cross-release `PlannerService.ResolveGlobalRef`.
**Risk tier** — medium: new write paths (create/update/delete) against user data, gated by confirmation, but the reference-resolution plumbing is genuinely new and non-trivial.
**Ticket** — _none linked_

## Verdict

Needs changes — one data-loss bug in the widget-filter round-trip. 1 critical, 1 major, 2 minor.

## Findings

### 🔴 Critical
- **`cli/internal/services/view.go:397-420` (`widgetsToTemplate`) + `:211-227` (`resolveFilterRefs`)** — A widget filter UUID that isn't also a top-level view item is reverse-mapped as a bare UUID string with a warning (line 414). Fed back through `resolveFilterRefs` → `ResolveGlobalRef` (`planner.go:369`), a bare string with no `/` is treated as a release name, fails, and is silently dropped as "not found." Concretely: `view update --view my-dashboard --display-name "X"` (metadata-only, no `--file`) rebuilds the template from the current view and resubmits it — any such widget filter is silently deleted, and the same happens round-tripping a `get`-emitted file through `create --file`. Fix: skip `ResolveGlobalRef` when the ref already `looksLikeUUID` and pass it through unchanged.

### 🟠 Major
- **`cli/internal/services/view.go:231-239` (`isNotFound`)** — The `strings.HasPrefix(err.Error(), "no release named ")` branch string-matches `ResolveReleaseID`'s literal message (`planner.go:222`, not sentinel-wrapped) instead of `errors.Is`, which the rest of the package uses 10+ times. No test exercises an unresolvable bare release name in an item/filter (only an unresolvable test path is tested, hitting the wrapped-sentinel branch), so a future wording change to that message silently turns warn-and-skip into a hard error on create/update. Fix: add that test case, and prefer `errors.Is` for the sentinel branch.

### 🟡 Minor
- **`cli/cmd/view/delete.go:88-99`** — `confirm()` is byte-for-byte copy-pasted from `cli/cmd/planner/delete.go:90-101`. No shared home exists for it yet; extract to e.g. `cli/internal/prompt.Confirm` before a third copy appears.
- **`cli/cmd/view/*`** — Unlike `cmd/planner`, `cmd/search`, `cmd/test` (each with a `*_test.go`), the new `view` cmd package has no flag-registration/overlay-precedence tests (c.f. `planner_test.go`'s `TestOverlayFlags_FlagOverFilePrecedence`).

## Tests

`view_test.go` is solid on happy paths (cross-release resolution, UUID↔reference mapping, request shaping, name-vs-UUID addressing) but has the two gaps above: no unresolvable-release-name case, and nothing round-trips an unmapped filter UUID back through `BuildCreateRequest`/`BuildUpdateRequest` — exactly where the critical bug lives.

<details>
<summary>Reviewer agents</summary>

- **correctness-and-tests** — found the critical filter round-trip bug and confirmed the `isNotFound` coverage gap; the post-write-network-call error pattern in create/update is pre-existing (`planner` does the same), not new.
- **pragmatism** — mirroring of `planner` is faithful and proportionate; only real duplication cost is the copy-pasted `confirm()`.
</details>

soyacz:
One thing that it still makes this one incomplete: we need to show docs for available widgets. 
There should be backend api endpoint that will tell all the widgets we have with a docstring and available settings. Without it it's hard to build/update a view.
One use case I've heard is to update plan view by adding highlights widget after we build the plan (manual action currently).

k0machi:
> One thing that it still makes this one incomplete: we need to show docs for available widgets. There should be backend api endpoint that will tell all the widgets we have with a docstring and available settings. Without it it's hard to build/update a view. One use case I've heard is to update plan view by adding highlights widget after we build the plan (manual action currently).

Yes, I am thinking about widgets right now, I'd say this PR is 80% complete, currently what's left is the widget discovery/linking mechanism. I also plan to rework template for widgets because at this point it's too much information.

Maybe for creation I'll have simple `--widget $classname` chain since most of them you don't need to configure, only thing that matters is order.

k0machi:
> A widget filter UUID that isn't also a top-level view item is reverse-mapped as a bare UUID string with a warning (line 414). 

I do not understand this wording. Does it mean a filter item that is not present in the tests? In this case it should be removed. I'll double check the operation precedence anyhow for IsUuid checks. Also, template returns resolved build ids for widget filter items and tests, so I don't see if this'll be a problem.

Update:

```go
			warnings = append(warnings, fmt.Sprintf("widget %q filter entity %q is not a view item — kept as UUID", w.Type, id))
```
This situation is not a data loss - if a filtered item no longer exists, then it should not pass validation when being submitted.

soyacz:
From the UX perspective, looks good. See if these findings are valid:

### Critical — carried over, still present

**`cli/internal/services/view.go` `widgetsToTemplate` (kept-as-UUID
fallback) + `resolveFilterRefs`/`ResolveGlobalRef` (consumer)** — Confirmed
still reproducible against the current diff. Neither new commit added a
UUID short-circuit anywhere in this path; `resolveFilterRefs` still calls
`ResolveGlobalRef` unconditionally, and `ResolveGlobalRef` has no branch for
a bare UUID (only "single segment = release name" or "release/rest =
group/test path").

Concrete round-trip:

1. `ViewToTemplate` reverse-maps a widget's filter UUIDs against `refByID`,
   which is built only from the view's top-level `Items`. A filter UUID
   naming a valid test or group that is *not itself a top-level item* (a
   normal case — filtering a widget to one test inside a release/group item
   that is the actual view item) has no entry in `refByID`. It is kept
   verbatim as a bare UUID string and reported as a warning. This is exactly
   what `TestViewService_ViewToTemplate`'s `"unmapped-uuid"` case exercises.
2. Feed that template back through `view update` (metadata-only, no
   `--file`) or `view create --file <get output>`. `BuildUpdateRequest` /
   `BuildCreateRequest` call `resolveFilterRefs` on that same filter list,
   which calls `ResolveGlobalRef` on the bare UUID. A UUID has no `/`, so it
   is treated as a release name, `ResolveReleaseID` fails, `isNotFound`
   matches, and the entry is silently dropped with a warning (not an error).

Net effect: any widget filter entity that is not also a top-level view item
is silently deleted on the next metadata-only update or file round-trip.

**On the author's rebuttal** ("this is not data loss — if a filtered item
no longer exists, it should not pass validation on submit anyway"): this
conflates two different things. The dropped entity is not stale or
nonexistent — it resolves fine via `ResolveGlobalRef` given its
`build_system_id` path, because it is a real test/group in the release
grid. The actual bug is a representation mismatch: `widgetsToTemplate`'s
"kept as UUID" fallback produces a bare UUID, but the only consumer of that
value (`ResolveGlobalRef`) accepts build_system_id paths and release names,
never bare UUIDs. So the fallback's own output is unusable by the code that
has to re-resolve it. Existence/validation is not what causes the drop;
the format mismatch is.

Fix, as originally suggested: in `resolveFilterRefs` (and `resolveItemRefs`
if item UUIDs can hit the same path), check `looksLikeUUID(ref)` first and
pass the value through unchanged instead of calling `ResolveGlobalRef`.
`looksLikeUUID` already exists in the file for this exact purpose but is
currently used only for `--view` addressing, not for filter/item refs.

No test exercises this round-trip (template with an unmapped filter UUID
fed into `BuildUpdateRequest`/`BuildCreateRequest`); that gap, flagged in
the prior review, also still stands.

### Major — carried over, still present

**`cli/internal/services/view.go` `isNotFound`** — Still string-matches
`ResolveReleaseID`'s literal error text (`"no release named "`) instead of
using a sentinel + `errors.Is`, the pattern used everywhere else in the
package. No test exercises an unresolvable bare release name in an
item/filter ref (only an unresolvable test path is tested, which hits the
wrapped-sentinel branch, not this string-prefix branch). A future wording
change to that message would silently turn "warn and skip" into a hard
error on `create`/`update`.

k0machi:
> On the author's rebuttal ("this is not data loss — if a filtered item
no longer exists, it should not pass validation on submit anyway"): this
conflates two different things. The dropped entity is not stale or
nonexistent — it resolves fine via ResolveGlobalRef given its
build_system_id path, because it is a real test/group in the release
grid. The actual bug is a representation mismatch: widgetsToTemplate's
"kept as UUID" fallback produces a bare UUID, but the only consumer of that
value (ResolveGlobalRef) accepts build_system_id paths and release names,
never bare UUIDs. So the fallback's own output is unusable by the code that
has to re-resolve it. Existence/validation is not what causes the drop;
the format mismatch is.

Again, if an entity was removed from a view then the filter for that entity is invalid, it would be **HORRIBLE** UX if you later re-add that entity and not see it because your old filter still functions there.

I could add explicit validation, but this way of just dropping it because it does not resolve in the `items` array is absolutely fine.

And it's totally fine to fail due to format because *We do not support UUIDs in the template file*.

k0machi:
@soyacz ready again

k0machi:
> Again, if an entity was removed from a view then the filter for that entity is invalid, it would be HORRIBLE UX if you later re-add that entity and not see it because your old filter still functions there.

Actually, this gives me an idea, a filter for an entity being removed should also be removed to prevent stale references like this.

soyacz (approved):
LGTM

PR #1057, opened 2026-08-02
feature(ssh-tunnel): fetch one key per auth, add reconnect and failover

CodeLieutenant:
## Summary

Three changes to the SSH proxy tunnel. Every change is backward compatible.

The tunnel exists to keep automated SCT and dtest traffic off Cloudflare. Two problems worked
against that. The proxy host downloaded the whole key table on every SSH authentication attempt.
A client that failed to build a tunnel stayed on Cloudflare for the rest of the run, because
nothing retried in the background.

## Key lookup: one key per authentication

`GET /client/ssh/keys` called `SSHTunnelKey.objects.all()`, an unfiltered table scan that returned
every non-expired key of every user. The response grew with the size of the CI fleet, and the
proxy host fetched it on every connection attempt.

sshd expands the `%f` token to the SHA-256 fingerprint of the key the client offered. That string
is byte-identical to what `_derive_fingerprint()` already stores. Three parts use it:

- `SSHTunnelKey.fingerprint` carries a secondary index, so the lookup is an exact match.
- `get_authorized_keys(fingerprint=...)` returns only the matching key. An unknown fingerprint
  returns an empty body, which sshd reads as a denial. A malformed fingerprint gets a plain-text
  400. The shared JSON error handler replies 200, and sshd would read that JSON body as an
  `authorized_keys` file.
- `BuildSSHConnectArgs` sets `IdentitiesOnly=yes`. Without it, `ssh` offers every agent key and
  default identity before `-i`, and each rejected offer costs another lookup and one more
  `MaxAuthTries` slot. The Python client already set this, so only the Go path was affected.

The lookup is not scoped by tunnel. All proxies front the same backend, so a key registered
against one proxy and accepted by another is not an escalation. Scoping would break a client that
has failed over.

A call without the parameter still returns the full list and logs a warning. That path exists only
for proxy hosts that still run the old wrapper.

## Handshake cost

Measured on production, the handshake ran at **25.3 req/min while the tunnel carried 7.6**. Building
a tunnel costs a registration call plus one authorized-keys lookup, and both travel direct because
they are what create the tunnel. So the feature was spending more public requests than it removed.

Two changes, one per repository:

- **The client does not tunnel until it is worth it.** `TunneledSession` counts requests and the
  monitor stays idle below `ARGUS_TUNNEL_MIN_REQUESTS` (default 10). A one-shot client such as
  `argus-client-generic submit` never handshakes at all. A long SCT run crosses the threshold in
  seconds and spends ten direct requests out of thousands. This replaces the startup wait, so no
  request blocks on ssh at any point.
- **The proxy host looks keys up over the private network** (in `qatools-deployments`). The
  `AuthorizedKeysCommand` runs once per SSH authentication attempt, and it was pointed at
  `https://argus.scylladb.com` — every authentication went through Cloudflare, from inside the VPC,
  to reach a backend the host already forwards TCP to. `argus_tunnel_keys_url` now defaults to the
  backend's private address.

Which of the two dominates the 25.3 is worth confirming from
`http_request_by_endpoint_total`: `get_authorized_keys` is proxy-side, `register_tunnel` and
`get_tunnel_connection` are client-side.

## Client reconnect

`TunneledSession._backoff()` set a flat 30 s cooldown and cleared `self._tunnel`. The monitor
thread's health check returned early on exactly that condition, so it never re-attempted. The
tunnel was retried only on the next request after the cooldown, and that attempt ran inline under
the session lock. It could stall an SCT heartbeat for about 100 seconds.

The `argus-tunnel-monitor` thread now owns every tunnel state transition:

- `request()` reads the port and rewrites the URL. It never spawns ssh and never blocks on it. On
  `ConnectionError` it flags the tunnel and retries directly, so a dead proxy costs one direct
  request instead of a stalled caller.
- The monitor retries with an escalating delay, 30 s to 600 s, each carrying 20 % jitter. A fleet
  of runners that loses one proxy at the same moment must not retry in lockstep against its
  replacement.
- A tunnel that was working earns one immediate re-attempt and a fresh delay ladder.
- On success the log records `SSH tunnel restored`, and later requests move back onto the tunnel
  with no client restart.
- No request ever blocks on ssh, including the first.

New environment overrides: `ARGUS_TUNNEL_RETRY_MIN_SECONDS`, `ARGUS_TUNNEL_RETRY_MAX_SECONDS`,
`ARGUS_TUNNEL_MIN_REQUESTS`.

## Proxy failover and selection

A dead primary proxy used to mean the cached config named a host that no longer answered, and the
client sat on Cloudflare through the whole backoff ladder. Worse, a forced refresh could hand back
the same dead host.

The tunnel API now returns a `proxies` array holding every active proxy, best choice for the
caller first. `TunnelConfig.candidates()` and `SSHTunnelConfig.Candidates()` expose them in order,
and each establish attempt walks the list before it gives up. A dead primary now costs one SSH
connect timeout. The array is additive, so a client that predates it reads the scalar
`proxy_host` and `proxy_port` fields and behaves as before.

Selection uses rendezvous hashing: score each proxy with `sha256(f"{user_id}:{config.id}")` and
sort by the score. It replaces a counter kept in a single `RuntimeStore` row that was rewritten on
every registration. Three properties changed:

- **No shared state.** The counter was one contended partition under concurrent CI load, and lost
  updates skewed the distribution it was meant to even out.
- **Stable per user.** A cached config and a later `GET /ssh/tunnel` now name the same primary.
  The counter let them disagree, so a re-registration could silently move a client.
- **Minimal churn.** Measured over 20000 users, adding a third proxy to two reassigns 33.5 % with
  rendezvous hashing against 67.3 % with rotation by `hash % len`, at the same load spread.

## Also in this pull request

- The Go CLI fails closed on host key verification. It used to fall back to
  `StrictHostKeyChecking=no` when it could not parse the stored entry, which is MITM-able. Two
  `TODO` comments go with it.
- Both copies of `tunnel-server-setup.sh` are removed. Proxy host provisioning belongs to the
  `argus_tunnel` Ansible role in `qatools-deployments`, which already does it better: a dedicated
  `argus-keys` account reads the service token from a 0640 file, instead of the token sitting in
  the world-readable wrapper that sshd requires. The matching role change is in
  `qatools-deployments`.
- `request()` no longer mutates the caller's `headers` dictionary when it injects tunnel headers.
- Client tests no longer reach the live API. The monitor now runs at construction, so four tests
  started making real calls to argus.scylladb.com. An autouse fixture stubs the resolver. The
  suite went from 2.16 s to 0.18 s.

## Deploy order

ScyllaDB builds a new global secondary index over existing rows asynchronously. Between the
backend deploy and the end of that build, a fingerprint lookup can miss and SSH authentication can
fail. Clients fall back to Cloudflare, which defeats the purpose.

1. Deploy the backend.
2. Wait for the index build to finish. Check `system_distributed.view_build_status`.
3. Raise `argus_tunnel_cli_version` and `cli_tools_argus_version` in `qatools-deployments` to the
   release that carries `--fingerprint`, then re-apply the `argus_tunnel` role.

The deprecated full-list path exists so that this order is safe. Remove it once every proxy host
runs the new wrapper.

Step 3 is safe to run early. The role reads the Argus CLI on the host and emits `%f` only when the
binary accepts `--fingerprint`, because an older one exits non-zero and sshd reads that as "no
authorized keys". The proxy host needs OpenSSH 6.9 or later for the `%f` token.

## Testing

- `pytest argus/client/tests/` — 96 pass. Ran the tunnel suite five times with no flakes.
- `cd cli && go build ./... && go test ./...` — 15 packages pass. `golangci-lint run ./...` is
  clean.
- New backend tests cover fingerprint scoping, malformed input, proxy order stability, spread, and
  the churn bound. The proxy-order tests stub `_get_active_configs`, so they need no container.
- New client tests cover failover to the next proxy, giving up only after every proxy, key-id
  propagation onto alternates, cache round-trip, monitor-driven retry, the backoff ladder, and the
  handshake threshold in both directions.

Backend tests that carry `docker_required` did not run locally. Neither did any test against a
real proxy host.


soyacz:
# Review: scylladb/argus PR #1057 — findings

## 1. Doc/code mismatch — fix before merge

`docs/plans/ssh-tunnel-design.md` says the Go client sets
`PubkeyAcceptedAlgorithms=ssh-ed25519`. `cli/internal/services/ssh.go`'s
`BuildSSHConnectArgs` does not set it, and `ssh_test.go` asserts it is absent
(OpenSSH 8.5+ only, older clients exit 255 on unknown option). Code and test
agree with each other and disagree with the doc. Drop the doc line or replace
it with the OpenSSH-8.5 reasoning already in the test comment.

## 2. Dead code — should fix

`SSHTunnel.reconnect()` and `DEFAULT_RECONNECT_RETRIES` in
`argus/client/tunnel/ssh.py` are unused. `_attempt_tunnel()` builds a fresh
`SSHTunnel()` each attempt instead of calling `.reconnect()`. Remove both, or
add a comment explaining why they stay.

## 3. Secondary-index read-after-write lag on the auth path — confirm before ship

`SSHTunnelKey.fingerprint` now gets a global secondary index, and
`get_authorized_keys()` reads through it instead of scanning the base table.
GSI propagation from base table to index is asynchronous on every write, not
just during the one-time backfill. A client that registers a key and
immediately attempts SSH (the normal fast path) can hit one denied auth
attempt under index lag, especially under concurrent registration load from
many CI runners — the load pattern this PR targets. New tests cover
register → lookup only as a single write, not under concurrent load, so this
gap is untested. Likely masked by client retry/failover, not correctness-
breaking, but confirm against a real proxy host or staging cluster under load
before merge.

## 4. Two unlocked race windows — non-blocking, minor fix

- `TunneledSession._request_count += 1` in `request()` has no lock. Worst
  case: a lost increment delays crossing `ARGUS_TUNNEL_MIN_REQUESTS` by one
  request.
- `_report_tunnel_failure()` sets `_tunnel_suspect = True` then
  `_tunnel_port = None` without a lock. A monitor tick reading between the
  two writes can re-publish a port about to be torn down, for one tick, then
  self-heals on the next tick. The comment above this code overstates the
  ordering guarantee — narrow the comment to match actual behavior.

argus/backend/models/ssh_key.py
```diff
@@ -27,12 +27,37 @@ class SSHTunnelKey(Model):
     expires_at = columns.DateTime(required=True)
 
 
+class SSHTunnelKeyByFingerprint(Model):
```
soyacz: why not just adding index to `SSHTunnelKey`?
Or maybe even replacint id to be a fingerprint?

CodeLieutenant: Is't this the Scylla way, create a separate table?

soyacz: No, we can use MV's and this is a good scenario. 
But better would be using fingerprint as a partition key - that would drop whole need for another table/view
@k0machi WDYT?

k0machi: I'd say if you only need a fingerprint to be queryable, an MV or SI on original table works as well, no need to duplicate. Only when you need to query the data by multiple columns you'd want a separate table.

CodeLieutenant:
> I'd say take an extra look at the table decision, it looks like it could just be a GSI instead of deploying a second table here.

Fixed using MV

